PR #101, opened 2024-10-16
Add Ollama and Microsoft.Extensions.AI chat sample

elbruno:
## PR Checklist

- [ ] New integration
  - [ ] Docs are written

## Other information

- Add a new Chat Web Sample app to [samples // ollama] 
- The new sample uses [Microsoft.Extensions.AI.Ollama](https://www.nuget.org/packages/Microsoft.Extensions.AI.Ollama/) to showcase Ollama chat integration.
- The main sample AppHost [Aspire.CommunityToolkit.Hosting.Ollama.AppHost] includes the ChatWeb App.

image

- The running project shows both chat implementations using OllamaSharp and Microsoft.Extensions.AI.Ollama

image


Directory.Packages.props
```diff
@@ -9,6 +9,8 @@
     <PackageVersion Include="Aspire.Hosting.NodeJS" Version="$(AspireVersion)" />
     <!-- AspNetCore packages -->
     <PackageVersion Include="Microsoft.AspNetCore.OpenApi" Version="$(AspNetCoreVersion)" />
+    <PackageVersion Include="Microsoft.Extensions.AI" Version="9.0.0-preview.9.24507.7" />
+    <PackageVersion Include="Microsoft.Extensions.AI.Ollama" Version="9.0.0-preview.9.24507.7" />
```
aaronpowell, line 13: Let's create a variable to use for that version

examples/ollama/Aspire.CommunityToolkit.Hosting.Ollama.AppHost/Program.cs
```diff
@@ -1,13 +1,18 @@
 var builder = DistributedApplication.CreateBuilder(args);
 
-var ollama = builder.AddOllama("ollama", port: null)
-    .AddModel("phi3")
-    .WithDefaultModel("phi3")
+var ollama = builder.AddOllama("ollama", port: 11434)
```
aaronpowell, line 3: Is there a reason to hard-code the port?

examples/ollama/Aspire.CommunityToolkit.Hosting.Ollama.AppHost/Program.cs
```diff
@@ -1,13 +1,18 @@
...
+    .AddModel("phi3.5")
+    .WithDefaultModel("phi3.5")
     .WithOpenWebUI();
 
 builder.AddProject<Projects.Aspire_CommunityToolkit_Hosting_Ollama_Web>("webfrontend")
     .WithExternalHttpEndpoints()
     .WithReference(ollama)
-    .WithEnvironment("ollama:model", "phi3");
+    .WithEnvironment("ollama:model", "phi3.5");
```
aaronpowell, line 11: There's no need to specify the model as an environment variable, we inject all model names as environment variables to the project/resource that it's referenced to:

Environment variables for Ollama models

I do need to document that...

examples/ollama/Aspire.CommunityToolkit.Hosting.Ollama.AppHost/Program.cs
```diff
@@ -1,13 +1,18 @@
...
+
+builder.AddProject<Projects.Aspire_CommunityToolkit_Hosting_Ollama_Web_MEAI>("webfrontendmeai")
+    .WithExternalHttpEndpoints()
+    .WithReference(ollama)
+    .WithEnvironment("ollama:model", "phi3.5");
```
aaronpowell, line 16: See above comment.

examples/ollama/Aspire.CommunityToolkit.Hosting.Ollama.Web.MEAI/Program.cs
```diff
@@ -0,0 +1,52 @@
...
+    .AddInteractiveServerComponents();
+
+builder.Services.AddOutputCache();
+
+builder.AddOllamaApiClient("ollama");
+
+builder.Services.AddChatClient(
+    builder => 
+        {
+            return new OllamaChatClient(new Uri("http://localhost:11434/"), modelId: "phi3.5");
```
aaronpowell, line 20: This should just rely on service discovery and use the model info from the config values exposed by the App Host

aaronpowell (changes requested):
I think that, rather than putting it into a separate project, we'd be better having a combined project for all Ollama client integrations (although, admittedly this isn't an integration for MEAI.Ollama).

Also, the structure of this does raise an interesting design aspect for how to consume Ollama in the client, as it's so far very structured towards OllamaSharp (there's a lot of referencing of that directly) which isn't ideal when there are multiple different clients that could consume the Ollama hosting package.

@eerhardt @SteveSandersonMS you might be interested in this.

examples/ollama/Aspire.CommunityToolkit.Hosting.Ollama.Web.MEAI/Program.cs
eerhardt: Why isn't this using 

https://github.com/CommunityToolkit/Aspire/blob/bd5e72f81db9d6ec83d9ba9ccfe99a1bb02927e8/src/Aspire.CommunityToolkit.OllamaSharp/AspireOllamaSharpExtensions.cs#L12-L19

?

aaronpowell: The `OllamaChatClient` is from MEAI.Ollama whereas the `IOllamaApiClient` type is what is registered by OllamaSharp.

eerhardt: I'm not 100% following. We are able to use both the Aspire.OpenAI client integration and Microsoft.Extensions.AI.OpenAI together. Why can't we do it with Ollama?

See

https://github.com/dotnet/eShop/blob/1acac04d4ad4247b6661103676718850e580c40b/src/WebApp/Extensions/Extensions.cs#L117-L122

aaronpowell: The `AddChatClient` method provided by MEAI expects an instance of `IChatClient`, which is what we get from https://github.com/dotnet/extensions/blob/ef606f2e1f40c69d4289d15e37c05e07dc20fd14/src/Libraries/Microsoft.Extensions.AI.Ollama/OllamaChatClient.cs#L21

Presently though, the OllamaSharp library doesn't implement that interface https://github.com/awaescher/OllamaSharp/blob/c9be5f578f598ec2f0b00e68c68f3be69acdace4/src/OllamaApiClient.cs#L22

This means that we can't put it in as an alternative for MEAI to access Ollama.

eerhardt: I get it now. `Microsoft.Extensions.AI.Ollama` doesn't take a dependency on OllamaSharp. It makes its own client. Unlike the M.E.AI.OpenAI library, which does depend on the OpenAI package.

Any timeline for OllamaSharp getting an implementation for `IChatClient`?

aaronpowell: Ah yes, sorry I should have opened with that.

I believe @SteveSandersonMS has approached the author of OllamaSharp on that topic

aaronpowell:
FYI - OllamaSharp has added support for the `IChatClient` interface, which we'll need to upgrade to and that addresses this feature

PR #106, opened 2024-10-17
Adding `ResourceName` attribute to our extension methods

### LINKED ISSUE #89: Add `ResourceNameAttribute` to the resource name arguments on integrations
This is a new attribute in Aspire 9 and while it doesn't appear to be used _yet_, it's setting the ground work for analyzers in the future.

aaronpowell:
Does require a bit of hackery in the tests to tell the analyser to be quiet but that's ok.

Fixes #89

### DIFF
--- a/src/Aspire.CommunityToolkit.Hosting.Azure.DataApiBuilder/DataApiBuilderHostingExtension.cs
+++ b/src/Aspire.CommunityToolkit.Hosting.Azure.DataApiBuilder/DataApiBuilderHostingExtension.cs
@@ -17,7 +17,7 @@ public static class DataApiBuilderHostingExtension
     /// <param name="port">The port number for the Data API Builder container.</param>"
     /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
     public static IResourceBuilder<DataApiBuilderContainerResource> AddDataAPIBuilder(this IDistributedApplicationBuilder builder,
-        string name,
+        [ResourceName] string name,
         string configFilePath = "./dab-config.json",
         int? port = null)
     {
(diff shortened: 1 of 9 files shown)


Alirexaa:
> Does require a bit of hackery in the tests to tell the analyser to be quiet but that's ok.

We could use `NoWarn` in tests.
```xml
<PropertyGroup>
  <NoWarn>ASPIRE006</NoWarn>
</PropertyGroup>
```

Alirexaa (approved):
I like to see `Resource` classes annotated by this attribute but it seems the analyser doesn't work with classes.
``` csharp
public class JavaAppContainerResource([ResourceName] string name, string? entrypoint = null)
    : ContainerResource(name, entrypoint), IResourceWithServiceDiscovery
{
    internal const string HttpEndpointName = "http";
}
```
but this LGTM

aaronpowell:
> > Does require a bit of hackery in the tests to tell the analyser to be quiet but that's ok.
> 
> We could use `NoWarn` in tests.
> 
> ```
> <PropertyGroup>
>   <NoWarn>ASPIRE006</NoWarn>
> </PropertyGroup>
> ```

That was a thought I had but I think it's best that we avoid having too much `NoWarn` in place because that is a good way to hide away stuff that we _probably_ should know about. Same reason why I did inline disabling of CTASPIRE001 in the tests rather than `NoWarn` on it.

aaronpowell:
> I like to see `Resource` classes annotated by this attribute but it seems the analyser doesn't work with classes.
> 
> ```cs
> public class JavaAppContainerResource([ResourceName] string name, string? entrypoint = null)
>     : ContainerResource(name, entrypoint), IResourceWithServiceDiscovery
> {
>     internal const string HttpEndpointName = "http";
> }
> ```
> 
> but this LGTM

Currently, the Aspire core hosting integrations don't do this, see https://github.com/dotnet/aspire/blob/269864e6fac3fd843072e676b2e87131899710be/src/Aspire.Hosting.PostgreSQL/PostgresServerResource.cs#L23 as an example

So I think it's best that we don't add it so we're consistent with how the core integrations are written.

PR #122, opened 2024-10-21
Better handling of ollama model discovery

### LINKED ISSUE #118: AddModel on Ollama not downloading
### Describe the bug

Adding two models to Ollama where the second one is a substring of the other does prevent the second one from downloading because to this check:
```csharp
var localModels = await ollamaClient.ListLocalModels(cancellationToken);
return localModels.Any(m => m.Name.StartsWith(model));
```

This can also happen when using a data valume and then changing the model to or adding a new one which is a substring of an already downloaded one

### Regression

_No response_

### Steps to reproduce

```csharp
var ollama = builder.AddOllama("ollama", port: null)
    .WithDataVolume()
    .AddModel("phi3")
    .AddModel("phi");
```

### Expected behavior

Determining if a model is downloaded should now rely on substring matching

### Screenshots

_No response_

### IDE and version

VS 2022 Preview

### IDE version

_No response_

### Nuget packages

- [ ] Aspire.CommunityToolkit.Hosting.Azure.StaticWebApps
- [ ] Aspire.CommunityToolkit.Hosting.Java
- [ ] Aspire.CommunityToolkit.Hosting.NodeJS.Extensions
- [x] Aspire.CommunityToolkit.Hosting.Ollama

### Nuget package version(s)

current git

### Additional context

_No response_

### Help us help you

aaronpowell:
**Closes #118**

Making the model name check happen on the exact model name rather than the start of the model name. This is to prevent false positives when the model name is a substring of another model name.

A unit test would have been good but I can't figure out how to do it without a huge overhaul to support mocking (which isn't ideal for a simple change).

## PR Checklist

- [ ] New integration
  - [ ] Docs are written

### DIFF
--- a/src/CommunityToolkit.Aspire.Hosting.Ollama/OllamaResourceLifecycleHook.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.Ollama/OllamaResourceLifecycleHook.cs
@@ -90,7 +90,7 @@ internal class OllamaResourceLifecycleHook(
             try
             {
                 var localModels = await ollamaClient.ListLocalModels(cancellationToken);
-                return localModels.Any(m => m.Name.StartsWith(model));
+                return localModels.Any(m => m.Name.Equals(model, StringComparison.OrdinalIgnoreCase));
             }
             catch (TaskCanceledException)
             {


src/Aspire.CommunityToolkit.Hosting.Ollama/OllamaResourceLifecycleHook.cs
```diff
@@ -90,7 +90,7 @@ private static async Task<bool> HasModelAsync(OllamaApiClient ollamaClient, stri
             try
             {
                 var localModels = await ollamaClient.ListLocalModels(cancellationToken);
-                return localModels.Any(m => m.Name.StartsWith(model));
+                return localModels.Any(m => m.Name.Equals(model, StringComparison.OrdinalIgnoreCase));
```
Mrxx99, line 93: One Note here: the localModels contain the tag, e.g. "phi3:latest". I guess that's why substring matching was added. I don't know if there is a problem when e.g. using a DataVolume and a older version is already there.

aaronpowell: I'm not sure how the internals of Ollama work, but I'd expect that it works similar to how Docker works with tags - if you asked for a mutable tag like `latest` it's "latest per last time I had nothing tagged latest" and it won't pull a new one while that tag matches.

So yes, if you're using a persistent model, you have the chance that what your volume thinks is `latest` isn't really the `latest` according to Ollama remote.

As I've been writing this, I'm thinking that with Aspire 9 I'll add a way to delete models using the [resource commands feature](https://learn.microsoft.com/dotnet/aspire/whats-new/dotnet-aspire-9-release-candidate-1?tabs=unix&pivots=visual-studio#resource-commands) which will help with managing models without having to delete the whole volume each time.

Like with any tag-based versioning system, you should always use an explicit tag rather than a movable one like `latest`.

Mrxx99: I just tested what happens, when you try to download a different tag version, e.g. you already have phi3:latest in the data volume, but when you now change it to phi:chat, it will try to download it, but the download method will just complete without doing anything. Maybe some logic should check which models are already there (which are not currently requested) and delete those where one with same name and different tag is requested, so the requested tag version can be installed.

aaronpowell: ok, I'll have a bit more of a play around on this before it can be merged

aaronpowell: I've been having a play around with this and I'm struggling to find how this fix won't address the problem, but there is a potential for a problem that isn't really a technical one to solve.

From what I can see, here's a chance where this will fall down:

* You run a project for the first time and it pulls a model using the specified tag, let's say it's `phi3:latest`
* You don't touch the project fora few weeks
* In that time, a new SHA is published for the `latest` tag of the model
* You return to the project and run it again, since the volume was persisted, the code check says that `phi3:latest` already exists in the Ollama volume, so we don't need to pull the model. What this fails to recognise is that `latest` as far as your volume is concerned is the SHA `3b387c8dd9b7` when it really is `61819fb370a3`, meaning that you don't pull the actual `latest` tag and thus are using an "outdated" model

This kind of data drift is something that's inherently going to happen in any system that persists data between application runs, especially systems that don't pin version dependencies.

I don't think that the hosting integration should try to be "smart" and figure out between runs if there has been a drift in the tags between local and upstream, if you're going to use persisted volumes then it is logical that you should also pin your dependencies.

But, as mentioned above, I think that it'd be a good idea that as part of the Aspire 9 release we should have resource commands that allow you to do model management.

Mrxx99: Yes, same tag changing is a problem not so easy to solve. The scenario above where the tag is changed is solvable though. But I can create a new issue for that. This PR I think is fine to solve the original issue.

Mrxx99 (approved):

PR #129, opened 2024-10-23
Removing deprecated API

aaronpowell:
Removes deprecated Ollama API


Alirexaa (commented):
Why we remove this API?
Is beacase that we publish beta packages to nuget?

aaronpowell:
Yep, with the packages now out and available, I'd like to target the Aspire 9 release as the time that we remove the deprecated API, since there's enough time to warrant testing and migrating.

Alirexaa (approved):

PR #141, opened 2024-10-23
Release v8.2.1

aaronpowell:
Moved all the unshipped APIs to shipped to release


Alirexaa (approved):
LGTM.
Did you use https://github.com/CommunityToolkit/Aspire/blob/main/eng%2Fapicompat%2Fmark-shipped.cmd for this?

aaronpowell:
No, I had Copilot generate me a bash script as I am on Linux not Windows so don't have cmd or PowerShell.

I should put a bash script in there rather than generating one 🤣

Alirexaa:
@aaronpowell, We should keep `release/*` branches in case patch updates.

aaronpowell:
> @aaronpowell, We should keep `release/*` branches in case patch updates.

We've got the tag on the SHA that was used to trigger the build that generated the NuGet packages, we can always branch off there.

Issue #107, opened 2024-10-17
Ollama hosting integration is very tied to OllamaSharp


aaronpowell:
As we see in #101 the way that we have the Ollama hosting integration is very much just expecting that you will be using it with the OllamaSharp client integration. But with the recent preview release of [Microsoft.Extensions.AI](https://devblogs.microsoft.com/dotnet/introducing-microsoft-extensions-ai-preview/) which also supports Ollama, we may need to rethink the way that everything is surfaced out, so that it is treated as a more "generic" integration.

The main concern point is on the config settings for the available models/default model, which registers a config setting of `Aspire:OllamaSharp:<resource>:...`. While this is fine for binding with OllamaSharp, if we want to use that same config value in the MEAI we still must call it OllamaSharp - which is sub-optimal.

aaronpowell:
Thinking through this some more and looking at how we're fleshing out the change in #143 it's looking like we might be better off using a connection string structure for the integration rather than environment variables and a raw endpoint.

So, the `OllamaResource` connection string would be `Endpoint=http://...` and then for the `OllamaModelResource` we would have `Endpoint=http://...;Model=<model name>`.

We could also put all available models in a comma-separated list on the connection string and parse that too, or just not provide them (which might be better).

aaronpowell:
We've merged in the new connection string format and support in OllamaSharp. This will be released in the v9 release.

PR #143, opened 2024-10-23
Change Ollama AddModel to create a separate resource

### LINKED ISSUE #119: Allow specifying selected model when calling WithReference with a Ollama resource

Yes

### Existing integration

Aspire.CommunityToolkit.Hosting.Ollama

### Overview

Allow specifying selected model when using WithReference with a Ollama resource.
Currently the selected model can only be easily set by using WithDefaultModel, but that locks to only one model and can't reference different ones in different resources easily. This can easily be solved by having an optional parameter on the custom WithReference for Ollama which allows specifying a selected model. If this optional parameter is not set, it will behave exactly the same as before: It would use the one specivied with WithDefaultModel on the Ollama resource.

```diff
+ public static IResourceBuilder<T> WithReference<T>(this IResourceBuilder<T> builder, IResourceBuilder<OllamaResource> ollama, string? selectedModel = null ) where T : IResourceWithEnvironment
- public static IResourceBuilder<T> WithReference<T>(this IResourceBuilder<T> builder, IResourceBuilder<OllamaResource> ollama) where T : IResourceWithEnvironment
```

### Usage example

```csharp
var ollama = builder.AddOllama("ollama", port: null)
    .AddModel("phi")
    .AddModel("phi3")
    .WithDefaultModel("phi")
    .WithOpenWebUI();

builder.AddProject<Projects.Aspire_CommunityToolkit_Hosting_Ollama_Web>("webfrontend")
    .WithExternalHttpEndpoints()
    .WithReference(ollama, /* adding this */ "phi3");
``` 

### Breaking change?

No

there is no release yet (also this only adds an optional parameter to an existing method)

### Alternatives

Setting the environment variable (Aspire__OllamaSharp__{resource.Name}__SelectedModel) manually, but that is not easily discoverable without looking at this source code. Maybe the way currently in the playground was added with the intention to do this, but it does not work:
```csharp
builder.AddProject<Projects.Aspire_CommunityToolkit_Hosting_Ollama_Web>("webfrontend")
    .WithExternalHttpEndpoints()
    .WithReference(ollama)
    .WithEnvironment("ollama:model", "phi3");
```

### Additional context

_No response_

### Help us help you

Mrxx99:
**Closes #119**

As discussed in #119 this changes the AddModel methods on Ollama resource to create and return a new OllamaModelResosurce, that can be used with WithReference instead of using the Ollama resource.
The advantage is, that now one Ollama resource can have multiple models, that are referenced by same or different applications. Previously only the default model could be used without manually setting complicated env variables.

## PR Checklist

- [ ] New integration
  - [ ] Docs are written

This is a breaking change so I guess v9 would be a good target for this.
It breaks how AddModel works previously it would return the Ollama resource back, so whenever this was used, the code would have to change to call AddModel on a variable that was returned from AddOllama instead. The WithDefaultValue still works, this limits a bit the breaking change.

## Other information

This is still work in progress, but I already created the PR, so some design decisions can be discussed.
E.g. should it still work with a custom WithReference extension and environment variables or could a connection string that includes the model name be used instead? The PR implementation provides both but uses the custom WithReference one, if that method was removed it would use the connection string one.


src/CommunityToolkit.Aspire.Hosting.Ollama/OllamaModelResource.cs
```diff
@@ -0,0 +1,28 @@
...
+/// <param name="modelName">The name of the LLM model, can include a tag.</param>
+/// <param name="parent">The <see cref="OllamaResource"/> parent.</param>
+public class OllamaModelResource(string name, string modelName, OllamaResource parent) : Resource(name), IResourceWithParent<OllamaResource>, IResourceWithConnectionString
+{
+    public OllamaResource Parent { get; } = ThrowIfNull(parent);
+
+    /// <summary>
+    /// Gets the connection string expression for the Ollama model.
+    /// </summary>
+    public ReferenceExpression ConnectionStringExpression => ReferenceExpression.Create($"{Parent}?Model={ModelName}");
```
aaronpowell, line 19: This is getting me to rethink the "connection string" concept that we have and whether we should go to a "real" connection string like `Endpoint=...;Model=...` structure.

That would be a breaking change but it does align better with how other Aspire integrations work (and we can use the `DbConnectionStringBuilder` class to parse it).

aaronpowell (commented):
This looks **amazing**, thanks @Mrxx99!

I have one question about the connection string for models and I think doing it would help address #107

aaronpowell:
Let's also change this to merge into the aspire-9 branch not main

Mrxx99:
@aaronpowell  I did the change to use a proper DB connection string and removed the environments variables. Should the 'WithDefaultModel' approach be kept or also removed?

aaronpowell:
Let's use the connection string approach the whole way down. We can accept a breaking change going into the v9 release (which this will land) as we're also really only soft launched at the moment so usage is low (and thus if we were to change things "radically" it won't be a huge problem).

Same goes for the `WithDefaultModel`, let's remove that as there's now a better approach to providing that through the work you've done.

examples/ollama/CommunityToolkit.Aspire.Hosting.Ollama.AppHost/Program.cs
```diff
@@ -1,13 +1,14 @@
 var builder = DistributedApplication.CreateBuilder(args);
 
 var ollama = builder.AddOllama("ollama")
-    .AddModel("phi3")
-    .WithDefaultModel("phi3")
+    .WithDefaultModel("phi:chat")
```
aaronpowell, line 4: As mentioned in a recent comment, let's remove the `WithDefaultModel` method and instead you always use `WithModel` and provide the model reference.

src/CommunityToolkit.Aspire.Hosting.Ollama/OllamaResource.cs
```diff
@@ -35,10 +35,21 @@ public class OllamaResource(string name) : ContainerResource(name), IResourceWit
...
+    internal ReferenceExpression BuildConnectionString(bool includeDefaultModel = false)
+    {
+        var builder = new ReferenceExpressionBuilder();
+        builder.AppendLiteral($"Endpoint=");
+        builder.Append($"{PrimaryEndpoint.Property(EndpointProperty.Scheme)}://{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)}");
+
+        if (includeDefaultModel && !string.IsNullOrEmpty(DefaultModel))
+        {
+            builder.AppendLiteral($";Model={DefaultModel}");
+        }
```
aaronpowell, line 49: We'll be able to drop this with the removal of the concept of default model

src/CommunityToolkit.Aspire.Hosting.Ollama/OllamaResourceBuilderExtensions.cs
```diff
@@ -55,63 +51,76 @@ public static IResourceBuilder<OllamaResource> WithDataVolume(this IResourceBuil
...
-    /// <param name="ollama">The Ollama resource to add as a reference.</param>
+    /// <param name="builder">The <see cref="IDistributedApplicationBuilder"/>.</param>
+    /// <param name="name">The name of the resource.</param>
+    /// <param name="modelName">The name of the LLM from Hugging Face to download on initial startup.</param>
     /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
-    /// <remarks>
-    /// This method adds the connection string and model references to the application model.
-    /// </remarks>
-    public static IResourceBuilder<T> WithReference<T>(this IResourceBuilder<T> builder, IResourceBuilder<OllamaResource> ollama) where T : IResourceWithEnvironment
+    public static IResourceBuilder<OllamaModelResource> AddHuggingFaceModel(this IResourceBuilder<OllamaResource> builder, string name, string modelName)
```
aaronpowell, line 91: This method could do a pass-through to the `AddModel` with the prefix of `hf.co/`, which would reduce a little bit of code duplication.

src/CommunityToolkit.Aspire.Hosting.Ollama/OllamaResourceBuilderExtensions.cs
```diff
@@ -55,63 +51,76 @@ public static IResourceBuilder<OllamaResource> WithDataVolume(this IResourceBuil
...
+        builder.AddModel(modelName);
 
-                if (model == ollama.Resource.DefaultModel)
-                {
-                    context.EnvironmentVariables[$"Aspire__OllamaSharp__{resource.Name}__SelectedModel"] = model;
-                }
-            }
-        });
+        builder.Resource.SetDefaultModel(modelName);
+        return builder;
```
aaronpowell, line 123: As discussed, let's remove this method and instead you'll use the `OllamaModelResource` as the thing you pass in if you want a default model specified.

src/CommunityToolkit.Aspire.Hosting.Ollama/PublicAPI.Shipped.txt
```diff
@@ -11,10 +11,8 @@ Aspire.Hosting.ApplicationModel.OpenWebUIResource.ConnectionStringExpression.get
 Aspire.Hosting.ApplicationModel.OpenWebUIResource.OpenWebUIResource(string! name) -> void
 Aspire.Hosting.ApplicationModel.OpenWebUIResource.PrimaryEndpoint.get -> Aspire.Hosting.ApplicationModel.EndpointReference!
 Aspire.Hosting.OllamaResourceBuilderExtensions
-static Aspire.Hosting.OllamaResourceBuilderExtensions.AddModel(this Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.ApplicationModel.OllamaResource!>! builder, string! modelName) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.ApplicationModel.OllamaResource!>!
```
aaronpowell, line 14: We need to leave this method here and instead it'll get marked with REMOVED in the unshipped file.

@Alirexaa we should document this in the contributing guide better.

src/CommunityToolkit.Aspire.Hosting.Ollama/PublicAPI.Shipped.txt
```diff
@@ -11,10 +11,8 @@ Aspire.Hosting.ApplicationModel.OpenWebUIResource.ConnectionStringExpression.get
...
 static Aspire.Hosting.OllamaResourceBuilderExtensions.WithDefaultModel(this Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.ApplicationModel.OllamaResource!>! builder, string! modelName) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.ApplicationModel.OllamaResource!>!
 static Aspire.Hosting.OllamaResourceBuilderExtensions.WithOpenWebUI<T>(this Aspire.Hosting.ApplicationModel.IResourceBuilder<T!>! builder, System.Action<Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.ApplicationModel.OpenWebUIResource!>!>? configureContainer = null, string? containerName = null) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<T!>!
-static Aspire.Hosting.OllamaResourceBuilderExtensions.WithReference<T>(this Aspire.Hosting.ApplicationModel.IResourceBuilder<T>! builder, Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.ApplicationModel.OllamaResource!>! ollama) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<T>!
```
aaronpowell, line 20: We need to leave this method here and instead it'll get marked with REMOVED in the unshipped file.

src/CommunityToolkit.Aspire.OllamaSharp/AspireOllamaSharpExtensions.cs
```diff
@@ -10,7 +11,7 @@ namespace Microsoft.Extensions.Hosting;
 /// </summary>
 public static class AspireOllamaSharpExtensions
 {
-    private const string DefaultConfigSectionName = "Aspire:OllamaSharp";
+    private const string DefaultConfigSectionName = "Aspire:Ollama";
```
aaronpowell, line 14: This needs to stick with `OllamaSharp` as that's the client this is relevant for

aaronpowell (changes requested):
Got a few more changes in here but it's shaping up really nicely.

I'm loving how the integration now looks

mitchdenny:
This PR is kind of setting a convention for how to pass around the model name in the connection string. I think this is probably the right thing to do but I know that this has been a contentious issue so should be discussed.

For me the most important thing is the understand how a developer moves between different API providers. So for local development they are probably going to use Ollama, but when they deploy do they want to adopt Open AI/Azure Open AI. Can they host things

```csharp
IResourceBuilder<IResourceWithConnectionString> model = builder.ExecutionContext.IsRunMode
                                                          ? builder.AddOllama(...).AddModel(...)
                                                          : builder.AddAzureOpenAI(...).AddModel(...);
```

aaronpowell (approved):
I feel like this is ready to go

aaronpowell:
> This PR is kind of setting a convention for how to pass around the model name in the connection string. I think this is probably the right thing to do but I know that this has been a contentious issue so should be discussed.
> 
> For me the most important thing is the understand how a developer moves between different API providers. So for local development they are probably going to use Ollama, but when they deploy do they want to adopt Open AI/Azure Open AI. Can they host things
> 
> ```cs
> var builder = DistributedApplication.CreateBuilder(args);
> IResourceBuilder<IResourceWithConnectionString> model = builder.ExecutionContext.IsRunMode
>                                                           ? builder.AddOllama(...).AddModel(...)
>                                                           : builder.AddAzureOpenAI(...).AddModel(...);
> ```

There are some complexities that come with transitioning between different AI services, but I think that the introduction of the Microsoft.Extensions.AI (and the updates coming in #164 to support it for Ollama) will help with that by having the client abstraction away and then it becomes a case of the client integration <-> hosting integration, as you'd likely use OllamaSharp + Ollama Hosting then move to the OpenAI/Azure OpenAI set.

aaronpowell:
@Mrxx99 - I think we're looking good, anything left from your end before putting this in for a merge?

aaronpowell:
> Should we keep the AddModel(string) overload where the resource name is generated? I am not 100% sure about the sanitizing.

I think we can combine them into a single method, make `modelName` optional and if it's not provided we use `name` for it.

Digging through Aspire's source, I can't see anywhere that would have a problem if the resource name as non-alphanumeric characters, so we could probably get away with a resource named `phi3:chat` or `huggingface.io/something`.

aaronpowell:
I've decided to merge it as-is so I can get a head start on the merge conflicts that will come with #164 

Thanks again, I'm really excited at how the Ollama integration is now looking.

PR #147, opened 2024-10-23
Changing versioning for builds

### LINKED ISSUE #145: Versioning changes
After the first release it's become apparent that the versioning approach of using the `AspireVersion` MSBuild variable as `PackageVersion` is going to be a problem. For example, how do we do CI builds when version 8.2.1 is released but the CI will make everything 8.2.1 still? we'd keep pushing pre-release packages under the 8.2.1 even though it's been released.

So instead, we'll move to a version scheme of `<Aspire Major>.<Toolkit Minor>.<Toolkit Patch>` (and then a pre-release suffix where required).

This gives us slightly more control while also still making in clear which release of Aspire we're dependant on.

aaronpowell:
Added a failure condition for release builds that if you try to release a version that doesn't match what we'd been building the CI will fail.

Example of failure

Fixes #145

### DIFF
--- a/.github/workflows/dotnet-release.yml
+++ b/.github/workflows/dotnet-release.yml
@@ -98,7 +98,7 @@ jobs:
 
       - name: Publish NuGet package
         if: ${{ matrix.os == 'ubuntu-latest' }}
-        run: dotnet pack --no-build -c ${{ env.DOTNET_CONFIGURATION }} -o ./nuget -p:PackageVersion=${{ steps.git_tag.outputs.tag }}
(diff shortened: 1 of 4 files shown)


Alirexaa:
Should we update https://github.com/CommunityToolkit/Aspire/blob/main/docs/versioning.md?

aaronpowell:
Ah, i didn't realize there was anything specific in the version doc on that. Yep, will update

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

PR #152, opened 2024-10-24
Code coverage reporting

### LINKED ISSUE #111: Introduce min code coverage across the integrations.
Probably should look at whether we can introduce min code coverage across the integrations. I'll have to look into how aspire does it.

_Originally posted by @aaronpowell in https://github.com/CommunityToolkit/Aspire/pull/66#discussion_r1790973392_

aaronpowell:
Adding code coverage reporting to the CI jobs.

Closes #111


.github/workflows/dotnet-ci.yml
```diff
@@ -135,3 +135,59 @@ jobs:
...
+        with:
+          filename: "Cobertura.xml"
+          badge: true
+          fail_below_min: false
+          format: markdown
+          hide_branch_rate: false
+          hide_complexity: false
+          indicators: true
+          output: both
+          thresholds: "10 30"
```
Alirexaa, line 179: Is this the the min coverage and will anything under this break the CI?
Is possible for any packages to have their own the min coverage instead of using this one for all packages?

aaronpowell: This should fail the build according to https://github.com/irongut/CodeCoverageSummary/tree/v1.3.0/?tab=readme-ov-file#thresholds although I probably need to learn more about how to best set them (I copied this from the CommuntiyToolkit/Datasync repo 😏).

To have each package provide their own min threshold (like the dotnet/aspire repo has) we'd need something more custom here and that feels like we'd have to manage a bit more code. I think that we can get away with a blanket rule of "your coverage must meet X" (and generally speaking, we're pretty highly covered).

aaronpowell: I've adjusted the thresholds and it looks like it is marking things as failed but not the build yet, I think it'll only fail if the overall threshold drops below the min value, not if one of them is below. I'll keep experimenting

aaronpowell:
@Alirexaa - how's that looking? We can fail builds with thresholds.

I think we probably want to do 60% minimum though, so we'd be currently passing (just with some needing improvement)

Alirexaa (commented):
LGTM.
Let's use 60% for min rate.
Do not fotget to update `dotnet-main` and `dotnet-release`.

aaronpowell:
All ready now, I've created a reusable workflow and it's shared across all three workflows so that we have consistency

Alirexaa (approved):

PR #175, opened 2024-10-29
SWA health checks

### LINKED ISSUE #74: Leverage `WaitFor` in SWA emulator
The SWA emulator has dependency on external resources, the frontend, possible backend and DAB resource. The emulator expects these to be running when it starts up, but it's a soft-dependency.

With Aspire 9 we'll have access to the `WaitFor` feature on resources to delay their start, so we should integrate that.

aaronpowell:
**Closes #74**

Using the new `WaitFor` API in Aspire 9 means that if the app/api resources provide health checks then we will wait for them to be healthy before the SWA emulator starts, otherwise it will wait until the resource launches.

Then added the built-in `WithHttpHealthCheck` for the SWA emulator itself and pointing that at a known endpoint (`/.auth/me`)  and once that's up we are healthy.

## PR Checklist

- [ ] New integration
  - [ ] Docs are written


Alirexaa (commented):
I'm not familiar with SWA, but I think we should have a test to check `waitfor`  wait for the resource to be ready in addition to checking annotation.

We have an example of Meilisearch:
https://github.com/CommunityToolkit/Aspire/blob/d10ec7f36d34585eb170a414f8bc6fde84545180/tests/CommunityToolkit.Aspire.Hosting.Meilisearch/MeilisearchFunctionalTests.cs#L193-L228

aaronpowell:
I'll have a play with it and test a few falling scenarios. 

The SWA emulator is a proxy in front of other http endpoints, so we wait for them before the emulator starts.

aaronpowell:
@Alirexaa I've had a play around with writing tests like that and I don't think they are going to give the best value to effort in writing and maintaining, at least not for SWA.

I have unit tests in the PR that assert that the `WaitAnnotation` is being added to the `SwaResource` for the app/api when you call their respective methods, which I think is the best approach here. If we try to write integration tests and simulate failing health checks, what we're really doing is testing how the health check system in Aspire works rather than just our integrations usage of it.

Alirexaa (approved):

PR #180, opened 2024-10-29
Update `ToolkitMinorVersion` to 0

Alirexaa:
`ToolkitMinorVersion` should be zero. we should not skip 9.0.* and 9.1.* versions.

@aaronpowell  What happens in the Nuget feed when we merge this PR?
Will the `9.3.0-beta.52` be listed as the last version?


aaronpowell (approved):
Thanks, must have stuffed up the merge at one point.

This will make some NuGet fun 😔

PR #182, opened 2024-10-29
Multi-target frameworks

### LINKED ISSUE #57: .NET 9 support
The packages created only target .NET 8 but we should multi-target with .NET 9 as well.

Alirexaa:
Closes #57 

This pull request updates the target framework configuration across multiple project files to support both `net8.0` and `net9.0`. The key changes include the introduction of a default target framework and the removal of individual target framework specifications from various project files.

Framework Configuration Updates:

* [`Directory.Build.props`](diffhunk://#diff-9da24614831c308827a1ae533ffea392c97638c261dd42bd0f5226baa136d16eL7-R10): Introduced `DefaultTargetFramework` set to `net8.0`, added `TargetFrameworks` and `AllTargetFrameworks` properties to support multiple frameworks (`net8.0` and `net9.0`).

## PR Checklist

- [ ] New integration
  - [ ] Docs are written


aaronpowell:
I've looked into this previously and had some problems with multiple TFM's when it came to the App Host and other ASP.NET web apps - I think they have to be a single TFM

src/CommunityToolkit.Aspire.Hosting.Meilisearch/CommunityToolkit.Aspire.Hosting.Meilisearch.csproj
```diff
@@ -3,6 +3,8 @@
   <PropertyGroup>
     <AdditionalPackageTags>hosting meilisearch</AdditionalPackageTags>
     <Description>Meilisearch support for .NET Aspire.</Description>
+    <!--MeiliSearch package have vulnerable dependency-->
+    <NoWarn>NU1903</NoWarn>
```
aaronpowell, line 7: We might need to do a PR with these into `main`, or remove .NET 9 from the CI job

Alirexaa: I will send a PR.

ErikEJ:
Why is this needed? Are there any particular .net 9 APIs that we want to use?

aaronpowell:
We've got .net 9 on CI so we can do multi targeting. With it now being released, it's the default . net SDK for builds, and since the 2.9.1 package doesn't support 9 it fails the build

ErikEJ:
@aaronpowell Thanks, I understand the SQL build SDK issue, but why do you need to multi target in general?

aaronpowell:
> @aaronpowell Thanks, I understand the SQL build SDK issue, but why do you need to multi target in general?

Sorry, thought I was replying in the other PR 🤣

The answer - NEW AND SHINY!!

But also, because there are going to be people who want to use .NET 9 so we can target multiple TFM's

aaronpowell (approved):
Changes all reviewed as of now and they look good.

Let's see what CI has to say.

Alirexaa:
@ErikEJ @jmezach
The following test fails in this branch. Please take a look.
```
CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.Tests.AddSqlProjectTests.AddSqlProject_WithProjectMetadata
```

Alirexaa:
> @Alirexaa can you share a specific link?

https://github.com/CommunityToolkit/Aspire/actions/runs/11823424047/job/32942697395

```
 CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.Tests.AddSqlProjectTests.AddSqlProject_WithProjectMetadata [FAIL]
  Failed CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.Tests.AddSqlProjectTests.AddSqlProject_WithProjectMetadata [2 s]
  Error Message:
   Assert.True() Failure
Expected: True
Actual:   False
  Stack Trace:
     at CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.Tests.AddSqlProjectTests.AddSqlProject_WithProjectMetadata() in /_/tests/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.Tests/AddSqlProjectTests.cs:line 24
```

jmezach:
@Alirexaa Just saw your commit. I was suggesting doing an exclusion for our example `SdkProject.csproj`, rather than the actual implementation project CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.csproj

Alirexaa:
@jmezach, With the last change, the problem still exists.

jmezach:
I can reproduce this issue locally as well. We are using MSBuild to load the referenced `.csproj` project file and then get the `TargetPath` property from there. This currently returns as an empty string on this branch, thus the test fails.

My guess is that when we're loading the project file the `$(DefaultTargetFramework)` property is not set and therefore we cannot resolve the `TargetPath`. This might be caused by the fact that we're trying to load the project file using a relative path. I'll try to see if we can make that absolute and see if that changes anything.

jmezach:
Ah, I see what's going wrong here. We're still using `<TargetFrameworks>`, as in **plural**. This is not something we currently support with MSBuild.Sdk.SqlProj.

Alirexaa:
> Ah, I see what's going wrong here. We're still using `<TargetFrameworks>`, as in **plural**. This is not something we currently support with MSBuild.Sdk.SqlProj.

I tried `<TargetFramework>` for SdkProject.csproj locally and it did not work.

jmezach:
I've just pushed that change, let's see what CI thinks

aaronpowell (commented):
Are we only going to be able to run tests as net8.0?

Alirexaa:
> Are we only going to be able to run tests as net8.0?

I think so.

aaronpowell:
> > Are we only going to be able to run tests as net8.0?
> 
> I think so.

If that unblocks us to move forward then let's go with it, but I'd like to revisit it.

Maybe we need to rethink testing long term and split unit and integration tests out, having unit tests across LTS and STS (so net8.0 and net9.0), while integration tests are LTS only, since they rely on examples which can't target multiple TFMs

PR #188, opened 2024-10-30
Adding multiple file support to AddDataAPIBuilder()

### LINKED ISSUE #187: AddDataAPIBuilder() doesn't support multiple configuration files.
The current syntax:

```csharp

// Add Data API Builder using dab-config.json 
var dab = builder.AddDataAPIBuilder("dab")
    .WithReference(sqlDatabase)
    .WaitFor(sqlServer);

builder.Build().Run();
```

This does not make it clear how to include configuration files. 

### I suggest two important methods:

|Method|Intent
|-|-
|`WithConfigurationFile(string)`|Mounts the file in the container
|`WithConfigurationFiles(string[])`|Mounts multiple files in the container

```csharp
var dab = builder.AddDataAPIBuilder("dab")
    .WithConfigurationFile("./dab-config.json") // new
    .WithReference(sqlDatabase)
    .WaitFor(sqlServer);
```

and

```csharp
var dab = builder.AddDataAPIBuilder("dab")
    .WithConfigurationFiles("./dab-config.json", "./dab-config-2.json") // new
    .WithReference(sqlDatabase)
    .WaitFor(sqlServer);
```

### This is important because:

1. Multiple data source support in DAB requires multiple configuration files.
2. Azure Cosmos DB support requires multiple configuration files (schema files). 

### Workaround:

I realize this can be done today with `Mount()` but this undermines the simplicity of the integration.

```csharp
var dab = builder.AddDataAPIBuilder("dab")
    .WithReference(sqlDatabase)
    .WaitFor(sqlServer);

// configuration
var config = "path";
var internalPath = $"/App/{System.IO.Path.GetFileName(config)}"
dab.WithBindMount(config, internalPath) 
```

and (multiple configuration files)

```csharp
var dab = builder.AddDataAPIBuilder("dab")
    .WithReference(sqlDatabase)
    .WaitFor(sqlServer);

// configurations
var configs = new[] { "path", "path", "path" };
foreach (var config in configs)
{
    var internalPath = $"/App/{System.IO.Path.GetFileName(config)}"
    dab.WithBindMount(config, internalPath) 
}
```

JerryNixon:
Closes #187

It is required that Data API builder accept more than one configuration file. This is imperative to support Cosmos DB and important to support developers using Data API builder against multiple data sources. 

## Changes

 - Removed `AddDataAPIBuilder(string, string, int?)` as it is redundant to new methods (below).
 - Added `AddDataAPIBuilder(string, int?, params string[])` overload to support multiple configuration files.
 - Added `AddDataAPIBuilder(string, params string[])` support without specifying port or named arguments.
 - Updated readme.md with new and additional samples & documentation.
 - Added Unit Tests against new signature(s). 
 - Included configuration files for testing purposes.
 - Modified `PublicAPI.Unshipped.txt` to reflect the updated `AddDataAPIBuilder` method signature(s).
 - Updated code and docs to support HTTP endpoints and explicitly not support HTTPS in dev.

## Previous signature

```csharp
public static IResourceBuilder<DataApiBuilderContainerResource> AddDataAPIBuilder(
    this IDistributedApplicationBuilder builder,
    [ResourceName] string name,
    string configFilePath = "./dab-config.json",
    int? port = null)
```

## New signature

```csharp
public static IResourceBuilder<DataApiBuilderContainerResource> AddDataAPIBuilder(
    this IDistributedApplicationBuilder builder,
    [ResourceName] string name,
    int? httpPort = null,
    params string[] configFilePaths)
```

### New usage (multiple files, parameter names required - when no port)

```csharp
var dab = builder
    .AddDataAPIBuilder("dab", configFilePaths: "./dab-condig.json", "./dab-condig-2.json");
```

## Also added overload

```csharp
public static IResourceBuilder<DataApiBuilderContainerResource> AddDataAPIBuilder(
    this IDistributedApplicationBuilder builder,
    [ResourceName] string name,
    params string[] configFilePaths)
```

### New usage (multiple files, no parameter names required)

```csharp
var dab = builder
    .AddDataAPIBuilder("dab", "./dab-condig.json", "./dab-condig-2.json");
```


src/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder/DataApiBuilderHostingExtension.cs
aaronpowell, line 1: As I think about this and the use-cases for multiple configs, I think that we should also add a method like `WithConfigurationFile(string file)`.

My thinking is that you may want to layer on a dev config file using a check in the code, and it would be easier if you could do that through a conditional method call rather than having to use different calls to `AddDataAPIBuilder`.

src/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder/DataApiBuilderHostingExtension.cs
```diff
@@ -13,25 +13,56 @@ public static class DataApiBuilderHostingExtension
...
+            configFilePaths = ["./dab-config.json"];
+        }
+
+        foreach (var configFilePath in configFilePaths)
+        {
+            var configFileName = File.Exists(configFilePath)
+                ? Path.GetFileName(configFilePath)
+                : throw new FileNotFoundException($"Config file not found: {configFilePath}");
+
+            rb.WithBindMount(configFilePath, $"/App/{configFileName}", true);
```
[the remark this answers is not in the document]
aaronpowell: Ah, I see that that won't work when it comes to the GraphQL file required with CosmosDB.

I think that we should handle Cosmos slightly differently though, since it's required that you provide `schema.graphql` when you do a Cosmos backend, we should enforce that in the API:

```csharp
public IResourceBuilder<DataApiBuilderContainerResource> WithReference(this IResourceBuilder<DataApiBuilderContainerResource> builder, CosmosResource resource, string schemaFile)
{
    return builder.WithResource(resource).WithBindMound(schemaFile, "/App/schema.graphql");
}
```

This way, from an API discoverability perspective, it's very obvious of what the requirements are.

tests/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder.Tests/ContainerResourceCreationTests.cs
```diff
@@ -49,4 +40,66 @@ public void AddDataAPIBuilderContainerDetailsSetOnResource()
...
+    }
+
+    [Fact]
+    public void AddDataAPIBuilderContainer_ValidFileWithPort_NoEx()
+    {
+        IDistributedApplicationBuilder builder = DistributedApplication.CreateBuilder();
+
+        // file exists in test project root
+        builder.AddDataAPIBuilder("dab", port: 1234, configFilePaths: "./dab-config.json");
+    }
```
aaronpowell: For these would could look at the annotations of the resource and find the `ContainerMountAnnotation` and assert that it contains the right value(s).

tests/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder.Tests/ContainerResourceCreationTests.cs
```diff
@@ -49,4 +40,66 @@ public void AddDataAPIBuilderContainerDetailsSetOnResource()
...
+
+    [Fact]
+    public void AddDataAPIBuilderContainer_ValidFiles_NoEx()
+    {
+        IDistributedApplicationBuilder builder = DistributedApplication.CreateBuilder();
+
+        // both files exist in test project root
+        builder.AddDataAPIBuilder("dab", "./dab-config.json", "./dab-config-2.json");
+    }
```
aaronpowell, line 102: This should assert that there are multiple `ContainerMountAnnotation` annotations on the resource reflecting the config files.

aaronpowell (changes requested):

tests/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder.Tests/ContainerResourceCreationTests.cs
JerryNixon: I am not sure how to do that. Can you point me to the syntax?

tests/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder.Tests/ContainerResourceCreationTests.cs
JerryNixon: I am not sure how to do that. Can you point me to the syntax?

src/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder/DataApiBuilderHostingExtension.cs
JerryNixon: The only gotcha with this is that we can't assume a `With()` method is coming when we assume the default `./dab-config.json` file from the `AddDataAPIBuilder()` method. If we add a `With()` method, we can no longer auto-add a default configuration. See what I mean?

src/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder/DataApiBuilderHostingExtension.cs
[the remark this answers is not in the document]
JerryNixon: And, no, the name of the file is not accounted for anywhere in DAB.

src/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder/DataApiBuilderHostingExtension.cs
aaronpowell: No, but we could use a life cycle hook (I think that'd be the best way) to check if there is at least one config provided (look at the annotations) and if there wasn't any annotations provided for config we can set it as unhealth and not start.

Or work on the assumption that if no `params` are provided we still add the default `dab-config.json`, like it does now.

src/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder/DataApiBuilderHostingExtension.cs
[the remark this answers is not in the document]
aaronpowell: Isn't it? I thought that there was some logic to look for specific file names only, otherwise how does DAB know what to load?

tests/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder.Tests/ContainerResourceCreationTests.cs
aaronpowell: Check out https://github.com/CommunityToolkit/Aspire/blob/main/tests/CommunityToolkit.Aspire.Hosting.Azure.StaticWebApps.Tests/ResourceCreationTests.cs#L54-L80

We build the distributed app and then you'd get the DAB resource from the app model and it exposes a method to access the annotations.

aaronpowell:
@JerryNixon do you think this is something we can tackle before the 13th? I'd like to have it in the stable DAB release.

src/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder/DataApiBuilderHostingExtension.cs
aaronpowell: Had a play around on implementing this and for now I'll drop the request as it introduced a bit of complexity with managing the config files.

As Jerry points out, you are going to need a config and you can't assume that one is provided. We could have some checks and stuff in place, but they'd become runtime checks which means you wouldn't know until runtime that you're missing a config file.

If someone wants to do a conditional add, they'll have to build a collection and pass that to the `params` instead.

aaronpowell (approved):

aaronpowell:
Looks like all tests are passing. @JerryNixon @FBoucher I'm happy with this to merge - if either of you wants to give it a once over and click the button go ahead!

src/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder.csproj
```diff
@@ -3,4 +3,10 @@
     <Description>An Aspire component leveraging the Data API Builder container.</Description>
     <AdditionalPackageTags>api dataapibuilder hosting</AdditionalPackageTags>
   </PropertyGroup>
+
+  <ItemGroup>
+    <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
+      <_Parameter1>CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder.Tests</_Parameter1>
+    </AssemblyAttribute>
```
Alirexaa, line 10:
```suggestion
    <InternalsVisibleTo Include="CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder.Tests"></InternalsVisibleTo>
```

PR #203, opened 2024-11-06
Ollama resource commands

### LINKED ISSUE #127: Ollama hosting component should expose resource commands for model management

Yes

### Existing integration

Ollama hosting

### Overview

Following on from this discussion [here](https://github.com/CommunityToolkit/Aspire/pull/122#discussion_r1809808146) it would be useful if as part of the Aspire 9 release the Ollama hosting component exposed some [resource commands](https://learn.microsoft.com/dotnet/aspire/whats-new/dotnet-aspire-9-release-candidate-1?tabs=unix&pivots=visual-studio#resource-commands) that would allow a user to manage the models in their Ollama instance.

### Usage example

- [ ] Re-download a model
- [ ] Delete a model

### Breaking change?

No

### Alternatives

Using the ollama API or Open WebUI can achieve this (to a degree)

### Additional context

_No response_

### Help us help you

aaronpowell:
This pull request introduces several updates to the Ollama hosting functionality, including enhancements to model management and the addition of a web UI. The most significant changes involve refactoring model-related methods, adding lifecycle hooks, and introducing new extension methods for easier configuration.

Fixes #127 

### Enhancements to model management:
* Refactored `Program.cs` to use `WithModel` and `WithHuggingFaceModel` methods for adding models to the Ollama container.
* Introduced `OllamaResourceBuilderExtensions.Model.cs` with methods `WithModel` and `WithHuggingFaceModel` for adding models, along with commands for managing model lifecycle such as redownloading, deleting, and stopping models.

### Lifecycle management improvements:
* Replaced `_notificationService` with `notificationService` in `OllamaModelResourceLifecycleHook.cs` to align with dependency injection practices. [[1]](diffhunk://#diff-2846156cb078751f632d5c772b2ec8afe55e96e8c7afc9bfa13bf8805ad43dc3L46-R45) [[2]](diffhunk://#diff-2846156cb078751f632d5c772b2ec8afe55e96e8c7afc9bfa13bf8805ad43dc3L65-R64) [[3]](diffhunk://#diff-2846156cb078751f632d5c772b2ec8afe55e96e8c7afc9bfa13bf8805ad43dc3L79-R78) [[4]](diffhunk://#diff-2846156cb078751f632d5c772b2ec8afe55e96e8c7afc9bfa13bf8805ad43dc3L89-R92)
* Removed the `PullModel` method from `OllamaModelResourceLifecycleHook.cs` and replaced its functionality with `OllamaUtilities.PullModelAsync`.

### Web UI integration:
* Added `OllamaResourceBuilderExtensions.OpenWebUI.cs` to provide methods for integrating the Open WebUI with the Ollama container, including `WithOpenWebUI` and `WithDataVolume`.

### Code cleanup:
* Removed unused `OllamaSharp.Models` import from `OllamaModelResourceLifecycleHook.cs`.
* Updated `OllamaResourceBuilderExtensions.cs` to be a partial class, facilitating the separation of different extension methods into multiple files.


Mrxx99:
From what I have seen from aspire, there is a convetion: `WithX` means returning the same builder, while `AddX` returns a new builder for the created resource. Like for postgres `AddDatabase`. So from that wouldn't it make more sense to keep them as `AddModel`?

aaronpowell:
> From what I have seen from aspire, there is a convetion: `WithX` means returning the same builder, while `AddX` returns a new builder for the created resource. Like for postgres `AddDatabase`. So from that wouldn't it make more sense to keep them as `AddModel`?

Actually, that's a good point. It was a change recommended by the Aspire team but now that you mention it it's not following the existing pattern.

aaronpowell:
Just re-read the thread and I misunderstood the comment on the API that I'd received. I'll revert it back to `AddModel`.

src/CommunityToolkit.Aspire.Hosting.Ollama/OllamaResourceBuilderExtensions.OpenWebUI.cs
```diff
@@ -0,0 +1,83 @@
...
+using Aspire.Hosting.Utils;
+using CommunityToolkit.Aspire.Hosting.Ollama;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Aspire.Hosting;
+
+public static partial class OllamaResourceBuilderExtensions
+{
+    /// <summary>
+    /// Adds an administration web UI Ollama to the application model using Attu. This version the package defaults to the main tag of the Open WebUI container image
```
Alirexaa, line 11:
```suggestion
    /// Adds an administration web UI Ollama to the application model using Open WebUI. This version the package defaults to the main tag of the Open WebUI container image
```

examples/deno/vite-demo/deno.lock
Alirexaa, line 1: I wonder why this file changed.

Alirexaa (commented):
Great work.
I need to run this code and see how it works.

examples/deno/vite-demo/deno.lock
aaronpowell: I think that it was caused by the vscode extension for Deno.

I'm considering disabling that extension as I'm finding a few annoyances from it.

src/CommunityToolkit.Aspire.Hosting.Ollama/OllamaModelHealthCheck.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        if (!success || endpoint is null)
+        {
+            return HealthCheckResult.Unhealthy("Invalid connection string");
+        }
+
+        var ollamaClient = new OllamaApiClient(endpoint);
+
+        // this will only return once the model is downloaded
+        await ollamaClient.ShowModelAsync(modelResource.ModelName, cancellationToken);
```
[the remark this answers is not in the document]
aaronpowell: I feel like health checks are going to be inherently hard to test due to the complex nature of the state that needs to be managed for Ollama and the size of the various Docker images we're going to be relying on.

The models in our sample results in a 2.8gb volume, combine that with the 3.28gb and 4.2gb container images for Ollama and Open WebUI, it's the better part of 10gb to just run these tests as they stand. I _think_ this is part of the underlying problem as to why the tests for Ollama aren't working anymore and I'm hesitant to add more integration tests as a result.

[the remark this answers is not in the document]
aaronpowell: Ok, I've fixed the test hang as part of https://github.com/CommunityToolkit/Aspire/pull/203/commits/8da2350e82cca5e4d2949b1e7c4170d20015888a

I understand why the test was failing (that was the result of the refactor in the PR to move to using `KnownResourceStates` rather than our own custom `Ready` for the model), and I _think_ I understand why the timeout didn't trigger it to fail but I'm not sure it's something that is easily solvable in our project (nor should it be).

For doing the unhealthy and healthy checks, well, we do test for healthy indirectly by waiting for the resource to go to a `KnownResourceStates.Running`, but we could change that to `WaitForResourceHealthyAsync`.

With Unhealthy testing, we'd have to write a test that does `WaitForResourceAsync` with a callback looking at the `resourceEvent.Snapshot.HealthStatus` and then pass if we find `Unhealthy`. But this runs the risk of an order-of-execution problem, if it becomes healthy before it hits that event registration then we may never see the event and it'll cause the problem that I was just debugging (I _think_... I'm assuming that the events aren't buffered, which they may be).

aaronpowell:
@Alirexaa let's commit with the tests disabled and I'll create a separate PR to try and work through them as they are failing across a lot of branches.

Alirexaa (approved):

PR #204, opened 2024-11-06
Add support for SQL Server Database Projects

### LINKED ISSUE #178: SQL Server Database Projects support
### .NET Aspire issue link

https://github.com/dotnet/aspire/discussions/1696

### Overview

As discussed in #110 we would like to contribute a new integration that provides support for deploying SQL Server Database Projects as part of an Aspire AppHost project. This will allow developers to quickly iterate on their SQL Server Database schema during their inner loop for people that prefer schema-first design.

### Usage example

We are currently proposing something like this:

```csharp

var database = builder.AddSqlServer("sql")
                    .AddDatabase("TargetDatabase");

builder.AddSqlProject<Projects.MySqlServerDatabaseProject>("my-sqlserver-database-project")
       .WithReference(database);

builder.Build().Run();
```

Upon running the AppHost project this will automatically publish the data-tier application package (.dacpac) created from the referenced project `MySqlServerDatabaseProject` once the SQL Server is ready (ie. Healthy). Said project could be either one using [MSBuild.Sdk.SqlProj(https://github.com/rr-wfm/MSBuild.Sdk.SqlProj) or [Microsoft.Build.Sql](https://github.com/Microsoft/DacFx).

Alternatively, if only a data-tier application package is available, this could also be deployed directly be leveraging the `WithDacpac()` method like so:

```csharp

var database = builder.AddSqlServer("sql")
                    .AddDatabase("TargetDatabase");

builder.AddSqlProject("my-sqlserver-database-project")
       .WithDacpac("path/to/somedatabase.dacpac")
       .WithReference(database);

builder.Build().Run();
```

### Additional context

_No response_

### Help us help you

jmezach:
**Closes #178 

This adds a new hosting integration for deploying SQL Server Database Projects to a SqlServerDatabaseResource in an Aspire AppHost project. See #178 for all the details.

## PR Checklist

- [x] New integration
  - [x] Docs are written


tests/.DS_Store
ErikEJ, line 1: This could be removed

jmezach: Removed it

jmezach:
It seems that CI is failing with a timeout, but that appears to be on something I haven't touched in this PR

aaronpowell:
I think the OpenWebUI tests are having some problems. I'm looking at the presently (seems OpenWebUI is broke in `main` as well which would result in inconsistent test runs)

src/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects/SqlProjectPublishService.cs
```diff
@@ -0,0 +1,67 @@
...
+            if (!File.Exists(dacpacPath))
+            {
+                logger.LogError("SQL Server Database project package not found at path {DacpacPath}.", dacpacPath);
+                await _resourceNotificationService.PublishUpdateAsync(sqlProject,
+                    state => state with { State = new ResourceStateSnapshot("Failed", KnownResourceStateStyles.Error) });
+                return;
+            }
+
+            var connectionString = await target.ConnectionStringExpression.GetValueAsync(cancellationToken);
+            if (connectionString == null)
```
aaronpowell, line 41:
```suggestion
            if (connectionString is null)
```

src/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects/SqlProjectPublishService.cs
```diff
@@ -0,0 +1,67 @@
...
+        ResourceNotificationService resourceNotificationService, IDistributedApplicationEventing eventing, IServiceProvider serviceProvider)
+    {
+        _deployer = deployer ?? throw new ArgumentNullException(nameof(deployer));
+        _resourceLoggerService = resourceLoggerService ?? throw new ArgumentNullException(nameof(resourceLoggerService));
+        _resourceNotificationService = resourceNotificationService ?? throw new ArgumentNullException(nameof(resourceNotificationService));
+        _eventing = eventing ?? throw new ArgumentNullException(nameof(eventing));
+        _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
+    }
+
+    public async Task PublishSqlProject(SqlProjectResource sqlProject, SqlServerDatabaseResource target, CancellationToken cancellationToken)
```
aaronpowell, line 25: I think it'd be better to use `KnownResourceStates.FailedToStart` rather than `Failed` when there are errors in here, as then we're more consistent with how Aspire reports statuses.

src/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects/SqlProjectResource.cs
```diff
@@ -0,0 +1,38 @@
...
+/// <summary>
+/// Represents a SQL Server Database project resource.
+/// </summary>
+/// <param name="name">Name of the resource.</param>
+public sealed class SqlProjectResource(string name) : Resource(name)
+{
+    internal string GetDacpacPath()
+    {
+        var projectMetadata = Annotations.OfType<IProjectMetadata>().FirstOrDefault();
+        if (projectMetadata != null)
```
aaronpowell, line 14:
```suggestion
        if (projectMetadata is not null)
```

src/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects/SqlProjectResource.cs
```diff
@@ -0,0 +1,38 @@
...
+            if (string.IsNullOrWhiteSpace(targetPath))
+            {
+                targetPath = project.GetPropertyValue("TargetPath");
+            }
+
+            return targetPath;
+        }
+
+        var dacpacMetadata = Annotations.OfType<DacpacMetadataAnnotation>().FirstOrDefault();
+        if (dacpacMetadata != null)
```
aaronpowell, line 31:
```suggestion
        if (dacpacMetadata is not null)
```

src/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects/SqlProjectResource.cs
```diff
@@ -0,0 +1,38 @@
...
+
+/// <summary>
+/// Represents a SQL Server Database project resource.
+/// </summary>
+/// <param name="name">Name of the resource.</param>
+public sealed class SqlProjectResource(string name) : Resource(name)
+{
+    internal string GetDacpacPath()
+    {
+        var projectMetadata = Annotations.OfType<IProjectMetadata>().FirstOrDefault();
```
aaronpowell, line 13: I recommend using either `TryGetLastAnnotation` or `TryGetAnnotationsOfType` so that we're using the built-in methods and it would also mean you don't need to do a null check, instead you do a bool test:

```csharp
if (!this.TryGetLastAnnotation<IProjectMetadada>(out var projectMetadata)) {
  throw new InvalidOperationException(...);
}

// do logic when found
```

tests/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.Tests/AddSqlProjectTests.cs
```diff
@@ -0,0 +1,66 @@
...
+        
+        // Act
+        using var app = appBuilder.Build();
+        var appModel = app.Services.GetRequiredService<DistributedApplicationModel>();
+        
+        // Assert
+        var sqlProjectResource = Assert.Single(appModel.Resources.OfType<SqlProjectResource>());
+        Assert.Equal("MySqlProject", sqlProjectResource.Name);
+
+        var annotation = Assert.Single(sqlProjectResource.Annotations.OfType<DacpacMetadataAnnotation>());
```
aaronpowell, line 42: Use the `TryGetLastAnnotation` or `TryGetAnnotationsOfType`, here's a test example: https://github.com/CommunityToolkit/Aspire/blob/9bb6f515f7b8b1e2e01bf856fd6e67de38d795c6/tests/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder.Tests/ContainerResourceCreationTests.cs#L47-L49

tests/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.Tests/AppHostTests.cs
```diff
@@ -0,0 +1,16 @@
+using CommunityToolkit.Aspire.Testing;
+using Aspire.Components.Common.Tests;
+using FluentAssertions;
+
+namespace CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.Tests;
+
+[RequiresDocker]
+public class AppHostTests(AspireIntegrationTestFixture<Projects.CommunityToolkit_Aspire_Hosting_SqlDatabaseProjects_AppHost> fixture) : IClassFixture<AspireIntegrationTestFixture<Projects.CommunityToolkit_Aspire_Hosting_SqlDatabaseProjects_AppHost>>
```
aaronpowell, line 8: Would it make sense to have a test which queries the database and verifies that the table exists? It's possible to access the connection string of the SQL database (see [these docs on testing](https://learn.microsoft.com/dotnet/aspire/testing/accessing-resources?pivots=mstest#access-other-resources)) and then provide that to a `SqlConnection` instance to then build out a query.

tests/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.Tests/AppHostTests.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    [Theory]
+    [InlineData("sdk-project")]
+    public async Task ResourceStartsAndRespondsOk(string resourceName)
+    {
```
aaronpowell, line 13:
```suggestion
    [Fact]
    public async Task ResourceStartsAndRespondsOk()
    {
        string resourceName = "sdk-project";
```

tests/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.Tests/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.Tests.csproj
```diff
@@ -0,0 +1,13 @@
...
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects\CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.csproj" />
+    <ProjectReference Include="..\CommunityToolkit.Aspire.Testing\CommunityToolkit.Aspire.Testing.csproj" />
+    <ProjectReference Include="..\..\examples\sql-database-projects\CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.AppHost\CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.AppHost.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="NSubstitute" />
+  </ItemGroup>
```
aaronpowell, line 11: I don't see us using NSubsititue anywhere, can we remove it as a dependency?

aaronpowell (changes requested):
Overall, it's looking pretty good. I've got some changes proposed to align with the patterns we have elsewhere (I think I need to tighten up some of the editorconfig stuff to make it more obvious/discoverable).

Also got a question around some additional integration testing that we can probably add here to give more confidence on successful runs.

src/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects/SqlProjectBuilderExtensions.cs
```diff
@@ -0,0 +1,110 @@
...
+    /// <summary>
+    /// Adds a SQL Server Database Project resource to the application based on a referenced MSBuild.Sdk.SqlProj project.
+    /// </summary>
+    /// <typeparam name="TProject">Type that represents the project that produces the .dacpac file.</typeparam>
+    /// <param name="builder">An <see cref="IDistributedApplicationBuilder"/> instance to add the SQL Server Database project to.</param>
+    /// <param name="name">Name of the resource.</param>
+    /// <returns>An <see cref="IResourceBuilder{T}"/> that can be used to further customize the resource.</returns>
+    public static IResourceBuilder<SqlProjectResource> AddSqlProject<TProject>(this IDistributedApplicationBuilder builder, string name)
+        where TProject : IProjectMetadata, new()
+    {
```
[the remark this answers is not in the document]
jmezach: Wouldn't we want to do a `ThrowIfNullOrWhitespace` on the `name` attribute?

jmezach:
@aaronpowell Thank you so much for the review, really appreciate it. I think I've processed all the comments.

src/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects/SqlProjectBuilderExtensions.cs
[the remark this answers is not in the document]
aaronpowell: No, and we probably don't even need to do the `ArgumentNullException` on the `name` param (but let's leave it for now), because the analyser will result in a compiler error (`ASPIRE006`) if your supplied name doesn't meet the requirements (ASCII letters, digits, and hyphens, greater than 1 character length and less than 64).

aaronpowell (approved):
Looks good - thanks for incorporating the feedback.

Once merged we'll get it on nuget and can you create docs using https://github.com/dotnet/docs-aspire/issues/new?template=05-community-toolkit-docs-request.yml as the issue template.

examples/sql-database-projects/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.AppHost/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.AppHost.csproj
```diff
@@ -0,0 +1,23 @@
...
+    <UserSecretsId>e290de9f-eab3-428b-bd01-f53b072f8406</UserSecretsId>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Aspire.Hosting.AppHost" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\src\CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects\CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.csproj" IsAspireProjectResource="False" />
+    <ProjectReference Include="..\SdkProject\SdkProject.csproj" />
```
aaronpowell, line 20: I just noticed in the build output this warning:

> Warning: /home/runner/.nuget/packages/aspire.hosting.apphost/9.0.0-rc.1.24511.1/build/Aspire.Hosting.AppHost.targets(178,5): warning ASPIRE004: '../SdkProject/SdkProject.csproj' is referenced by an Aspire Host project, but it is not an executable. Did you mean to set IsAspireProjectResource="false"? [/home/runner/work/Aspire/Aspire/examples/sql-database-projects/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.AppHost/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.AppHost.csproj]

The problem here is that Aspire does validation of the projects that are referenced and if they are not executable projects, then it expects the `IsAspireProjectResource` to be set to `false`.

Here's where the warning comes from: https://github.com/dotnet/aspire/blob/main/src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets#L178-L180

My _assumption_ is that Aspire doesn't want to generate any classes under `Projects` (like `Project.MyWebApp`) for anything that it isn't able to "start", which are executable projects, so it raises the error.

Given that it does make sense for us to be using the generated `Project` here for a clean usage on the `AddSqlProject` method I want to bring this up to the Aspire team to get their input.

aaronpowell (changes requested):
Just putting it back to pending review while following up that last comment I left

examples/sql-database-projects/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.AppHost/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.AppHost.csproj
ErikEJ: This is also noted in the readme

aaronpowell: I've had a chat with one of the Aspire folks today and yeah, kind of no "solution" for it at present. I'll give it another 24 hours (since I posted it after hours US) but we'll have to accept that this warning will exist for the time being I believe.

Unless someone comes up with some crazy solution by tomorrow I'll be merging this PR as the integration looks slick.

aaronpowell:
@jmezach @ErikEJ are you two happy to be the maintainers of this integration long term? If so, I'll add you to the team and set you up as the code owners.

ErikEJ:
Absolutely!

jmezach:
@aaronpowell Most definitely.

I was also thinking we should maybe add a `.sqlprojx` and the old `.sqlproj` format projects to the examples and incorporate those into the integration tests. But I guess we can do a follow up PR for that later.

ErikEJ:
I wonder if .sqlprojx is just a point in time extension? Maybe @sysadmindrew knows?

 But definitely ADS / VS code.sqlproj examples (ADS supports two .sqlproj formats)

aaronpowell:
Yeah let's add those as additional PR's, that way we have a known good baseline, and anything else is additional

dzsquared:
Re:.sqlprojx file extension
this extension was only for 17.12 preview 2, by the time preview 3 was released .sqlprojx was replaced with .sqlproj using Microsoft.Build.Sql

ErikEJ:
@dzsquared never knew that! Great news - now we just need side by side / migration tool/docs

aaronpowell (approved):

ErikEJ:
@aaronpowell When will the NuGet package get published?

aaronpowell:
Hopefully soon, we have a few manual tasks that have to be done before they are on nuget.

PR #210, opened 2024-11-09
SQL Database Projects codeowner

aaronpowell:
Adding @ErikEJ @jmezach as codeowners for the database projects integration


aaronpowell:
@ErikEJ @jmezach you should have both received an invite to the GitHub team, once accepted the CODEOWNERS file should be valid and we can merge this.

Alirexaa (approved):

PR #215, opened 2024-11-11
Fixing package versioning

### LINKED ISSUE #214: NuGet package versioning broken in PR and main
With https://github.com/CommunityToolkit/Aspire/commit/9d25d14a5e2fac1f685ec139cb4fa388bcf383ba#diff-4da790e243fabc7a08eef8f04f251cff1a708fd1aa498a6508da5ecb2bce1564R14 the versioning of packages is broken on the `main` and `release` branches because the `VersionPrefix` isn't being appropriately applied, and since it's not applied (only `VersionSuffix` via the CLI) we end up with `1.0.0-` as the prefix.

This means that we're getting invalid versions published on the NuGet and Azure Pipeline feeds.

aaronpowell:
Fixes #214


aaronpowell:
@CommunityToolkit/aspire-contributors package versioning looks to be fixed with this if someone wants to review and merge then push this up to other branches. I'll clean up NuGet and Azure Pipelines tomorrow (manual task)

Alirexaa (approved):

Issue #213, opened 2024-11-11
Ollama test stability


aaronpowell:
We disabled tests in #203 for some of the Ollama integration work.

This is tracking the fact they are disabled and what we need to do to re-enable them.

aaronpowell:
Fixed with #222

PR #222, opened 2024-11-12
Test stability

### LINKED ISSUE #220: flaky `CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.Tests.AppHostTests.ResourceStartsAndRespondsOk`
build : https://github.com/CommunityToolkit/Aspire/actions/runs/11777837131/job/32803012099

trx: [test-results-ubuntu-latest.zip](https://github.com/user-attachments/files/17704026/test-results-ubuntu-latest_2.zip)

Logs:
```
CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.Tests.AppHostTests.ResourceStartsAndRespondsOk [FAIL]
  Failed CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.Tests.AppHostTests.ResourceStartsAndRespondsOk [5 m]
  Error Message:
   System.TimeoutException : The operation has timed out.
  Stack Trace:
     at CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.Tests.AppHostTests.ResourceStartsAndRespondsOk() in /_/tests/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects.Tests/AppHostTests.cs:line 14
--- End of stack trace from previous location ---
```

aaronpowell:
It looks like the root cause of our test instability was that we were exceeding the disk space of the GitHub Actions runners (which is 14gb).

The main culprit here looks to be Ollama, which is a ~3gb image, then we were pulling two models, each 2gb or greater, then Open WebUI, which is ~4gb. Because of this, I've created a separate app host which we use for the tests that only contains the `tinyllama` model, and that reduces the amount of disk space we're using to around 4gb.

I've also removed the (sort of) duplicate Ollama tests, and instead leaned more heavily into using unit tests to assert we are providing Aspire with the right metadata.

For the SWA tests, it looks like they had a really short timeout which _may_ be their problem, and they weren't waiting for a Healthy status, just Running, so I've refactored to use that.

Fixes #220, #213


ErikEJ (approved):
LGTM, not sure I fully understand all the Ollama changes

PR #227, opened 2024-11-12
Silent vulnerable package warning

Alirexaa:
**Closes #<ISSUE_NUMBER>**

## PR Checklist

- [ ] New integration
  - [ ] Docs are written

### DIFF
--- a/examples/sql-database-projects/SdkProject/SdkProject.csproj
+++ b/examples/sql-database-projects/SdkProject/SdkProject.csproj
@@ -1,9 +1,10 @@
-<Project Sdk="MSBuild.Sdk.SqlProj/2.9.1">
+<Project Sdk="MSBuild.Sdk.SqlProj/3.0.0-beta.2">
     <PropertyGroup>
         <TargetFramework>netstandard2.1</TargetFramework>
         <SqlServerVersion>Sql150</SqlServerVersion>
         <RunSqlCodeAnalysis>True</RunSqlCodeAnalysis>
         <!-- For additional properties that can be set here, please refer to https://github.com/rr-wfm/MSBuild.Sdk.SqlProj#model-properties -->
+        <NoWarn>SQLPROJ0001</NoWarn>
     </PropertyGroup>
 
     <ItemGroup>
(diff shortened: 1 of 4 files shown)


Alirexaa:
@aaronpowell,

following error case build fail:
```
/home/runner/.nuget/packages/msbuild.sdk.sqlproj/2.9.1/Sdk/Sdk.targets(100,5): error : The current SDK bundles target framework '9.0' which is not supported by MSBuild.Sdk.SqlProj. Either install a .NET Core SDK that supports 3.1, 5.0 or 6.0, or file an issue at https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/ to add support. If you have preview SDKs installed, you may have to use a global.json file: https://docs.microsoft.com/dotnet/core/tools/global-json [/home/runner/work/Aspire/Aspire/examples/sql-database-projects/SdkProject/SdkProject.csproj]
```
Any idea?

aaronpowell:
@ErikEJ @jmezach do we need to upgrade SQL SDK?

aaronpowell:
Looks like we need to upgrade to the 3.0 release of the SQL SDK. This is because we build with .NET 9 on the runners and now it's not preview it is the primary .NET SDK that is used, but the 2.9.1 SQL SDK doesn't support .NET 9 and thus errors.

I've done that upgrade as looking at the issues it seems that 3.0 has just been waiting for the .NET 9 GA to drop the `-beta.2` tag.

aaronpowell (approved):

ErikEJ:
@aaronpowell yes, we will release 3.0.0 very soon

aaronpowell:
Once it's done and the package is published can you update the project file to drop `-beta.2` please?

I didn't dig into it completely but I'm assuming it's just a "quirk" of how the SDK Project types are handled with respect to .NET versions.

Also, I noticed that the 3.0.0 package is only .NET 9, would it not make more sense to use multiple TFM's?

ErikEJ:
@aaronpowell It is a bit more complex, we have a command line tool that targets multiple .NET versions, called by the Build SDK . Where do you see NET 9.0 only?

aaronpowell:
ah wait, my bad, I thought the nupkg was only .NET 9 but no, I must've been thinking of something else (I've been inside soooooo many nupkgs today 🤣)

ErikEJ:
@aaronpowell PR created: https://github.com/CommunityToolkit/Aspire/pull/232

PR #228, opened 2024-11-12
Adding .NET 9 to the devcontainer

aaronpowell:
This will help with local testing against .NET 9, since we use it on CI


aaronpowell:
This will help catch .NET 9 compile issues locally too

Alirexaa (approved):

PR #260, opened 2024-11-18
Fixing `WaitFor` support with Ollama model downloads

### LINKED ISSUE #256: Waiting on Ollama model resources blocks download
### Describe the bug

When using the `WaitFor` API on a resource that depends on the model resource from Ollama, the model will never be downloaded.

### Regression

_No response_

### Steps to reproduce

```text
1. Add `WaitFor(llama)` on the Ollama demo app
1. Launch the sample
1. Observe the model never downloading
```

### Expected behavior

The model should download and the resource waiting for it shouldn't start until the model is ready.

### Screenshots

_No response_

### IDE and version

Other

### IDE version

_No response_

### Nuget packages

- [ ] CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder
- [ ] CommunityToolkit.Aspire.Hosting.Azure.StaticWebApps
- [ ] CommunityToolkit.Aspire.Hosting.Deno
- [ ] CommunityToolkit.Aspire.Hosting.Golang
- [ ] CommunityToolkit.Aspire.Hosting.Java
- [ ] CommunityToolkit.Aspire.Hosting.Meilisearch
- [ ] CommunityToolkit.Aspire.Hosting.NodeJS.Extensions
- [x] CommunityToolkit.Aspire.Hosting.Ollama
- [ ] CommunityToolkit.Aspire.Meilisearch
- [ ] CommunityToolkit.Aspire.OllamaSharp

### Nuget package version(s)

9.0.0

### Additional context

This is because the lifecycle hook relies on `AfterResourcesCreatedAsync` which doesn't execute until after the resources have started, but since we have a resource waiting on health from the model resource, it doesn't start, resulting in a race condition.

Instead, we should use the new eventing API and the `ResourceReadyEvent`.

### Help us help you

aaronpowell:
**Closes #256**

## PR Checklist

- [ ] New integration
  - [ ] Docs are written

## Other information

The lifecycle hook would get blocked if you used WaitFor on the model resource, as it wouldn't fire until the model was healthy, which it couldn't achieve. Refactored to use an event instead that relies on the server resource ready event.

Had to turn the event into a non-blocking with Task.Run otherwise models couldn't be downloaded concurrently, as the events would run one at a time, waiting for model download completion before continuing on.

Also took this time to refactor the AddModel method to have a few extension methods, making it a bit clearer as to what happens when.


src/CommunityToolkit.Aspire.Hosting.Ollama/OllamaResourceBuilderExtensions.Model.cs
```diff
@@ -142,39 +141,136 @@ private static OllamaModelResource AddModelResourceCommand(
...
     {
-        modelResource.Annotations.Add(new ResourceCommandAnnotation(
-            name: name,
-            displayName: displayName,
-            updateState: context =>
-                context.ResourceSnapshot.State?.Text == KnownResourceStates.Running ?
-                    ResourceCommandState.Enabled :
-                    ResourceCommandState.Disabled,
-            executeCommand: async context =>
+        builder.ApplicationBuilder.Eventing.Subscribe<ResourceReadyEvent>(builder.Resource.Parent, (@event, cancellationToken) =>
```
Alirexaa, line 178: Can we subscribe to the `ResourceReadyEvent` for a resource multiple times?
What happens when we have various models?

aaronpowell: Yes we can, they will fire in sequence though, which is why I use `Task.Run` to execute the download, spawning a background job to download the model so that the event handler becomes non-blocking (originally I didn't do that and it would be blocking, which is not ideal).

Alirexaa (approved):

PR #273, opened 2024-11-19
Feature/issue 253 add active mq hosting

### LINKED ISSUE #253: [Feature] Add ActiveMQ hosting
We should migrate the ActiveMQ hosting from this PR (https://github.com/dotnet/aspire/pull/4094) to the CommunityToolkit.

Minimal requirements:
- The protocol should be configurable (https://masstransit.io/ expects a protocol activemq when using a connection-string)

anoordover:
**Closes #253 **

## PR Checklist

- [X] New integration
  - [X] Docs are written

## Other information

I'm looking for feedback on my implementation for the ActiveMQ hosting.
Please advice me about next steps I need to make to complete this functionality.


Directory.Packages.props
```diff
@@ -11,6 +11,8 @@
     <PackageVersion Include="Aspire.Hosting.SqlServer" Version="$(AspireVersion)" />
     <!-- AspNetCore packages -->
     <PackageVersion Include="AspNetCore.HealthChecks.SqlServer" Version="8.0.2" />
+    <PackageVersion Include="MassTransit" Version="8.3.1" />
+    <PackageVersion Include="MassTransit.ActiveMQ" Version="8.3.1" />
```
aaronpowell, line 15: Move these under the `External Packages` comment

examples/activemq/CommunityToolkit.Aspire.Hosting.ActiveMQ.AppHost/CommunityToolkit.Aspire.Hosting.ActiveMQ.AppHost.csproj
```diff
@@ -0,0 +1,29 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <Sdk Name="Aspire.AppHost.Sdk" Version="$(AspireAppHostSdkVersion)" />
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
```
aaronpowell, line 7:
```suggestion
```

Remove TFM as we control that higher up

examples/activemq/CommunityToolkit.Aspire.Hosting.ActiveMQ.AppHost/CommunityToolkit.Aspire.Hosting.ActiveMQ.AppHost.csproj
```diff
@@ -0,0 +1,29 @@
...
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\src\CommunityToolkit.Aspire.Hosting.ActiveMQ\CommunityToolkit.Aspire.Hosting.ActiveMQ.csproj" IsAspireProjectResource="false" />
+    <ProjectReference Include="..\CommunityToolkit.Aspire.Hosting.ActiveMQ.MassTransit\CommunityToolkit.Aspire.Hosting.ActiveMQ.MassTransit.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Reference Include="Aspire.Hosting">
+      <HintPath>..\..\..\..\..\.nuget\packages\aspire.hosting\9.0.0\lib\net8.0\Aspire.Hosting.dll</HintPath>
+    </Reference>
+  </ItemGroup>
```
aaronpowell, line 27: Don't directly reference assemblies

examples/activemq/CommunityToolkit.Aspire.Hosting.ActiveMQ.AppHost/Program.cs
```diff
@@ -0,0 +1,17 @@
...
+
+var builder = DistributedApplication.CreateBuilder(args);
+
+var amq = builder.AddActiveMQ("amq", 
+        builder.AddParameter("user", "admin"),
+        builder.AddParameter("password", "admin", secret: true), 
+        61616, 
+        "activemq")
+        .PublishAsConnectionString()
+        .WithEndpoint(port: 8161, targetPort: 8161, name: "web", scheme: "http");
```
aaronpowell, line 11:
```suggestion
var username = builder.AddParameter("user", "admin");
var password = builder.AddParameter("password", "admin", secret: true);

var amq = builder.AddActiveMQ("amq", username, password, 61616, "activemq")
        .PublishAsConnectionString()
        .WithEndpoint(port: 8161, targetPort: 8161, name: "web", scheme: "http");
```

Improving readability

examples/activemq/CommunityToolkit.Aspire.Hosting.ActiveMQ.MassTransit/CommunityToolkit.Aspire.Hosting.ActiveMQ.MassTransit.csproj
```diff
@@ -0,0 +1,19 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
```
aaronpowell, line 4:
```suggestion
```

Remove TFM as we control that higher up

examples/activemq/CommunityToolkit.Aspire.Hosting.ActiveMQ.MassTransit/HelloWorldConsumer.cs
```diff
@@ -0,0 +1,19 @@
+using MassTransit;
+
+namespace CommunityToolkit.Aspire.Hosting.ActiveMQ.MassTransit;
+
+public class HelloWorldConsumer : IConsumer<Message>
```
aaronpowell, line 5: Use a primary constructor and remove the field to be more consistent across the codebase.

examples/activemq/CommunityToolkit.Aspire.Hosting.ActiveMQ.ServiceDefaults/CommunityToolkit.Aspire.Hosting.ActiveMQ.ServiceDefaults.csproj
```diff
@@ -0,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
```
aaronpowell, line 4:
```suggestion
```

Remove TFM as we control that higher up

src/CommunityToolkit.Aspire.Hosting.ActiveMQ/ActiveMQBuilderExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+    /// </remarks>
+    /// <param name="builder">The <see cref="IDistributedApplicationBuilder"/>.</param>
+    /// <param name="name">The name of the resource. This name will be used as the connection string name when referenced in a dependency.</param>
+    /// <param name="userName">The parameter used to provide the user name for the ActiveMQ resource. If <see langword="null"/> a default value will be used.</param>
+    /// <param name="password">The parameter used to provide the password for the ActiveMQ resource. If <see langword="null"/> a random password will be generated.</param>
+    /// <param name="port">The host port that the underlying container is bound to when running locally.</param>
+    /// <param name="scheme">The scheme of the endpoint, e.g. tcp or activemq (for masstransit). Defaults to tcp.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<ActiveMQServerResource> AddActiveMQ(this IDistributedApplicationBuilder builder,
+        string name,
```
aaronpowell, line 30:
```suggestion
        [ResourceName] string name,
```

The `ResourceNameAttribute` is used to enforce the naming convention via an analyzer.

src/CommunityToolkit.Aspire.Hosting.ActiveMQ/ActiveMQBuilderExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+        // don't use special characters in the password, since it goes into a URI
+        ParameterResource passwordParameter = password?.Resource
+                                              ?? ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(builder, $"{name}-password", special: false);
+
+        ActiveMQServerResource activeMq = new(name, userName?.Resource, passwordParameter, scheme);
+        IResourceBuilder<ActiveMQServerResource> activemq = builder.AddResource(activeMq)
+                              .WithImage(ActiveMQContainerImageTags.Image, ActiveMQContainerImageTags.Tag)
+                              .WithImageRegistry(ActiveMQContainerImageTags.Registry)
+                              .WithEndpoint(port: port, targetPort: 61616, name: ActiveMQServerResource.PrimaryEndpointName,
+                                  scheme: scheme)
```
aaronpowell, line 48:
```suggestion
                              .WithEndpoint(port: port, targetPort: 61616, name: ActiveMQServerResource.PrimaryEndpointName, scheme: scheme)
```

src/CommunityToolkit.Aspire.Hosting.ActiveMQ/ActiveMQBuilderExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+
+    /// <summary>
+    /// Adds a named volume for the config folder to a ActiveMQ container resource.
+    /// </summary>
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="name">The name of the volume. Defaults to an auto-generated name based on the application and resource names.</param>
+    /// <param name="isReadOnly">A flag that indicates if this is a read-only volume.</param>
+    /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<ActiveMQServerResource> WithConfVolume(this IResourceBuilder<ActiveMQServerResource> builder, string? name = null, bool isReadOnly = false)
+        => builder
```
aaronpowell, line 81:
```suggestion
    public static IResourceBuilder<ActiveMQServerResource> WithConfVolume(this IResourceBuilder<ActiveMQServerResource> builder, string? name = null, bool isReadOnly = false) =>
        builder
```

src/CommunityToolkit.Aspire.Hosting.ActiveMQ/ActiveMQBuilderExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+
+    /// <summary>
+    /// Adds a named volume for the data folder to a ActiveMQ container resource.
+    /// </summary>
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="name">The name of the volume. Defaults to an auto-generated name based on the application and resource names.</param>
+    /// <param name="isReadOnly">A flag that indicates if this is a read-only volume.</param>
+    /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<ActiveMQServerResource> WithDataVolume(this IResourceBuilder<ActiveMQServerResource> builder, string? name = null, bool isReadOnly = false)
+        => builder
```
aaronpowell, line 66:
```suggestion
    public static IResourceBuilder<ActiveMQServerResource> WithDataVolume(this IResourceBuilder<ActiveMQServerResource> builder, string? name = null, bool isReadOnly = false) =>
        builder
```

src/CommunityToolkit.Aspire.Hosting.ActiveMQ/ActiveMQBuilderExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+    /// Adds a bind mount for the data folder to a ActiveMQ container resource.
+    /// </summary>
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="source">The source directory on the host to mount into the container.</param>
+    /// <param name="isReadOnly">A flag that indicates if this is a read-only mount.</param>
+    /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<ActiveMQServerResource> WithDataBindMount(this IResourceBuilder<ActiveMQServerResource> builder, string source, bool isReadOnly = false)
+        => builder.WithBindMount(source,
+            "/opt/apache-activemq/data",
+            isReadOnly);
```
aaronpowell, line 98:
```suggestion
    public static IResourceBuilder<ActiveMQServerResource> WithDataBindMount(this IResourceBuilder<ActiveMQServerResource> builder, string source, bool isReadOnly = false) =>
        builder.WithBindMount(source, "/opt/apache-activemq/data", isReadOnly);
```

src/CommunityToolkit.Aspire.Hosting.ActiveMQ/ActiveMQBuilderExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+    /// Adds a bind mount for the conf folder to a ActiveMQ container resource.
+    /// </summary>
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="source">The source directory on the host to mount into the container.</param>
+    /// <param name="isReadOnly">A flag that indicates if this is a read-only mount.</param>
+    /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<ActiveMQServerResource> WithConfBindMount(this IResourceBuilder<ActiveMQServerResource> builder, string source, bool isReadOnly = false)
+        => builder.WithBindMount(source,
+            "/opt/apache-activemq/conf",
+            isReadOnly);
```
aaronpowell, line 110:
```suggestion
    public static IResourceBuilder<ActiveMQServerResource> WithConfBindMount(this IResourceBuilder<ActiveMQServerResource> builder, string source, bool isReadOnly = false) =>
        builder.WithBindMount(source, "/opt/apache-activemq/conf", isReadOnly);
```

aaronpowell (changes requested):
Looks like a nice, solid starting point. There are some code style improvements I've put in as suggestions.

Also, can you add some integration tests in the pattern of https://github.com/CommunityToolkit/Aspire/blob/main/tests/CommunityToolkit.Aspire.Hosting.Meilisearch/AppHostTests.cs to verify that the hosting integration can start and that it will be able to be used.

Lastly, should there be some health checks in place?

examples/activemq/CommunityToolkit.Aspire.Hosting.ActiveMQ.MassTransit/HelloWorldConsumer.cs
anoordover: Resolved?

src/CommunityToolkit.Aspire.Hosting.ActiveMQ/ActiveMQServerResource.cs
```diff
@@ -0,0 +1,59 @@
...
+    public ActiveMQServerResource(string name, ParameterResource? userName, ParameterResource password,
+        string scheme) : base(name)
+    {
+        _scheme = scheme;
+        ArgumentNullException.ThrowIfNull(password);
+
+        PrimaryEndpoint = new EndpointReference(this, PrimaryEndpointName);
+        UserNameParameter = userName;
+        PasswordParameter = password;
+    }
```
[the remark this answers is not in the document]
aaronpowell: You can use a private field and use the [null-coalescing assignment (aka `??=`) operator](https://learn.microsoft.com/dotnet/csharp/language-reference/operators/null-coalescing-operator) like this:

```csharp
public class ActiveMQServerResource(.....)
{
    private EndpointReference? _primaryEndpoint;

    public EndpointReference PrimaryEndpoint => _primaryEndpoint ??= new(this, PrimaryEndpointName);

....
```

aaronpowell:
> I'm looking into the healthcheck and I don't know how I could implement this. When port 8161 is exposed I can do a GET to a jolokia endpoint that returns a json result containing the text "good". See https://stackoverflow.com/questions/52500782/how-to-check-if-activemq-is-working-properly
> 
> Maybe I could also try to connect to the broker using the apache.NMS nuget package but I don't know if I should add this package.
> 
> b.t.w. I would expect that the container itself would tell if it is healthy or not...

The container healthy status would be part of it, if the container is running, then it's "healthy", if it's stopped then it's "unhealthy" (well, the resource is stopped so that's an implied unhealthy status).

But depending on how the app inside the container works, just having the container running may not be enough. What if the app crashes in the container but the container doesn't stop? it's "healthy" in the fact that it's running, but it's really unhealthy as you can't use the endpoint(s).

If there's a HTTP endpoint that you can probe, that should do, and there's a built-in HTTP health check feature, we use it here for example: https://github.com/CommunityToolkit/Aspire/blob/main/src/CommunityToolkit.Aspire.Hosting.Azure.StaticWebApps/SwaAppHostingExtension.cs#L55

tests/CommunityToolkit.Aspire.Hosting.ActiveMQ.Tests/AppHostTests.cs
```diff
@@ -0,0 +1,43 @@
...
+
+[RequiresDocker]
+public class AppHostTests(AspireIntegrationTestFixture<Projects.CommunityToolkit_Aspire_Hosting_ActiveMQ_MassTransit> fixture) : IClassFixture<AspireIntegrationTestFixture<Projects.CommunityToolkit_Aspire_Hosting_ActiveMQ_MassTransit>>
+{
+    [Fact]
+    public async Task ResourceStartsAndRespondsOk()
+    {
+        var resourceName = "masstransitExample";
+        await fixture.ResourceNotificationService.WaitForResourceHealthyAsync(resourceName).WaitAsync(TimeSpan.FromMinutes(2));
+        await Task.Delay(15000);
```
aaronpowell, line 15: We shouldn't put a `Task.Delay`, this suggests that there's something else we should be doing to report the status of the resource, possibly a health check that will then hold the `WaitForResourceHealthy` to complete.

aaronpowell (changes requested):

tests/CommunityToolkit.Aspire.Hosting.ActiveMQ.Tests/AppHostTests.cs
anoordover: I added a healthcheck to the MassTransit example application. I used this to wait for health. I also tried to implement a unittest to check if a message-received counter is increased or not. I needed to add a retry in this unittest. Is this acceptable?

After this I will try to implement a HealthCheck on the ActiveMQ host. Maybe I will ask for advice.

anoordover:
I use this comment to share my ideas about the health-check for the ActiveMQ hosting and try to help my thinking proces.
- ActiveMQ has a webconsole that is exposed locally on port 8161
- On this port there is a so called Jolokia service avaible that you can use to query the JVM
- As shown in [http calls](https://github.com/anoordover/CommunityToolkit-Aspire/blob/feature/issue_253_AddActiveMqHosting/examples/activemq/CommunityToolkit.Aspire.Hosting.ActiveMQ.MassTransit/CommunityToolkit.Aspire.Hosting.ActiveMQ.MassTransit.http) there is an endpoint the return a text that contains the text "good" when ActiveMQ is running
- This endpoint is secured with the same username/password as the primary endpoint that is used to connect to ActiveMQ

My ideas/questions:
- Should we expose this endpoint/port (8161) by default?
- I suppose I can reuse the same username/password when exposing this port. Am I correct?
- ...

anoordover:
I tried to add a healthcheck using the Jolokia endpoint on another branch: https://github.com/anoordover/CommunityToolkit-Aspire/tree/feature/issue_253_AddActiveMqHosting_WithHealth
Should we proceed with this solution?

src/CommunityToolkit.Aspire.Hosting.ActiveMQ/ActiveMQServerResource.cs
```diff
@@ -0,0 +1,53 @@
...
+
+/// <summary>
+/// A resource that represents a ActiveMQ resource.
+/// </summary>
+/// <param name="name">The name of the resource.</param>
+/// <param name="userName">A parameter that contains the ActiveMQ server username, or <see langword="null"/> to use a default value.</param>
+/// <param name="password">A parameter that contains the ActiveMQ server password.</param>
+/// <param name="scheme">Scheme used in the connectionString (e.g. tcp or activemq, see MassTransit)</param>
+public class ActiveMQServerResource(string name, ParameterResource? userName, ParameterResource password,
+    string scheme) : ContainerResource(name), IResourceWithConnectionString, IResourceWithEnvironment
```
aaronpowell, line 17:
```suggestion
public class ActiveMQServerResource(string name, ParameterResource? userName, ParameterResource password, string scheme) : ContainerResource(name), IResourceWithConnectionString, IResourceWithEnvironment
```

Let's be careful on the arbitrary line breaks

tests/CommunityToolkit.Aspire.Hosting.ActiveMQ.Tests/AppHostTests.cs
```diff
@@ -0,0 +1,50 @@
...
+        MessageCounter? oldMessageCounter = await httpClient.GetFromJsonAsync<MessageCounter>("/received");
+        oldMessageCounter.Should().NotBeNull();
+        
+        await httpClient.PostAsync("/send/Hello%20World", new StringContent(""));
+
+        MessageCounter? messageCounter = null;
+        for (int i = 0; i < 10; i++)
+        {
+            messageCounter = await httpClient.GetFromJsonAsync<MessageCounter>("/received");
+            await Task.Delay(1000);
```
aaronpowell, line 40: Is this to simulate a delay that might happen between messages going to the endpoint? I'm somewhat bothered by having `Task.Delay` in a test as it results in a forced duration of a test.

aaronpowell (changes requested):

aaronpowell:
> I tried to add a healthcheck using the Jolokia endpoint on another branch: https://github.com/anoordover/CommunityToolkit-Aspire/tree/feature/issue_253_AddActiveMqHosting_WithHealth Should we proceed with this solution?

Was looking at that code and thought "this looks familiar" 🤣

Yep, that looks right on how it could be done. I'd collapse a few of those methods together, since you don't really need to have something that's supporting varying parameters though.

anoordover:
So here are my next steps in my solution:
- rebased on latest changes from master;
- changed the MassTransit implementation to use a Request/Response pattern so a delay in the integrationtest isn't needed anymore;
- ported the healthcheck from the temporary branch to this branch;
- tried to simplify the healthcheck. Questions: Should I move the extension method to another class? Should I move the constants to the top of the file?

What are your best practices about line-breaks? Can I find this somewhere in the guidelines?

aaronpowell (approved):
Looking good there, I think it's ready to merge.

As I'm unfamiliar with this problem space, can you explain what (if any) the intersection between this and #278 would be? since I see MassTransit used in this PR.

anoordover:
To explain a little bit about this implementation:
1. I wanted to create an ActiveMQ aspire hosting package;
2. I wanted to be able to use it in MassTransit;
3. I wanted to be able to use a connectionstring when connecting to ActiveMQ from MassTransit. To be able to use ActiveMQ like this the scheme of the connectionString being exposed should be configurable because MassTransit expect "activemq" as the scheme in the connectionString while others might use something else.

Because of this I did this:
1. Created the ActiveMQ aspire hosting package where the "user" has the ability to configure the scheme of the connectionString being provided;
2. Created an example project to show that the communication with ActiveMQ works using MassTransit. I used MassTransit because I was able to implement a request/response "flow" easily so that Task.Delay wasn't needed in the integrationtests of the hosting package.

In short: The implementation only facilitates using MassTransit by exposing the configuration of the scheme being used in the connectionString.
MassTransit is only being used for the integration test.

anoordover:
I tried to do a rebase on main but I didn't succeed in merging this way. I did an ordinary merge from main to my branch.

PR #308, opened 2024-12-03
Adding copilot-instructions

aaronpowell:
Feature info: https://docs.github.com/en/copilot/customizing-copilot/adding-custom-instructions-for-github-copilot

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,26 @@
+## Coding Style
+
+-   Use spaces for indentation with four-spaces per level, unless it is a csproj file, then use two-spaces per level.
+-   For integrations use the following namespace rules:
+    -   If the integration is a hosting integration (starts with CommunityToolkit.Aspire.Hosting), extension methods should be placed in `Aspire.Hosting`.
+    -   If the integration is a client integration (starts with CommunityToolkit.Aspire), extension methods should be placed in `Microsoft.Extensions.Hosting`.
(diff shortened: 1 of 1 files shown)


.github/copilot-instructions.md
```diff
@@ -0,0 +1,26 @@
+## Coding Style
+
+-   Use spaces for indentation with four-spaces per level, unless it is a csproj file, then use two-spaces per level.
+-   For integrations use the following namespace rules:
+    -   If the integration is a hosting integration (starts with CommunityToolkit.Aspire.Hosting), extension methods should be placed in `Aspire.Hosting`.
+    -   If the integration is a client integration (starts with CommunityToolkit.Aspire), extension methods should be placed in `Microsoft.Extensions.Hosting`.
+    -   Use file-scoped namespaces.
+-   All public methods require doc comments
```
IEvangelist, line 8:
```suggestion
-   All public members require doc comments.
```

.github/copilot-instructions.md
```diff
@@ -0,0 +1,26 @@
...
+-   Prefer type declarations over `var`
```
IEvangelist, line 9: This seems contradicting to the code sample be shown below, as it uses `var`. Reworded a bit.

```suggestion
-   Prefer type declarations over `var` when the type isn't obvious.
```

aaronpowell: Do as I say and not as I do 🤣

PR #334, opened 2024-12-16
Add support for deploying dacpacs from referenced NuGet packages

### LINKED ISSUE #327: Support deploying .dacpac from referenced NuGet packages

Yes

### Existing integration

SqlDatabaseProjects

### Overview

A feature I would like to add to the existing SqlDatabaseProjects integration is the ability to publish a .dacpac from a NuGet package referenced by the AppHost project. At our company we already have an existing ecosystem of NuGet packages that contain a `.dacpac` and this is in fact one of the supported scenarios in [MSBuild.Sdk.SqlProj](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj). All these packages have essentially the same semantics, in the sense that the package contains a `tools` folder in which we can then find a `.dacpac` file that has the same name as the NuGet Package ID (ie. if the package is called `MyDatabasePackage.nupkg` there would be a `tools\MyDatabasePackage.dacpac` within that package). I'm not sure if [Microsoft.Build.Sql](https://github.com/Microsoft/DacFx) has the same semantics though. That being said, we could probably support custom paths fairly easily by providing an overload that would allow one to specify the path within the package.

### Usage example

I was thinking of something along the following lines:

```csharp

var server = builder.AddSqlServer("sql")
                    .AddDatabase("TargetDatabase");

builder.AddSqlPackage<Packages.SomeReferencedPackage>("some-referenced-package")
       .WithReference(server);

builder.Build().Run();
```

In case someone would want to customize the path to the `.dacpac` within the package we could provide an overload like so:

```csharp

var server = builder.AddSqlServer("sql")
                    .AddDatabase("TargetDatabase");

builder.AddSqlPackage<Packages.Microsoft_SqlServer_Dacpacs>("some-referenced-package", "tools/master.dacpac")
       .WithReference(server);

builder.Build().Run();
```

### Breaking change?

No

### Alternatives

We already support deploying a `.dacpac` directly from an arbitrary path using something like:

```csharp

var server = builder.AddSqlServer("sql")
                    .AddDatabase("TargetDatabase");

builder.AddSqlProject("some-referenced-package", "<full-path-to-package>.dacpac")
       .WithReference(server);

builder.Build().Run();
```

Although this works just fine, it will be hard for developers to figure out the right path to the `.dacpac` from a referenced NuGet package. These usually reside in the NuGet package cache, but will change every time the package reference is updated, make this code brittle.

### Additional context

The proposed solution is akin to the existing solution for referenced projects, but this mechanism doesn't currently exist for referenced packages within Aspire afaik. I haven't looked in detail yet at how Aspire currently does this for projects, but maybe something similar could be created for referenced packages as well. Perhaps this could be achieved by adding some MSBuild logic to CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects for the time being, which would generate a class similar to `Projects` that exposes `IPackageMetadata`? I would love some input from the Aspire folks on this.

/cc @ErikEJ

### Help us help you

jmezach:
**Closes #327**

This adds support to the SQL Database Projects integration to deploy a `.dacpac` from a referenced NuGet package. It works by including some MSBuild magic to create classes that inherit from the new `IPackageMetadata` interface. This interface exposes the Package ID, Version and the physical path to the package in the users NuGet package cache.

We then leverage this metadata at runtime to find the correct path to the `.dacpac` and use the existing infrastructure for deploying said `.dacpac` to a target SQL Server.

## PR Checklist

- [] New integration
  - [ ] Docs are written


jmezach:
@aaronpowell @ErikEJ I think this is ready for review now :).

ErikEJ (approved):
LGTM!

src/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects/SqlProjectBuilderExtensions.cs
```diff
@@ -57,6 +57,24 @@ public static IResourceBuilder<SqlProjectResource> AddSqlProject(this IDistribut
...
     }
 
+    /// <summary>
+    /// Adds a SQL Server Database Project resource to the application based on a referenced NuGet package.
+    /// </summary>
+    /// <typeparam name="TPackage">Type that represents the NuGet package that contains the .dacpac file.</typeparam>
+    /// <param name="builder">An <see cref="IDistributedApplicationBuilder"/> instance to add the SQL Server Database project to.</param>
+    /// <param name="name">Name of the resource.</param>
+    /// <returns>Am <see cref="IResourceBuilder{T}"/> that can be used to further customize the resource.</returns>
+    public static IResourceBuilder<SqlProjectResource> AddSqlPackage<TPackage>(this IDistributedApplicationBuilder builder, [ResourceName]string name)
```
aaronpowell, line 67:
```suggestion
    public static IResourceBuilder<SqlProjectResource> AddSqlPackage<TPackage>(this IDistributedApplicationBuilder builder, [ResourceName] string name)
```

aaronpowell (approved):
Only the slightest nitpick in formatting

src/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects/SqlProjectBuilderExtensions.cs
jmezach: Made the same change to the other methods as well.

jmezach:
@aaronpowell Do we publish NuGet packages from these PR's? I wanted to make sure that the NuGet package actually contains the MSBuild logic so this will work end-to-end in a real project ;)

jmezach:
@aaronpowell Never mind, I've downloaded the NuGet packages from the CI pipeline artifacts and added those to a sample project and it all seems to work as expected. So I guess we're good to go here :).

jmezach:
Now that I've tried this out on my own I'm wondering if we should make a change. Specifically related to the resource being returned by `AddSqlPackage`. This currently returns a `SqlProjectResource` which I guess can be confusing for some. But more importantly it is not easy to distinguish in code between different `SqlPackageResource`s (if we call it that). That kind of makes it hard to further customize those resources.

For example, one of the scenarios I'm thinking of using this for is for our "ConfigurationDatabase". This is a central databased used by many of our services which contains a registry of tenants and their associated database(s). So, what I would like to provide is something like this:

```csharp

var sqlServer = builder.AddSqlServer("sql");
var configurationDatabase = sqlServer.AddDatabase("ConfigurationDatabase");
var tenantDatabase1 = sqlServer.AddDatabase("Tenant1");
var tenantDatabase2 = sqlServer.AddDatabase("Tenant2");

builder.AddSqlPackage<Packages.ConfigurationDatabase>("ConfigurationDatabase-Schema")
       .WithReference(configurationDatabase)
       .WithTenantDatabase("Tenant1", tenantDatabase1)
       .WithTenantDatabase("Tenant2", tenantDatabase2)

builder.Build().Run();
```

Obviously this would require implementing a `WithTenantDatabase` extensions method. But currently I'd add have to make that an extension method on `IResourceBuilder<SqlProjectResource>`, meaning you could call this on any database. But that won't work, since the extension method will make assumptions about the underlying schema of the database.

So perhaps we should have `AddSqlPackage<TPackageMetadata>` return a `IResourceBuilder<SqlPackageResource<TPackageMetadata>>` instead of a `IResourceBuilder<SqlProjectResource>`?

aaronpowell:
> @aaronpowell Do we publish NuGet packages from these PR's? I wanted to make sure that the NuGet package actually contains the MSBuild logic so this will work end-to-end in a real project ;)

I see you found it, but also, they are published to the Azure Pipeline feed on PR (main will land on NuGet).

aaronpowell:
If there's the potential that you'd want to do different things depending on whether it's a Project or Package based setup, then I'd suggest that we want to return a different type.

But if it's possible that there would be methods that would make sense whether it's a Project or Package, should there be an interface that both `SqlPackageResource` and `SqlProjectResource` could implement (even if it's just an empty interface) so a method could have a constraint of `where TResource : ISqlDeployResource` (or whatever it gets called).

jmezach:
Technically I don't think there is a reason to differentiate between packages and projects. In the end both are based on a `.dacpac`, it is just a matter of where that `.dacpac` is being sourced from. So I guess it should have been a `DacpacResource` but that would require doing a breaking change at this point. Perhaps we could do this in a non-breaking way by adding an `IDacpacResource` interface that both `SqlProjectResource` and `SqlPackageResource` would implement?

My main point though was about being able to differentiate between multiple `SqlPackageResource`s in code. If you're sourcing multiple different NuGet packages with different schemas I think you'd want to be able to target the different resources in your project.

aaronpowell:
Looks like CodeQL hasn't run (and I don't know how to force it - that's going to be a 2025 task 🤣).

@jmezach is this good to merge otherwise?

jmezach:
@aaronpowell I just tried this out with the latest changes in my local environment and it seems to work like a charm. Now I can add extension methods on specific `SqlPackage<TPackage>` resources by targeting the specific `TPackage`.

So yeah I think we are all good here. Just one thing I was wondering though. The PublicAPI.Unshipped.txt file still contains some API's that I believe have already shipped in 9.1. Shouldn't those be in PublicAPI.Shipped.txt?

ErikEJ:
@jmezach Should the readme be updated and/or the docs (in other repo) ?

jmezach:
> @jmezach Should the readme be updated and/or the docs (in other repo) ?

Yeah, we should definitely update the docs for this. What repo are those in again?

ErikEJ:
https://github.com/dotnet/docs-aspire/blob/main/docs/community-toolkit/hosting-sql-database-projects.md - happy to do that, but unsure what to actually write 😄

jmezach:
@ErikEJ @aaronpowell Added some docs here: https://github.com/dotnet/docs-aspire/pull/2349

PR #348, opened 2024-12-24
Java app build extension

### LINKED ISSUE #339: Remove Java as a build dependency
Java is required to build due to how the Java integration sample is written. This makes it harder to contribute (since you're building multiple runtimes at once) but also is different to how other integrations work.

We should find a way to remove the dependency during the .NET build phase.

aaronpowell:
This uses an event to run the mvnw build (or it can be customised via options), so it is only used when the app host is running the resource

**Closes #339**

## PR Checklist

- [ ] New integration
  - [ ] Docs are written


src/CommunityToolkit.Aspire.Hosting.Java/JavaAppHostingExtension.Executable.cs
```diff
@@ -0,0 +1,160 @@
...
+            await mvnw.WaitForExitAsync(ct).ConfigureAwait(false);
+
+            if (mvnw.ExitCode != 0)
+            {
+                await notificationService.PublishUpdateAsync(javaAppResource, state => state with
+                {
+                    State = new($"mvnw exited with {mvnw.ExitCode}", KnownResourceStates.FailedToStart)
+                }).ConfigureAwait(false);
+
+                throw new InvalidOperationException($"mvnw build failed with exit code {mvnw.ExitCode}");
```
davidfowl, line 147: Why throw?

src/CommunityToolkit.Aspire.Hosting.Java/JavaAppHostingExtension.Executable.cs
```diff
@@ -0,0 +1,160 @@
...
+            }
+        });
+
+        return builder;
+    }
+
+    private static IResourceBuilder<JavaAppExecutableResource> WithJavaDefaults(
+        this IResourceBuilder<JavaAppExecutableResource> builder,
+        JavaAppExecutableResourceOptions options) =>
+        builder.WithOtlpExporter()
```
davidfowl, line 157: Does this work without making https work?

src/CommunityToolkit.Aspire.Hosting.Java/JavaAppHostingExtension.Executable.cs
aaronpowell: It's how I've seen other hooks implemented in the past.

src/CommunityToolkit.Aspire.Hosting.Java/JavaAppHostingExtension.Executable.cs
aaronpowell: Without HTTPS on the dashboard?

davidfowl: Yea does this work right now?

src/CommunityToolkit.Aspire.Hosting.Java/JavaAppHostingExtension.Executable.cs
davidfowl: Where does this show up? Wouldn't it be better to show it in the dashboard? Also what happens when the build fails? Do I need to restart the entire apphost to rebuild or can I trigger it from the dashboard? (custom command)

Worth thinking through that I think.

src/CommunityToolkit.Aspire.Hosting.Java/JavaAppHostingExtension.Executable.cs
aaronpowell: I assume so 🤣

src/CommunityToolkit.Aspire.Hosting.Java/JavaAppHostingExtension.Executable.cs
aaronpowell: The notification service reporting shows up on the dashboard and I think the exception bubbles to the console of the resource that it was relative for.

A custom command would be a good idea to run a rebuild, I'll add that.

examples/java/CommunityToolkit.Aspire.Hosting.Java.AppHost/Program.cs
```diff
@@ -16,6 +16,7 @@
                                Port = 8085,
                                OtelAgentPath = "../../../agents",
                            })
+                           .WithMavenBuild()
```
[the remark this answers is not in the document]
aaronpowell: Probably best to add those others in another PR with related issues than overloading this one.

PR #352, opened 2024-12-30
Removing our version of PasswordGenerator.cs

aaronpowell:
We can access the functionality of this via the GenerateParameterDefault type, so we don't need to bring in the internal class to our codebase.

Reduces a dependency outlined here: https://github.com/dotnet/aspire/issues/6807


Alirexaa (approved):
LTGM, Thanks.

PR #365, opened 2025-01-03
Add SurrealDB hosting/client integration

Odonno:
**Closes #<ISSUE_NUMBER>**

## PR Checklist

- [ ] New integration
  - [ ] Docs are written


aaronpowell:
Having a look into the test failure, and it's because they are timing out, which made me think that the problem is to do with the container not starting. This is... sort of true, the container used for the tests does start, but it immediately exists because it's missing the `args` to tell Surreal to **actually** start.

In this block https://github.com/CommunityToolkit/Aspire/pull/365/files#diff-bae88f8ce8a5624289417d505a6ba0a190e20d70c801f6da997ae6121820b481R50-R56

You'll need to add `WithCommand("start", "memory")` (I don't think you need the entrypoint too) and then it gets onto the next place where the tests are failing.

The next place it seems to fail is that it can't resolve an `IHttpClientFactory`, which looking at the `SurrealDbClient` type is a required constructor argument.

My guess is that for the `ConfromanceTests` we will need to manually register that type (it'd be there in the API project, which is why it's not surfaced as an issue), and probably for the others in the SurrealDB client integration.

Looking at how you can register SurrealDB via their native SDK, I see you can pass a type argument that implements `ISurrealDbClient`, would it make sense to expose that in the integration too?

Odonno:
> You'll need to add `WithCommand("start", "memory")` (I don't think you need the entrypoint too) and then it gets onto the next place where the tests are failing.

I will make the change and see how it goes.

> The next place it seems to fail is that it can't resolve an `IHttpClientFactory`, which looking at the `SurrealDbClient` type is a required constructor argument.

Yes, it can take an `IHttpClientFactory` but it is optional (not `Required`). So, if not provided, it can be nullified and the client can work with the default `HttpClient`. I don't why there could be an issue with that. I will dig into it.

> Looking at how you can register SurrealDB via their native SDK, I see you can pass a type argument that implements `ISurrealDbClient`, would it make sense to expose that in the integration too?

Yes, it was there because KeyedService did not exist. I will get rid of this feature because Keyed Service is a better solution. So, no need to change anything for this.

aaronpowell:
> > The next place it seems to fail is that it can't resolve an `IHttpClientFactory`, which looking at the `SurrealDbClient` type is a required constructor argument.
> 
> Yes, it can take an `IHttpClientFactory` but it is optional (not `Required`). So, if not provided, it can be nullified and the client can work with the default `HttpClient`. I don't see why there could be an issue with that. I will dig into it.

I wasn't paying attention properly to the error I was seeing, it is when using the `AddKeyedSurrealClient` that I see this failure on, and that's because this method uses `GetRequestedService<IHttpClientFactory>`: https://github.com/surrealdb/surrealdb.net/blob/main/SurrealDb.Net/Extensions/DependencyInjection/ServiceCollectionExtensions.cs#L413-L427

If you don't register keyed, it uses `GetService<IHttpClientFactory>` which will return a `null` if the service wasn't registered. Looking at the history, this might just be an oversight, as I see it was changed in https://github.com/surrealdb/surrealdb.net/commit/a7a5a30c2cd00d3ff0e8bef84768a54fed880472, but obviously you're more familiar with the Surreal codebase than I am.

aaronpowell:
> Yes, it was there because KeyedService did not exist. I will get rid of this feature because Keyed Service is a better solution. So, no need to change anything for this.

Is it uncommon that someone would create their own type that implements `ISurrealDbClient`? Or is that more of a legacy approach?

aaronpowell (changes requested):
Overall, the code is looking pretty much ready to go, with the obvious exception of the tests needing to be fixed.

Odonno:
> > Yes, it was there because KeyedService did not exist. I will get rid of this feature because Keyed Service is a better solution. So, no need to change anything for this.
> 
> Is it uncommon that someone would create their own type that implements `ISurrealDbClient`? Or is that more of a legacy approach?

Legacy approach

Odonno:
> > > The next place it seems to fail is that it can't resolve an `IHttpClientFactory`, which looking at the `SurrealDbClient` type is a required constructor argument.
> > 
> > 
> > Yes, it can take an `IHttpClientFactory` but it is optional (not `Required`). So, if not provided, it can be nullified and the client can work with the default `HttpClient`. I don't see why there could be an issue with that. I will dig into it.
> 
> I wasn't paying attention properly to the error I was seeing, it is when using the `AddKeyedSurrealClient` that I see this failure on, and that's because this method uses `GetRequestedService<IHttpClientFactory>`: https://github.com/surrealdb/surrealdb.net/blob/main/SurrealDb.Net/Extensions/DependencyInjection/ServiceCollectionExtensions.cs#L413-L427
> 
> If you don't register keyed, it uses `GetService<IHttpClientFactory>` which will return a `null` if the service wasn't registered. Looking at the history, this might just be an oversight, as I see it was changed in [surrealdb/surrealdb.net@a7a5a30](https://github.com/surrealdb/surrealdb.net/commit/a7a5a30c2cd00d3ff0e8bef84768a54fed880472), but obviously you're more familiar with the Surreal codebase than I am.

I suppose bad copy/paste due to different commit changes. I will fix that.

Odonno:
I kind of struggle with the conformance tests (non keyed). It fails to read `Aspire:Surreal:Client` configuration because it expects a `string` (connection string) and I suppose it gets an `Options` type.

I think it would be easier/better to only deal with connection string, at least for DB integrations which is the case here. I don't know what is the best practice for this. Can you guide me?

Odonno:
Ok. I found the issue. I copied misleading code. :( Missing key in the default connection string.

aaronpowell:
Cool, looks like that's sorted most of the tests and we're just going to wait for https://github.com/surrealdb/surrealdb.net/pull/163 to be resolved to unblock most of the others.

tests/CommunityToolkit.Aspire.SurrealDb.Tests/AspireSurrealClientExtensionsTest.cs
```diff
@@ -0,0 +1,141 @@
...
+using SurrealDb.Net;
+
+namespace CommunityToolkit.Aspire.SurrealDb.Tests;
+
+public class AspireSurrealClientExtensionsTest(SurrealDbContainerFixture containerFixture) : IClassFixture<SurrealDbContainerFixture>
+{
+    private const string DefaultConnectionName = "db";
+
+    private string DefaultConnectionString =>
+            RequiresDockerAttribute.IsSupported ? containerFixture.GetConnectionString() : "http://localhost:27011";
```
aaronpowell, line 17: I think this needs to be changed to work on Windows, as it's providing just a raw endpoint rather than a formatted connection string.

So rather than `http://...` it should be `Endpoint=...`

aaronpowell:
@Odonno - I've done some digging into the problem that we've got in the tests, and it looks like it's caused because the `rocksdb` endpoint is failing with a permission error.

I don't think I can push directly to your fork, so I've created a patchfile with the changes that I've been experimenting with, you'll find it here: https://gist.github.com/aaronpowell/4ed1c55cb2feb556f9b0e18266767cc9 (copy the `diff.patch` to a local file and run `git apply <patch file>`). **Note** the tests still fail, they just fail faster and with slightly easier to observe errors.

Here's some things I changed:

* Using a `CancellationTokenSource` of 5 minutes and passing the token into all async methods (this means that the test has ~5 minutes to run, not an infinite amount of time).
* Removed the usage of `WaitForText` and using the built-in Aspire features, since there are health checks we can rely on.
* Specified a full file path, not just a directory for `rocksdb`
  * The path that `WithDataVolume` and `WithDataBindMount` then mount that path <-- I think this is the key problem, as it was trying to put the database at `/data` which it didn't have permission to (or at least, that seemed like a viable guess).

Hopefully this is a helpful next step for figuring out the underlying problem - it really seems to be related to how the mounted path is handled by the container for storing the db.

Odonno:
This is weird that my issue locally is with `ApiServiceStartsAndRespondsOk` and `SurrealResourceStartsAndRespondsOk` tests (`VerifySurrealDbResource` works fine) while in the CI it is the complete opposite.

Odonno:
Still have the same issue. @aaronpowell Do you have any idea why `ConnectionStringAvailableEvent` is never triggered? Is there any documentation about this event somewhere?

aaronpowell:
I don't know of any reason why it wouldn't trigger the event (esp why it would on local but not CI). I'll try and find some time to dig into it next week

Odonno:
Finally, all my tests are green now.

image

Odonno:
Still having an issue on Linux. Reading the logs, it seems like Aspire is trying to use the resource before it gets created or it tries to create a new one before the previous one was destroyed. Any magic idea @aaronpowell?

Odonno:
@aaronpowell Feels good when all tests are passing. :)

src/CommunityToolkit.Aspire.Hosting.SurrealDb/SurrealDbBuilderExtensions.cs
```diff
@@ -0,0 +1,418 @@
...
+            var options = new SurrealDbOptionsBuilder().FromConnectionString(connectionString).Build();
+            surrealDbClient = new SurrealDbClient(options);
+        });
+
+        string namespaceName = builder.Resource.Name;
+        string serverName = builder.Resource.Parent.Name;
+
+        string healthCheckKey = $"{serverName}_{namespaceName}_{name}_check";
+        // TODO : Bug to be fixed
+        //builder.ApplicationBuilder.Services.AddHealthChecks().AddSurreal(_ => surrealDbClient!, healthCheckKey);
```
aaronpowell, line 192: What's the bug to be fixed?

src/CommunityToolkit.Aspire.Hosting.SurrealDb/SurrealDbBuilderExtensions.cs
```diff
@@ -0,0 +1,418 @@
...
+                    writer.WriteString("defaultNamespace", uniqueNamespace.NamespaceName);
+                }
+                if (uniqueDatabase is not null)
+                {
+                    writer.WriteString("defaultDatabase", uniqueDatabase.DatabaseName);
+                }
+
+                writer.WriteStartObject("authentication");
+                writer.WriteString("protocol", "ws");
+                // How to do host resolution?
```
aaronpowell, line 392: The `EndpointReference` type has a `Host` property which should be the host of the allocated endpoint (since they are containers it'd be the container network reference IIRC). Or maybe you can use the container name, like https://github.com/dotnet/aspire/blob/d621f86daff51dcdbc8860858b9571ce64ecb44f/src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs#L548 does

aaronpowell (approved):
Looks good to merge. The comments aren't blocking that I left, more wanting to know more.

src/CommunityToolkit.Aspire.Hosting.SurrealDb/SurrealDbBuilderExtensions.cs
Odonno: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/2398

src/CommunityToolkit.Aspire.Hosting.SurrealDb/SurrealDbBuilderExtensions.cs
Odonno: It has to be the host because it is calling the surreal dependency not from the container but from the client. It works with `endpoint.Host`.

src/CommunityToolkit.Aspire.Hosting.SurrealDb/SurrealDbBuilderExtensions.cs
aaronpowell: Want to create a task to track fixing that when a new package is released?

Odonno: I don't think it's necessary. Just waiting for the release with .NET 10 I suppose.

PR #387, opened 2025-01-14
221 python uv

### LINKED ISSUE #221: Python uv and poetry modules
### .NET Aspire issue link

_No response_

### Overview

An Integration to support Python modules packed by uv and poetry.

### Usage example

```csharp
var pythonModule = builder.AddUvicornApp("ServiceName",
                                            projectDirectory: "path/to/module"),
                                            scriptPath: "moduleName",
                                            packageManager: "uv")
```

### Additional context

_No response_

### Help us help you

tommasodotNET:
**Closes #221**

## PR Checklist

- [x] New integration
  - [x] Docs are written


src/CommunityToolkit.Aspire.Hosting.Python.Extensions/UvAppHostingExtension.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                // If the project is to be automatically instrumented, add the instrumentation executable arguments first.
+                if (!string.IsNullOrEmpty(instrumentationExecutable))
+                {
+                    AddOpenTelemetryArguments(context);
+
+                    // // Add the python executable as the next argument so we can run the project.
+                    // context.Args.Add(pythonExecutable!);
+                }
+            });
```
aaronpowell, line 74: may as well add this to the below `if` block and just have it `resourceBuilder.WithArgs(AddOpenTelemetryArguments);` since the `if` block is duplicated.

aaronpowell (changes requested):
do we need to ensure that `uv` is installed as part of the CI job?

I think we might need to add it to the `post-create.sh` file for the devcontainer too.

src/CommunityToolkit.Aspire.Hosting.Python.Extensions/UvAppHostingExtension.cs
tommasodotNET: I was following the same logic found in https://github.com/dotnet/aspire/blob/d7e27b157b0266fc17a116115ba55d773684eba6/src/Aspire.Hosting.Python/PythonAppResourceBuilderExtensions.cs#L131

aaronpowell (approved):

src/CommunityToolkit.Aspire.Hosting.Python.Extensions/UvAppHostingExtension.cs
aaronpowell: ah right, let's leave it

PR #390, opened 2025-01-14
Disable Deno tests to unblock clean CI

Alirexaa:
**Closes #<ISSUE_NUMBER>**

## PR Checklist

- [ ] New integration
  - [ ] Docs are written


aaronpowell:
I'm not sure this is a good idea. Since Deno seems to be broken, disabling the tests gives a false indication that the integration is usable, when it appears not to be (the Deno sample fails to launch).

Alirexaa:
> I'm not sure this is a good idea. Since Deno seems to be broken, disabling the tests gives a false indication that the integration is usable, when it appears not to be (the Deno sample fails to launch).

I know that but other PRs CI failed due to this. We should unlock the PRs by disabling these tests just for now.

aaronpowell:
I've tagged you on #391 which is a better fix - even if it's also a temporary workaround.

Alirexaa:
Close favor of #391

Issue #377, opened 2025-01-10
Deno test timing out


aaronpowell:
https://github.com/CommunityToolkit/Aspire/actions/runs/12701892107/job/35407179902

One of the Deno tests has started timing out, need to investigate why.

aaronpowell:
Looks like this is failing because the `oak` demo app isn't starting via the DCP when Aspire launches.

Trying to figure out why that is, but it looks like something might have changed in the 2.1.15 release of Deno.

aaronpowell:
DCP logs:

```
warn: Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServer[9]
      The ASP.NET Core developer certificate is only trusted by some clients. For information about trusting the ASP.NET Core developer certificate, see https://aka.ms/aspnet/https-trust-dev-cert
Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServer: Warning: The ASP.NET Core developer certificate is only trusted by some clients. For information about trusting the ASP.NET Core developer certificate, see https://aka.ms/aspnet/https-trust-dev-cert
info: Aspire.Hosting.DistributedApplication[0]
      Aspire version: 9.0.0+01ed51919f8df692ececce51048a140615dc759d
Aspire.Hosting.DistributedApplication: Information: Aspire version: 9.0.0+01ed51919f8df692ececce51048a140615dc759d
Aspire.Hosting.DistributedApplication: Information: Distributed application starting.
info: Aspire.Hosting.DistributedApplication[0]
      Distributed application starting.
info: Aspire.Hosting.DistributedApplication[0]
      Application host directory is: /workspaces/CommunityToolkit-Aspire/examples/deno/CommunityToolkit.Aspire.Hosting.Deno.AppHost
Aspire.Hosting.DistributedApplication: Information: Application host directory is: /workspaces/CommunityToolkit-Aspire/examples/deno/CommunityToolkit.Aspire.Hosting.Deno.AppHost
warn: Microsoft.AspNetCore.DataProtection.Repositories.FileSystemXmlRepository[60]
      Storing keys in a directory '/home/vscode/.aspnet/DataProtection-Keys' that may not be persisted outside of the container. Protected data will be unavailable when container is destroyed. For more information go to https://aka.ms/aspnet/dataprotectionwarning
[...]

PR #391, opened 2025-01-14
Deno integration fix

aaronpowell:
With #377 we've identified a problem that means that CI is broken for the Deno integration.

Looking at the build logs it seems that Deno 2.1.5 (released in the past week) is causing failures.

This PR pins the version of 2.1.4 for Deno, which is the last known working version. Admittedly, this is a temporary workaround until we find the underlying problem that comes with Deno 2.1.5


Alirexaa (approved):
It seems a better approach than #390.

PR #428, opened 2025-01-29
Adding ability to configure extensions for Sqlite

aaronpowell:
Adding the ability to specify SQLite extensions that you want to load on the database. Extensions can be shipped as NuGet packages (https://www.nuget.org/packages/mod_spatialite is an example) or you can specify a path to the extension binary (an absolute path).

This is done in the app host and the info is projected as part of the connection string for the client.

In the client library, it'll read the extensions from the connection string (then delete that value so it is a valid connection string) and load the extensions using the SqliteConnection. Since SQLite needs the extension to be in PATH ([see this doc](https://learn.microsoft.com/dotnet/standard/data/sqlite/extensions)) there is a bit of code to ensure the path for the binary is loaded correctly.

I've added some tests for _observing_ that the extension loads, but it's a hard one to write tests for as you have to have a bunch of different binaries checked into the repo (I'd use mod_spatialite but it only ships Windows binaries in the nuget package). Because if this, I've marked the feature as experimental.

This doesn't support the EF SQLite integration as there's no way I can find using the EF API in which you can add extensions, you have to do it manually when the DbContext is created.

### DIFF
--- a/src/CommunityToolkit.Aspire.Hosting.Sqlite/SqliteResource.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.Sqlite/SqliteResource.cs
@@ -1,3 +1,5 @@
+using System.Text.Json;
+
 namespace Aspire.Hosting.ApplicationModel;
 
 /// <summary>
@@ -15,5 +17,17 @@ public class SqliteResource(string name, string databasePath, string databaseFil
     internal string DatabaseFilePath => Path.Combine(DatabasePath, DatabaseFileName);
 
     /// <inheritdoc/>
-    public ReferenceExpression ConnectionStringExpression => ReferenceExpression.Create($"Data Source={DatabaseFilePath};Cache=Shared;Mode=ReadWriteCreate;");
+    public ReferenceExpression ConnectionStringExpression => ReferenceExpression.Create($"Data Source={DatabaseFilePath};Cache=Shared;Mode=ReadWriteCreate;Extensions={JsonSerializer.Serialize(Extensions)}");
+
+    private readonly List<string> extensions = [];
+
+    /// <summary>
+    /// Gets the extensions to be loaded into the database.
+    /// </summary>
+    /// <remarks>
+    /// Extensions are not loaded by the hosting integration, the information is provided for the client to load the extensions.
+    /// </remarks>
+    public IReadOnlyCollection<string> Extensions => extensions;
+
+    internal void AddExtension(string extension) => extensions.Add(extension);
 }
--- a/src/CommunityToolkit.Aspire.Hosting.Sqlite/SqliteResource.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.Sqlite/SqliteResource.cs
@@ -1,3 +1,4 @@
+using Microsoft.Extensions.Hosting;
 using System.Text.Json;
 
 namespace Aspire.Hosting.ApplicationModel;
@@ -19,7 +20,7 @@ public class SqliteResource(string name, string databasePath, string databaseFil
     /// <inheritdoc/>
     public ReferenceExpression ConnectionStringExpression => ReferenceExpression.Create($"Data Source={DatabaseFilePath};Cache=Shared;Mode=ReadWriteCreate;Extensions={JsonSerializer.Serialize(Extensions)}");
 
-    private readonly List<ExtensionMetadata> extensions = [];
+    private readonly List<SqliteExtensionMetadata> extensions = [];
 
     /// <summary>
     /// Gets the extensions to be loaded into the database.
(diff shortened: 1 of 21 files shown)


src/CommunityToolkit.Aspire.Hosting.Sqlite/SqliteResource.cs
```diff
@@ -15,5 +17,26 @@ public class SqliteResource(string name, string databasePath, string databaseFil
...
 }
+
+/// <summary>
+/// Represents metadata for an extension to be loaded into a database.
+/// </summary>
+/// <param name="Extension">The name of the extension binary, eg: vec0</param>
+/// <param name="PackageName">The name of the NuGet package. Only required if <paramref name="IsNuGetPackage"/> is <see langword="true" />.</param>
+/// <param name="IsNuGetPackage">Indicates if the extension will be loaded from a NuGet package.</param>
+/// <param name="ExtensionFolder">The folder for the extension. Only required if <paramref name="IsNuGetPackage"/> is <see langword="false" />.</param>
+public record ExtensionMetadata(string Extension, string? PackageName, bool IsNuGetPackage, string? ExtensionFolder);
```
Alirexaa, line 42:
```suggestion
public record SqliteExtensionMetadata(string Extension, string? PackageName, bool IsNuGetPackage, string? ExtensionFolder);
```

I prefer `SqliteExtensionMetadata` to `ExtensionMetadata` because it prevents feature conflicts (e.g., PostgreSQL extensions like pgvector).

src/CommunityToolkit.Aspire.Microsoft.Data.Sqlite/SqliteConnectionSettings.cs
```diff
@@ -17,4 +17,18 @@ public sealed class SqliteConnectionSettings
...
 }
+
+/// <summary>
+/// Represents metadata for an extension to be loaded into a database.
+/// </summary>
+/// <param name="Extension">The name of the extension binary, eg: vec0</param>
+/// <param name="PackageName">The name of the NuGet package. Only required if <paramref name="IsNuGetPackage"/> is <see langword="true" />.</param>
+/// <param name="IsNuGetPackage">Indicates if the extension will be loaded from a NuGet package.</param>
+/// <param name="ExtensionFolder">The folder for the extension. Only required if <paramref name="IsNuGetPackage"/> is <see langword="false" />.</param>
+public record ExtensionMetadata(string Extension, string? PackageName, bool IsNuGetPackage, string? ExtensionFolder);
```
Alirexaa, line 34: This seems duplicate. Can we use a shared file (like dbgate)?

src/CommunityToolkit.Aspire.Hosting.Sqlite/SqliteResourceBuilderExtensions.cs
```diff
@@ -85,4 +86,51 @@ public static IResourceBuilder<SqliteResource> WithSqliteWeb(this IResourceBuild
...
+    /// </summary>
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="extension">The name of the extension file with to add, eg: vec0, without file extension.</param>
+    /// <param name="extensionPath">The path to the extension file.</param>
+    /// <returns>The resource builder.</returns>
+    /// <remarks>
+    /// Extensions are not loaded by the hosting integration, the information is provided for the client to load the extensions.
+    /// 
+    /// This extension is experimental while the final design of extension loading is decided.
+    /// </remarks>    find . -type f -name "*.orig" -delete
```
Alirexaa, line 124:
```suggestion
    /// </remarks>
```

Alirexaa (commented):
@aaronpowell, I understand that writing functional tests for this situation is challenging, but I believe we need to proceed with it. For now, we could create tests that run only on Windows, but we should definitely include some functional tests. Additionally, we might want to update the example project to utilize the one extension, which I could then attempt to test locally.

src/CommunityToolkit.Aspire.Hosting.Sqlite/SqliteResourceBuilderExtensions.cs
aaronpowell: Haha thanks Copilot 🤣

src/CommunityToolkit.Aspire.Microsoft.Data.Sqlite/SqliteConnectionSettings.cs
aaronpowell: Yeah could do

aaronpowell:
I'll have to think about a functional test because yes, it needs to run on Windows only (which means I can't actually run the tests myself).

aaronpowell:
@Alirexaa I've created a test that runs on Windows only to verify the extension loading, but as you can see, it doesn't pass on GitHub Actions.

Running locally, the test passes:

Passing test

Can you run the test on your machine and see if it passes? I'm not sure why it's failing, looking at the logging I've added it's finding the right location to load the extension from (it comes from `D:\a\Aspire\Aspire\tests\CommunityToolkit.Aspire.Hosting.Sqlite.Tests\bin\Release\net8.0\runtimes\win-x64\native`), so I wonder if there's something more with how the test runner is running and doing isolation, and we're going to either have to do a lot of workarounds or drop this test (could have it run locally but not in CI).

Alirexaa:
> Can you run the test on your machine and see if it passes?

I'll try.

aaronpowell:
Apparently you can load extensions from an absolute path, so I'll try that instead as it might simplify the code and maybe fix our issue.

aaronpowell:
Going to put this back to draft, as the package that I'm using seems to be really out of date and it's not meant to be used the way that I was thinking.

I actually am more interested in https://github.com/asg017/sqlite-vec but we depend on https://github.com/asg017/sqlite-vec/issues/193

aaronpowell:
> @aaronpowell, test passed locally. I think there is a limitation in the runner.

That's my guess. Are you ok if we merge without tests, as I'm going to be looking into the sqlite-vec plugin and shipping it via NuGet which I'd then migrate to instead.

Alirexaa (approved):
Let's keep this test and skip it for now. Overall, it looks good to me.

krwq:
@aaronpowell I believe EnsureLoadableFromNuGet won't be needed after https://github.com/dotnet/efcore/pull/35617 is merged

aaronpowell:
> @aaronpowell I believe EnsureLoadableFromNuGet won't be needed after [dotnet/efcore#35617](https://github.com/dotnet/efcore/pull/35617) is merged

Nice - I'll keep an eye on that being released.

PR #438, opened 2025-02-03
Improving CI

aaronpowell:
Converting our workflow to use a set of composite actions for the runtime setup and caching. See https://docs.github.com/en/actions/sharing-automations/avoiding-duplication for info on reducing duplication


aaronpowell:
@Alirexaa I want to see if we can merge this before anything else goes into `main`, so we can fix the current broken `main` (which I forgot to setup Dapr on)

Alirexaa (approved):

PR #443, opened 2025-02-03
Fix CI, Followup #438

Alirexaa:
This PR adds the `DAPR_VERSION` variable to the `dotnet-main` and the `dotnet-release` actions, which we should have added in #438.


aaronpowell:
Good catch

Issue #415, opened 2025-01-22
[Feature] Integration with DbGate and WhoDB applications


jakoss:
### Related to an existing integration?

No

### Overview

[DbGate](https://dbgate.org) and [WhoDB](https://whodb.clidey.com/docs/) are small and fast web-based database management tools (think of PgAdmin, PgWeb, or Mongo Express). The advantage over mentioned applications is that both DbGate and WhoDB can support multiple databases.

## My scenario

I'm currently working on a huge microservice system. We have some microservices based on Mongo, other on Postgres, but (since it's partly white-label tool) we want to support more SQL variants like SqlServer and Oracle soon. Also we use redis for caching and few message brokers. For all our local development needs we are using Aspire and it's being super helpful. 

After we hit a 20 microservice mark we spot one common issue people had - computers started running out of memory. The amount of containers was getting too big, CPUs were getting swetty, annoying stuff. Aspire is great, so we could run some containers optionally limiting the issue.

But there is one issue that is still there and we did not had a great solution for that - until now. The problem - each database needs to have it's own management ui. Need to spin postgres? Add PgWeb to that (PgAdmin is so slow we gave up on it very fast). Want to add Mongo? Sure, add express. Redis? Redis insights come in. So we have this issue where number of containers is growing with no gain. Each of those tools have the same reason to exists - manage data.

Hence the idea to use DbGate or WhoDB. Effectively we are able to replace all those additional containers with a single one. We can manage the data in a single place. We are using it a single day so far and already love it.

## Why both?

DbGate and WhoDB have the same purpose, but different set of features:

DbGate seems to be more mature and [have a lot more database supported](https://dbgate.org/features/database/)

WhoDB looks and feels much better, and [have some interesting AI features](https://whodb.clidey.com/docs/usage-houdini/what-is-houdini) (even supporting local LLMs)

Supporting both is not a big overhead, since the implementation will be very similar.

Moving forward I will describe API from a perspective of DbGate, but the WhoDB API will be pretty much identical to use.

## Nuget

Most of the implementation will live in the `CommunityToolkit.Aspire.Hosting.DbGate.Core`, but user won't be using that. Instead, there will be bunch more nugets for each database user wants to use, for example:
- `CommunityToolkit.Aspire.Hosting.DbGate.Postgres`
- `CommunityToolkit.Aspire.Hosting.DbGate.MySql`
- `CommunityToolkit.Aspire.Hosting.DbGate.MongoDB`
[...]

PR #453, opened 2025-02-06
Add DbGate support for MongoDB

Alirexaa:
**Contributes to #415  **

## PR Checklist

- [x] New integration
  - [ ] Docs are written


Alirexaa:
Running the Postgres-ext example causes multiple DbGate containers to spin up—one for Postgres and another for MongoDB.

This occurred because we shared and compiled `DbGateContainerResource.cs` in two projects. This means these types are not what we expected(as same or equal type), as they have different assemblies. Therefore, the following code does not function as intended.

```csharp
        //The `else` branch will be hit because DbGateContainerResource is not the same.
        if (builder.Resources.OfType<DbGateContainerResource>().SingleOrDefault() is { } existingDbGateResource)
        {
            var builderForExistingResource = builder.CreateResourceBuilder(existingDbGateResource);
            return builderForExistingResource;
        }
        else
        {
            var dbGateContainer = new DbGateContainerResource(name);
            var dbGateContainerBuilder = builder.AddResource(dbGateContainer)
                                               .WithImage(DbGateContainerImageTags.Image, DbGateContainerImageTags.Tag)
                                               .WithImageRegistry(DbGateContainerImageTags.Registry)
                                               .WithHttpEndpoint(targetPort: 3000, port: port, name: DbGateContainerResource.PrimaryEndpointName)
                                               .ExcludeFromManifest();

            return dbGateContainerBuilder;
        }
```

@aaronpowell, I suggest adding CommunityToolkit.Aspire.Hosting.DbGate and placing these shared files there.

aaronpowell:
Ah yes, that's a really good point.

You're right, we're going to need to create that package for the shared types.

src/CommunityToolkit.Aspire.Hosting.DbGate/README.md
aaronpowell, line 1: Should we just provide a mostly empty README since this is a package that isn't really designed for direct consumption?

aaronpowell (approved):
Just a question on the README

src/CommunityToolkit.Aspire.Hosting.DbGate/README.md
Alirexaa: We can not pack the packages with an empty README, but I updated the README to indicate that this package is designed for internal consumption.

PR #460, opened 2025-02-09
Dapr component schema

### LINKED ISSUE #459: Define Dapr Component Schema Model & WithMetadata Extension Method

Yes

### Existing integration

Hosting.Dapr

### Overview

Define strongly typed C# model that reflects the generic Dapr component schema. This model will serve as the baseline for all Dapr component configurations, capturing common properties such as name, type, and metadata. In addition, we want to enhance the resource builder API by adding a WithMetadata extension method for IResourceBuilder<IDaprComponentResource>. This extension should allow runtime overwriting of metadata properties whether we are loading an existing YAML component file or creating a new component dynamically.

Additionally currently StateStore & PubSub components look for a matching yaml file in dapr default location and then fall back to inmemory versions
We should enhance this to check for the existence of .dapr/components/ yaml file in the apphost directory

lastly we should create a secretstore.local.env when running app host to manage secrets

e.g apphost/.dapr/components/pubsub.yaml

### Usage example

given a pubsub.yaml in .dapr/components of 

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: pubsub
  namespace: default
spec:
  type: pubsub.rabbitmq
  version: v1
  metadata:
  - name: protocol
    value: amqp
  - name: hostname
    value: localhost
  - name: username
    value: guest
  - name: password
    value: guest
```
We should allow

```

var rmq = builder.AddRabbitMQ("rabbitMQ")
                   .WithManagementPlugin()
                   .WithEndpoint("tcp", e => e.Port = 5672)
                   .WithEndpoint("management", e => e.Port = 15672);


var stateStore = builder.AddDaprStateStore("statestore");

var pubSub = builder.AddDaprPubSub("pubsub")
                           .WithMetadata("password", rmq.Resource.PasswordParameter)
                    .WaitFor(rmq);

builder.AddProject<Projects.CommunityToolkit_Aspire_Hosting_Dapr_ServiceA>("servicea")
       .WithReference(stateStore)
       .WithReference(pubSub)
       .WithDaprSidecar()
       .WaitFor(rmq);

builder.AddProject<Projects.CommunityToolkit_Aspire_Hosting_Dapr_ServiceB>("serviceb")
       .WithReference(pubSub)
       .WithDaprSidecar()
       .WaitFor(rmq);

// console app with no appPort (sender only)
builder.AddProject<Projects.CommunityToolkit_Aspire_Hosting_Dapr_ServiceC>("servicec")
       .WithReference(stateStore)
       .WithDaprSidecar();

builder.Build().Run();
```
where the resulting yaml will be 

```
apiVersion: dapr.io/v1alpha1
      kind: Component
      auth:
        secretStore: secretstore
      metadata:
        name: pubsub
        namespace: default
      spec:
        type: pubsub.rabbitmq
        version: v1
        metadata:
        - name: protocol
          value: amqp
        - name: hostname
          value: localhost
        - name: username
          value: guest
        - name: password
          value: guest
        - name: password
          secretKeyRef:
            name: rabbitMQ-password
            key: rabbitMQ-password
```

with secretstore:

```yaml
apiVersion: dapr.io/v1alpha1
      kind: Component
      metadata:
        name: secretstore
      spec:
        type: secretstores.local.env
        version: v1
        metadata: []
```

### Breaking change?

No

### Alternatives

NA

### Additional context

_No response_

### Help us help you

FullStackChef:
**Closes #459 & #507 **

## PR Checklist

- [ ] New integration
  - [ ] Docs are written
- [] Every new API (including internal ones) has full XML docs


src/CommunityToolkit.Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -269,16 +291,24 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
...
-
                         context.Writer.WriteEndObject();
                     }));
 
             sideCars.Add(daprCli);
         }
 
+        foreach (var secret in secrets)
+        {
+            Environment.SetEnvironmentVariable(secret.Key, secret.Value);
```
davidfowl, line 302: In the current process? Why?

src/CommunityToolkit.Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -430,82 +468,87 @@ private async Task<IReadOnlyDictionary<string, string>> StartOnDemandDaprCompone
...
-            return await contentWriter(GetInMemoryStateStoreContent(component)).ConfigureAwait(false);
-        }
+        // If the component is not found in the default components directory, use the in-memory secret store
+        _logger.LogInformation("Using in-memory provider for dapr component '{ComponentName}'.", component.Name);
+
+        var content = new DaprComponentSchema(component.Name, inMemoryProvider).ToString();
+        return await contentWriter(content).ConfigureAwait(false);
+    }
+
+    private string GetAppHostRelativePath(string componentName)
```
davidfowl, line 510: Aside, we're creating a place to store things like this in 9.1, the aspire store.

examples/dapr/CommunityToolkit.Aspire.Hosting.Dapr.AppHost/Program.cs
```diff
@@ -5,18 +5,23 @@
                    .WithEndpoint("tcp", e => e.Port = 5672)
                    .WithEndpoint("management", e => e.Port = 15672);
 
+
 var stateStore = builder.AddDaprStateStore("statestore");
+
 var pubSub = builder.AddDaprPubSub("pubsub")
+                           .WithMetadata("password", rmq.Resource.PasswordParameter)
```
davidfowl, line 12:
```suggestion
                 .WithMetadata("password", rmq.Resource.PasswordParameter)
```

src/CommunityToolkit.Aspire.Hosting.Dapr/DaprComponentSecretAnnotation.cs
```diff
@@ -0,0 +1,4 @@
+﻿using Aspire.Hosting.ApplicationModel;
+
+namespace CommunityToolkit.Aspire.Hosting.Dapr;
+internal record DaprComponentSecretAnnotation(string Key, string Value) : IResourceAnnotation;
```
davidfowl, line 4: nit:
```suggestion

internal record DaprComponentSecretAnnotation(string Key, string Value) : IResourceAnnotation;
```

src/CommunityToolkit.Aspire.Hosting.Dapr/DaprComponentSchema.cs
```diff
@@ -0,0 +1,138 @@
...
+
+/// <summary>
+/// Represents a Dapr secret key reference
+/// </summary>
+public sealed class DaprSecretKeyRef
+{
+    /// <summary>
+    /// The name of the secret
+    /// </summary>
+    public required string Name { get; init; } = default!;
```
davidfowl, line 132: I don't think you need default! if it's required.

src/CommunityToolkit.Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
FullStackChef: haha "because it works"

Setting these environment variables loads them against the command line is executed - if there's a better place to do so I'd be happy to change as this felt a little like a random place to do it. 

The dapr local.env secret store  pulls variables from the env.

src/CommunityToolkit.Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
FullStackChef: nice -Apphost relative? wanna point me to the code and I could update inline with that?

src/CommunityToolkit.Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
davidfowl: What about collisions? You're setting this on the app host process. Explain to me how this is supposed to work? The app calls the dapr side car for secrets, that sidecar is configured with a secret store component that gets secrets from the side cars env?

src/CommunityToolkit.Aspire.Hosting.Dapr/DaprComponentSchema.cs
FullStackChef: Ah you're right - I needed it for a couple of the other properties in order to allow the parameter less constructor - I've removed these ones

src/CommunityToolkit.Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
FullStackChef: Collisions with other aspire vars - should totally prefix this. collisions with multiple sidecars... too bad name your parameters better 😬 

Yes it is setting it on the app host process - the app host is what spawns the dapr command line right? 

dapr local env secret store works by exposing environment variables as "secrets" - it's designed for local development only

1. We have component (pubsub.rabbitmq) that has a parameter resource for password 
2. when we start daprcomponentsecretannotation exist so we need to add secret store
3. we add secret store component and add it to the sidecar paths
4.  setting the environment vars here is similar to exporting a var on the command line prior to calling dapr run ...
5. pubsub component has access to secret store therefore connection to rabbitmq is achieved

(the daprclient in dotnet also has access to these variables - but the primary focus here is the component access
e.g

```bash
export rabbitmq-password=SuperSecretPassword123
dapr run --app-id myapp dotnet run
```

sidecar has secret store and pubsub components - pubsub component retrieves password for rabbitmq from secret store

FullStackChef: > > The dapr local.env secret store pulls variables from the env.
> 
> Missed this. I see, then I would just env variables on the sidecar itself.

I tried this (or at least I thought I did) I will take another crack at it see if I can't get it done that way

davidfowl: This should work.

FullStackChef: Have updated - I think I had tried setting the vars on the resource rather than the sidecar accidentally which is why it wasn't working - setting on sidecar is working. ick code gone

src/CommunityToolkit.Aspire.Hosting.Dapr/DaprComponentSchema.cs
```diff
@@ -0,0 +1,138 @@
...
+        };
+    }
+    public override string ToString()
+    {
+        return serializer.Serialize(this);
+    }
+    public static implicit operator DaprComponentSchema(string yamlContent)
+    {
+        return deserializer.Deserialize<DaprComponentSchema>(yamlContent);
+    }
```
davidfowl, line 38: This feels bit too cute 😄

FullStackChef: I **think** I know what you mean 😅
```csharp
public static DaprComponentSchema FromYaml(string yamlContent) =>
    deserializer.Deserialize<DaprComponentSchema>(yamlContent);
```
Does this make you happier? and can you tell me why?  Is it just that you prefer code to be more explicit?

davidfowl: You're specifically trying to deserialize YAML. Besides the parameter name of this method, the implicit cast makes that invisible and this just leads to makes things less type safe.

paule96:
Maybe if it is not to big of a no scope of this PR we should already start deviding dapr sidecar and component lifecycles to make our lifes easier later.

So that we would have two LifecycleHooks and that everytime you have something like `WithReference` on a sidecare you wait for this resource to complete.

With this we will have an easier life if we need to consider the endpoints to generate the correct YAML files

FullStackChef:
@paule96 my preference would be not to increase the scope of the my PR and I think the concept of separating lifecycle hooks needs further thought /design

src/CommunityToolkit.Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -392,6 +428,12 @@ private async Task<IReadOnlyDictionary<string, string>> StartOnDemandDaprCompone
                 .Where(component => component.Options?.LocalPath is null)
                 .ToList();
 
+        // If any of the components have secrets, we will add an on-demand secret store component.
+        if (onDemandComponents.Any(component => component.TryGetAnnotationsOfType<DaprComponentSecretAnnotation>(out var annotations) && annotations.Any()))
+        {
+            onDemandComponents.Add(new DaprComponentResource("secretstore", DaprConstants.BuildingBlocks.SecretStore));
```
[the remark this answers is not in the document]
FullStackChef: I don't think it's that simple. As a developer when I'm working locally, I want to be able to reference aspire parameters in my dapr metadata. If I have set those parameters as secrets then a dapr integration should respect that and treat them as such without me having to specifically set up a component.

Especially because when my resource requires secrets aspire automatically adds these secrets to a key vault and then references the key vault

However - looking at my implementation - I have overlooked what should happen if the user has configured a secrets component which should take priority

src/CommunityToolkit.Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
FullStackChef: I took a look at the aspire store interface - but it looks like a place to write data to rather than a location for user defined data - the idea of the app relative path is to provide a consistent location for users to place dapr components relative to the project rather than in the userprofile directory

examples/dapr/CommunityToolkit.Aspire.Hosting.Dapr.AppHost/.dapr/components/pubsub.yaml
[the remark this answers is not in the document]
FullStackChef: this file is used by this line of the apphost
```csharp
var pubSub = builder.AddDaprPubSub("pubsub")
```
this is not new  - it's simply in a project relative location

examples/dapr/CommunityToolkit.Aspire.Hosting.Dapr.AppHost/Program.cs
```diff
@@ -1,22 +1,28 @@
...
 builder.AddProject<Projects.CommunityToolkit_Aspire_Hosting_Dapr_ServiceA>("servicea")
-       .WithDaprSidecar()
        .WithReference(stateStore)
-       .WithReference(pubSub);
+       .WithReference(pubSub)
+       .WithDaprSidecar();
 
 builder.AddProject<Projects.CommunityToolkit_Aspire_Hosting_Dapr_ServiceB>("serviceb")
+       .WithReference(pubSub)
        .WithDaprSidecar()
```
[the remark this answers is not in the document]
FullStackChef: this is up to the developer the both work the same

src/CommunityToolkit.Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -418,9 +461,10 @@ private async Task<IReadOnlyDictionary<string, string>> StartOnDemandDaprCompone
...
                 string componentPath = await (component.Type switch
                 {
-                    DaprConstants.BuildingBlocks.PubSub => GetPubSubAsync(component, contentWriter, cancellationToken),
-                    DaprConstants.BuildingBlocks.StateStore => GetStateStoreAsync(component, contentWriter, cancellationToken),
-                    _ => throw new InvalidOperationException($"Unsupported Dapr component type '{component.Type}'.")
+                    DaprConstants.BuildingBlocks.PubSub => GetBuildingBlockComponentAsync(component, contentWriter, "pubsub.in-memory", cancellationToken), // NOTE: In memory component can only be used within a single Dapr application.
+                    DaprConstants.BuildingBlocks.StateStore => GetBuildingBlockComponentAsync(component, contentWriter, "state.in-memory", cancellationToken),
+                    DaprConstants.BuildingBlocks.SecretStore => GetBuildingBlockComponentAsync(component, contentWriter, "secretstores.local.env", cancellationToken),
+                    _ => GetComponentAsync(component, contentWriter, cancellationToken)
                 }).ConfigureAwait(false);
```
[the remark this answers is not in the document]
FullStackChef: because the first three cases in the switch statement handle specific building blocks with local development defaults

src/CommunityToolkit.Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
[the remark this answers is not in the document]
FullStackChef: I've looked at this again and it turns out I did consider a user defined secrets component - like the existing pubsub & state, the secretstore will infact only use the local env in the absence of a secretstore.yaml in the apphost relative component directory (apphost/.dapr/components) and the dapr default component directory (userprofile/.dapr/components)

FullStackChef:
Quickly added parent relationship to address https://github.com/CommunityToolkit/Aspire/issues/507
image

aaronpowell:
@FullStackChef is this ready to be considered reviewable/mergable?

FullStackChef:
From my point of view, yes

aaronpowell (commented):
From a purely code perspective, it looks all good, but as I'm not across this problem space I'll delegate the "does it work" stuff to others.

But if there are no complaints over the next few days I'd take that as "ready to merge"

FullStackChef:
@aaronpowell  - yep keen to get this merged so I can start looking at #516 as well as start building out some more functionality

aaronpowell:
Can you remove the PublicAPI file and it'll be good to merge.

I think I'll add this with the Redis one, and call that 9.3 so we can ship the new integration - that sound good?

FullStackChef:
@aaronpowell sounds good - I've removed the public api files and fixed the conflict with the packages file. Once this is merged I'll try to get a PR for #516 later today so we can squeeze it into 9.3 if possible - it should be a pretty small fix (check for customization annotation and configure based appropriately based on ) but it represents a fairly blocking issue

aaronpowell:
ok, let's merge this in, that way you're unblocked but I won't release a 9.2.2, we'll have that in pre-release and go straight to 9.3.

PR #465, opened 2025-02-11
Aligning OllamaSharp with the Aspire OpenAI + MEAI design

### LINKED ISSUE #461: Improving integration with the `ChatClientBuilder` and middleware of MEAI
Firstly, this integration is a great way to get started with local models and configure them in your application! There are a few rough edges when it comes to API composability when using Microsoft.Extensions.AI. This simple API:

```C#
builder.AddOllamaSharpChatClient("ollama-phi3-5");
```

Turns into:

```C#
builder.AddOllamaApiClient("ollama-phi3-5");

builder.Services.AddChatClient(sp =>
{
    var client = sp.GetRequiredService<IOllamaApiClient>();
    return (IChatClient)client;
})
.UseOpenTelemetry()
.UseLogging();

builder.Services.AddOpenTelemetry().WithTracing(t => t.AddSource("Experimental.Microsoft.Extensions.AI"));
```

When I want to wire up middleware on the `IChatClient` instance exposed by MEAI, I need to do these gymnastics.

Instead, could we add:

```C#
builder.AddOllamaSharpChatClient("ollama-phi3-5");

// PS: These should be on by default
builder.AddOllamaSharpChatClient("ollama-phi3-5", configure: builder => builder.UseOpenTelemetry().UseLogging()); 
```

aaronpowell:
Moving away from our current ways of registering an IChatClient/IEmbeddingGenerator to leverage the builder pattern from MEAI. This allows the consumer to add middleware to the AI pipeline.

Existing methods have been marked as Obsolete.

Set some default middleware with logging and OTEL.

Fixes #461

### DIFF
--- a/src/CommunityToolkit.Aspire.OllamaSharp/AspireOllamaApiClientBuilder.cs
+++ b/src/CommunityToolkit.Aspire.OllamaSharp/AspireOllamaApiClientBuilder.cs
@@ -1,0 +1,27 @@
+using OllamaSharp;
+
+namespace Microsoft.Extensions.Hosting;
+
+/// <summary>
+/// Builder class for configuring and creating an instance of AspireOllamaApiClient.
+/// </summary>
+/// <param name="hostBuilder">The <see cref="IHostApplicationBuilder"/> with which services are being registered.</param>
+/// <param name="serviceKey">The service key used to register the <see cref="OllamaApiClient"/> service, if any.</param>
+/// <param name="disableTracing">A flag to indicate whether tracing should be disabled.</param>
+public class AspireOllamaApiClientBuilder(IHostApplicationBuilder hostBuilder, string serviceKey, bool disableTracing)
+{
+    /// <summary>
+    /// The host application builder used to configure the application.
+    /// </summary>
+    public IHostApplicationBuilder HostBuilder { get; } = hostBuilder;
+
+    /// <summary>
+    /// Gets the service key used to register the <see cref="OllamaApiClient"/> service, if any.
+    /// </summary>
+    public string ServiceKey { get; } = serviceKey;
+
+    /// <summary>
+    /// Gets a flag indicating whether tracing should be disabled.
+    /// </summary>
+    public bool DisableTracing { get; } = disableTracing;
+}
--- a/src/CommunityToolkit.Aspire.OllamaSharp/AspireOllamaSharpExtensions.cs
+++ b/src/CommunityToolkit.Aspire.OllamaSharp/AspireOllamaSharpExtensions.cs
@@ -1,9 +1,8 @@
-using Aspire;
 using CommunityToolkit.Aspire.OllamaSharp;
 using Microsoft.Extensions.AI;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using OllamaSharp;
 using System.Data.Common;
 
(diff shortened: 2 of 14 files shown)


aaronpowell:
@davidfowl

src/CommunityToolkit.Aspire.OllamaSharp/AspireOllamaApiClientBuilder.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Microsoft.Extensions.Hosting;
+
+/// <summary>
+/// Builder class for configuring and creating an instance of AspireOllamaApiClient.
+/// </summary>
+/// <param name="hostBuilder">The <see cref="IHostApplicationBuilder"/> with which services are being registered.</param>
+/// <param name="serviceKey">The service key used to register the <see cref="OllamaApiClient"/> service, if any.</param>
+/// <param name="disableTracing">A flag to indicate whether tracing should be disabled.</param>
+public class AspireOllamaApiClientBuilder(IHostApplicationBuilder hostBuilder, string serviceKey, bool disableTracing)
```
davidfowl, line 11: Maybe remove this constructor and just use required properties? Feels like we're going to add more properties over time.

src/CommunityToolkit.Aspire.OllamaSharp/AspireOllamaSharpExtensions.cs
```diff
@@ -23,11 +22,11 @@ public static class AspireOllamaSharpExtensions
     /// <param name="connectionName">A name used to retrieve the connection string from the ConnectionStrings configuration section.</param>
     /// <param name="configureSettings">An optional delegate that can be used for customizing options. It's invoked after the settings are read from the configuration.</param>
     /// <exception cref="UriFormatException">Thrown when no Ollama endpoint is provided.</exception>
-    public static void AddOllamaApiClient(this IHostApplicationBuilder builder, string connectionName, Action<OllamaSharpSettings>? configureSettings = null)
+    public static AspireOllamaApiClientBuilder AddOllamaApiClient(this IHostApplicationBuilder builder, string connectionName, Action<OllamaSharpSettings>? configureSettings = null)
```
davidfowl, line 25: This is a breaking change. Are you OK with that?

davidfowl:
APIs look much nicer! It's a breaking change though. What's the policy on that?

src/CommunityToolkit.Aspire.OllamaSharp/AspireOllamaApiClientBuilder.cs
aaronpowell: It was to be consistent with https://github.com/dotnet/aspire/blob/main/src/Components/Aspire.OpenAI/AspireOpenAIClientBuilder.cs#L26-L32

src/CommunityToolkit.Aspire.OllamaSharp/AspireOllamaSharpExtensions.cs
aaronpowell: Yes, it is a breaking change, but in terms of breaking changes it's a pretty unobtrusive one. Having it as a `void` method previously means that you weren't capturing anything to now having a returned value, but since C# doesn't require return assignment it shouldn't result in a compilation error.

It would be a problem if anyone is using reflection to find the method and invoking it based off an expected signature, and _maybe_ a problem in F# as the return needs to be handled (or piped to `ignore`), but both are pretty obscure edge cases.

aaronpowell:
> APIs look much nicer! It's a breaking change though. What's the policy on that?

"Aaron does what he wants" 😂.

As I mentioned on the specific comment, I consider this a minor breaking change and acceptable at that. We could introduce a new method and deprecate all the current ones, but the naming of that would deviate from the generally used Aspire method naming structure of `Add<Type To Add>`.

Behaviourally, the code is the same though, with the exception that we now add the middleware for logging and OTEL to `IChatClient` and `IEmbeddingGenerator`, but those middleware stages are unobtrusive and (arguably) not impactful to the consumer.

PR #466, opened 2025-02-11
Updating the deno version to latest

aaronpowell:
Seeing if this unblocks #377


aaronpowell:
Looks like we still have the same problem, so cancelling this PR

PR #468, opened 2025-02-12
New tests workflow

aaronpowell:
This is an overhaul of the runners and makes it more like the Aspire repo approach - https://github.com/dotnet/aspire/blob/5e270847fda789982de623bd5486f777989bc7cd/.github/workflows/tests.yml

The basic idea is that we use a matrix to run each of the test projects (across each OS), so that rather than having a single long test run, we have multiple shorter runs (it's taking <10 minutes consistently vs the ~20minutes currently).

This improves the visibility of the test runs, especially if there are fails, since we don't need to scroll through the whole test list, we can see exactly which workflow the failure was on.

- [x] Refactor CI test job to call new workflow.
- [x] Verify tests pass.
- [x] Verify coverage is reported.
- [x] Update `main` and `release` workflows.
- [x] Remove obsolete workflows.
- [x] Create way to generate test list.
- [x] Include workflow to verify all test projects are included in CI job

Here's what we see now in the summary of the job:
image

There is a summary of each individual test run (for each OS), which means that we get more fine grain details on the test runs.

It also gives us better insights into test failures. Here's an artificial failure I have added:

image


aaronpowell:
@ErikEJ @jmezach I'm trying to do some tweaks to the CI pipeline and it looks like there's an assumption on the `GetDacpackPath` that results in it wanting to get the dacpack from the `Debug` folder, despite the CI job running in `Release`.

You can see that in this test failure: https://github.com/CommunityToolkit/Aspire/actions/runs/13384342116/job/37378295078?pr=468#step:7:46

I'm unfamiliar with how it does the resolution for the dacpack, but I'd have thought it would use the configuration that is being used to run the tests/app.

Can you have a look at what might be wrong here?

ErikEJ:
@aaronpowell Maybe the project here https://github.com/CommunityToolkit/Aspire/blob/main/examples/sql-database-projects/SdkProject/SdkProject.csproj must be built in order to produce a .dacpac? (still odd why it is looking in the debug folder)

jmezach:
We are essentially loading the referenced project file using the MSBuild API's. We aren't explicitly passing a configuration or anything, so I guess it defaults to Debug. Not sure if there's much we can do about that. Perhaps we should pass the same configuration at the one used to build the AppHost? Not sure if we can get to that though.

aaronpowell:
Looking at the build log we can see the dacpack is being generated, so there's obviously something that the MSBuild API isn't doing that we need.

I'll debug and see if I can find how it generates that path

aaronpowell:
Figured it out - https://github.com/CommunityToolkit/Aspire/pull/468/commits/88895231af36613892aec5043bf5495423720455

jmezach:
@aaronpowell Nice!

aaronpowell:
@CommunityToolkit/aspire-contributors anyone want to have a review of this?

Here's an example of a failing test run if you want to see what that looks like: https://github.com/CommunityToolkit/Aspire/actions/runs/13403676338

esskar (approved):

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,133 @@
...
+          /p:TrxLogFileNameSuffix=${{ matrix.os }} `
+          --blame `
+          --blame-hang-timeout 7m `
+          --blame-crash `
+          --results-directory testresults `
+          --collect "XPlat Code Coverage" `
+          --no-restore `
+          --no-build -- RunConfiguration.CollectSourceInformation=true
+        shell: pwsh
+        if: runner.os == 'Windows'
```
mderriey, line 111: Heelo, I came to have a look after your BlueSky post.

Is it expected that only the Windows step has `/p:TrxLogFileNameSuffix=${{ matrix.os }}`?

If it isn't, and you'd like to avoid duplicating the arguments to avoid future mistakes, you could use `>-` instead of `|` so that new lines are replaced by spaces, also avoiding the `` ` `` and `\` difference between PowerShell and Bash, and leaving GitHub Actions to decide which shell to use based on the OS.

```yaml
name: Run tests
run: >-
  dotnet test ${{ github.workspace }}/${{ env.TEST_PROJECT_PATH }}
  --configuration ${{ env.DOTNET_CONFIGURATION }}
```

YAML will expand that to `dotnet test ${{ github.workspace }}/${{ env.TEST_PROJECT_PATH }} --configuration ${{ env.DOTNET_CONFIGURATION }}`.

See <https://yaml-multiline.info/> to visualise the differences between the several multiline options.

aaronpowell:
Ah good catch. It's not meant to be used anymore on any OS as it causes the GitHub summary reporter that I added to not report.

I'll consolidate it

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

.github/workflows/tests.yaml
aaronpowell, line 111:
```suggestion
      - name: Run tests
        run: >-
          dotnet test ${{ github.workspace }}/${{ env.TEST_PROJECT_PATH }}
          --configuration ${{ env.DOTNET_CONFIGURATION }}
          --logger "console;verbosity=normal"
          --logger "trx"
          --logger "GitHubActions;summary.includePassedTests=true;summary.includeSkippedTests=true"
          --blame
          --blame-hang-timeout 7m
          --blame-crash
          --results-directory testresults
          --collect "XPlat Code Coverage"
          --no-restore
          --no-build -- RunConfiguration.CollectSourceInformation=true
```

aaronpowell: Good catch. The suffix isn't meant to be on either as using that was causing a problem with the GitHubAction logger that is introduced.

And nice tip on the simplifying the YAML, I *thought* there was something I could do but couldn't remember. I've updated and let's see how it goes.

Issue #477, opened 2025-02-17
Feat: Add Integration for LavinMQ


prom3theu5:
### Related to an existing integration?

No

### Overview

Add support for a LavinMQ container resource.
LavinMQ is a wire compatible AMQP server thats a drop in replacement for RabbitMQ.
It's largely more performant, requiring less system resources (about 30mb ram for example at idle) to run.

https://lavinmq.com/

Unlike RabbitMQ which uses a plugin based approach, the management console is built into the lavinmq binary and always enabled.

This integration would be compatible with the existing MassTransit.RabbitMQ package

### Usage example

```csharp
var builder = DistributedApplication.CreateBuilder(args);

var lavinmq = builder.AddLavinMQ(
        name: "lavinmq",
        port: 5672,
        managementPort: 15672)
    .WithExternalHttpEndpoints();

var api = builder.AddProject<Projects.CommunityToolkit_Aspire_MassTransit_RabbitMQ_ApiService>("api")
    .WithReference(lavinmq).WaitFor(lavinmq);

builder.AddProject<Projects.CommunityToolkit_Aspire_MassTransit_RabbitMQ_Publisher>("publisher")
    .WaitFor(api).WaitFor(lavinmq).WithReference(lavinmq);

builder.Build().Run();
```

### Breaking change?

No

### Alternatives

```csharp
var builder = DistributedApplication.CreateBuilder(args);

var rmq = builder.AddRabbitMQ(
        name: "rmq",
        port: 5672)
    .WithExternalHttpEndpoints()
    .WithManagementPlugin(port: 15672);

var api = builder.AddProject<Projects.CommunityToolkit_Aspire_MassTransit_RabbitMQ_ApiService>("api")
    .WaitFor(rmq).WithReference(rmq);

builder.AddProject<Projects.CommunityToolkit_Aspire_MassTransit_RabbitMQ_Publisher>("publisher")
    .WaitFor(api).WaitFor(rmq).WithReference(rmq);

builder.Build().Run();
```

### Help us help you

Yes, I'd like to be assigned to work on this item

prom3theu5:
Covered in #486

PR #486, opened 2025-02-18
feat: Add LavinMQ support #477

prom3theu5:
This pull request adds support for a LavinMQ integration, including unit tests.

Documentation has been created under: https://github.com/dotnet/docs-aspire/pull/2634

This fulfils the proposal outlined in https://github.com/CommunityToolkit/Aspire/issues/477

- [x] New integration
  - [x] Docs are written

Please note - Unlike RabbitMQ, LavinMQ does not support configuration utilising environmental variables.
As such, the default credentials of `guest:guest` are used in this implementation, with a virtual host of `/`
User management and virtual host management can be performed using the LavinMQ management UI, which is built in.
Utilising AddConnectionString, a custom post configured instance of lavinmq could be configured and passed down to consuming (reference) apps, but thats outside the scope of the actual container resource.

LavinMQ is wire compatible with AMQP 0-9-1, meaning we can reuse the RabbitMQ.Client to connect to this.
Its a drop in for it in reality - meaning the existing MassTransit extensions will work with this too.
As such - the health check added utilises the `AspNetCore.HealthChecks.RabbitMQ` package, as their isn't a native one for LavinMQ. It's fully functional though as I use it in a production AKS cluster with MassTransit.


src/CommunityToolkit.Aspire.Hosting.LavinMQ/LavinMQContainerResource.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    private static ReferenceExpression UserNameReference => ReferenceExpression.Create($"{DefaultUserName}");
+
+    private static ReferenceExpression PasswordReference => ReferenceExpression.Create($"{DefaultPassword}");
+
+    /// <summary>
+    /// ConnectionString for the LavinMQ server in the form of amqp://guest:guest@host:port/.
+    /// </summary>
+    public ReferenceExpression ConnectionStringExpression =>
+        LavinMQConnectionString.Create(UserNameReference, PasswordReference, PrimaryEndpoint);
```
aaronpowell, line 33: Is there a reason you introduced a class rather than using `ReferenceExpression.Create` like other integrations?

src/CommunityToolkit.Aspire.Hosting.LavinMQ/LavinMQHostingExtension.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        LavinMQContainerResource instance = new LavinMQContainerResource(name);
+
+        string? connectionString = null;
+
+        builder.Eventing.Subscribe<ConnectionStringAvailableEvent>(instance, async (_, ct) =>
+        {
+            connectionString = await instance.ConnectionStringExpression.GetValueAsync(ct).ConfigureAwait(false);
+
+            if (connectionString == null)
```
aaronpowell, line 41:
```suggestion
            if (connectionString is null)
```

aaronpowell (changes requested):

src/CommunityToolkit.Aspire.Hosting.LavinMQ/LavinMQContainerResource.cs
prom3theu5: Ahh. This was purely because Lavin could introduce env var config. If they did, we’d have to switch on VirtualHost being null or not to create the connection string with optional expressions. 

Right now they dont support that though, so I could change it?

aaronpowell (approved):

PR #534, opened 2025-03-06
Using the built-in method to register a HttpClient in .NET for OllamaSharp

### LINKED ISSUE #474: Implement resilience for OllamaSharp integration
> I don't have a strong opinion on if ollamasharp should built in resiliency, I do care that the aspire integration does. AFAIK the constructor already supports an HttpClient so this is what I did:
> 
> https://github.com/davidfowl/aspire-ai-chat-demo/blob/1779a65d4d93b52be102758f188b2581fd0348fe/ChatApi/ChatClientExtensions.cs#L38-L49
> 
> Which turns out to be pretty required when you run on underpowered devices. 

 _Originally posted by @davidfowl in [#461](https://github.com/CommunityToolkit/Aspire/issues/461#issuecomment-2661636395)_

aaronpowell:
This means that we can leverage the configurations applied to HttpClient via ServiceDefault around stuff like resiliancy, rather than creating a HttpClient in isolation that is not aware of any of that configuration.

No new tests are added as there's not we could easily test with this, but all existing tests should continue to pass

Fixes #474

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -35,6 +35,7 @@
     <PackageVersion Include="Microsoft.Extensions.AI" Version="$(MEAIVersion)" />
     <PackageVersion Include="Microsoft.Extensions.Configuration.Binder" Version="$(DotNetExtensionsVersion)" />
     <PackageVersion Include="Microsoft.Extensions.DependencyModel" Version="$(DotNetExtensionsVersion)" />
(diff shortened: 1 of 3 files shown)


aaronpowell:
You'll be able to simplify your demos shortly @davidfowl 😉

davidfowl:
🥂

PR #546, opened 2025-03-07
Use “uvicorn” from .venv if available

### LINKED ISSUE #545: AddUvicornApp doesn't use uvicorn from the virtual environment
### Describe the bug

Trying to launch a python service from Aspire using `AddUvicornApp()`, the launch fails with:

> ModuleNotFoundError: No module named 'redis'

When I launch the python service manually, using `poetry run uvicorn app.main:app` everything works fine instead.

I found that the different behavior is because `AddUvicornApp` uses `uvicorn` (and python) from the host, rather than the one inside `.venv`:

Manual => `poetry run uvicorn app.main:app` => Ok
AddUvicornApp  => `uvicorn app.main:app` => Fail

Could you change the implementation to allow using uvicorn from the virtual env (.venv dir)?

### Regression

_No response_

### Steps to reproduce

```text
Create app with Poetry, add a dependency missing in the host, integrate the service using `AddUvicornApp`. The app crashes.
```

### Expected behavior

`AddUvicornApp` should leverage uvicorn from the app .venv folder

### Screenshots

_No response_

### IDE and version

Other

### IDE version

CLI

### Nuget packages

```text
Aspire.Hosting.AppHost 9.1.0
CommunityToolkit.Aspire.Hosting.Python.Extensions 9.2.1
```

### Additional context

_No response_

### Help us help you

dluc:
When adding a Python resource using `AddUvicornApp` and using `uvicorn`, use `uvicorn` from the virtual environment if available, before choosing `uvicorn` from the current active python env.

This is particularly important when running multiple python resources, each with a potentially different version of uvicorn, each stored in separate environments under the resource.

Closes #545

## PR Checklist

- [ ] New integration
  - [ ] Docs are written

## Other information

This might be considered a breaking change for local development when running python resources off the host python env, without activating the resource venv. Worth evaluating, but the previous implementation seems to depend on some lucky factors, like python resources aligned on the same version and the venv being activated before launching Aspire AppHost.

### DIFF
--- a/src/CommunityToolkit.Aspire.Hosting.Python.Extensions/UvicornAppHostingExtension.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.Python.Extensions/UvicornAppHostingExtension.cs
@@ -50,7 +50,8 @@ public static class UvicornAppHostingExtension
             : Path.Join(projectDirectory, virtualEnvironmentPath));
 
         var instrumentationExecutable = virtualEnvironment.GetExecutable("opentelemetry-instrument");
-        var projectExecutable = instrumentationExecutable ?? "uvicorn";
+        var uvicornExecutable = virtualEnvironment.GetExecutable("uvicorn") ?? "uvicorn";
+        var projectExecutable = instrumentationExecutable ?? uvicornExecutable;
 
         var projectResource = new UvicornAppResource(name, projectExecutable, projectDirectory);
 


dluc:
@dotnet-policy-service agree [company="Microsoft"]

dluc:
@dotnet-policy-service agree company="Microsoft"

dluc:
note: the same fix might ne necessary for `uv` - I'll leave that out since I don't use uv and I wouldn't be able to test it

dluc:
@dotnet-policy-service agree company="Microsoft"

aaronpowell (approved):

PR #585, opened 2025-03-20
Use WithParentRelationship to nest resources

### LINKED ISSUE #502: Add resource relationships for SQL database projects

Yes

### Existing integration

SqlDatabaseProjects

### Overview

With .NET Aspire 9.1 releasing I was wondering if we should add the new relationships to the SQL database projects resources so that could become nested under the server resource in the dashboard. Haven’t looked yet at what would be required but should be fairly simple.

### Usage example

I don’t think this would require new public API’s. Technically the relationship is already there, we just don’t model it currently 

### Breaking change?

No

### Alternatives

I guess we could just not do it, but I think it makes sense and removes a fair amount of clutter for folks with many database projects being deployed to the same server.

### Additional context

@ErikEJ Any thoughts?

### Help us help you

jmezach:
**Closes #502**

This ensures that `SqlProject` and `SqlPackage` resources are nested underneath the `SqlServerDatabase` resource in the Aspire dashboard, like this:

Screenshot 2025-03-20 at 10 03 42

## PR Checklist

- [-] New integration
  - [-] Docs are written
  - [-] Added description of major feature to project description for NuGet package (4000 total character limit, so don't push entire description over that)

### DIFF
--- a/src/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects/SqlProjectBuilderExtensions.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects/SqlProjectBuilderExtensions.cs
@@ -192,7 +192,7 @@ public static class SqlProjectBuilderExtensions
         builder.ApplicationBuilder.Services.TryAddSingleton<IDacpacDeployer, DacpacDeployer>();
         builder.ApplicationBuilder.Services.TryAddSingleton<SqlProjectPublishService>();
 
-        builder.WithRelationship(target.Resource, "Reference");
+        builder.WithParentRelationship(target.Resource);
 
         if (target.Resource is SqlServerDatabaseResource)
         {


jmezach:
@ErikEJ Hmm, yeah, good point. I'll look into that.

jmezach:
@ErikEJ Not sure what changed, but it looks like the resources are no longer put into the `Waiting` state. Even our initial state that we're setting as Pending is not shown.

ErikEJ:
@jmezach Maybe @aaronpowell or @davidfowl has some ideas??

aaronpowell:
is it only doing that with the change to a parent/child relationship?

jmezach:
@aaronpowell Interestingly, it does something else if I don't add the `WithParentRelationship` call.

Screenshot 2025-03-21 at 08 15 37

From the screenshot you can see that they are now in a Pending state, which is what we are setting explicitly [here](https://github.com/CommunityToolkit/Aspire/blob/6e92a52e583bf25efd97aa3f9d0b56779e27623e/src/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects/SqlProjectBuilderExtensions.cs#L57-L62) and [here](https://github.com/CommunityToolkit/Aspire/blob/6e92a52e583bf25efd97aa3f9d0b56779e27623e/src/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects/SqlProjectBuilderExtensions.cs#L83-L88).

So it looks like adding the `WithParentRelationship` call somehow makes our initial state not come through. Regardless of that, the waiting state doesn't seem to work anymore either?

aaronpowell:
Weird, does the info of the resource tell you why the state is that?

I've seen weird things if I have the wrong Aspire SDK and NuGet package versions

jmezach:
@aaronpowell Not sure what I'm looking for. What I did notice as well as that in the past I believe I saw in the console logs for the resource that it is waiting on some other resource. I'm not seeing that anymore either?

aaronpowell:
Just tried on Windows and Linux and I'm not seeing the problems:

Linux:

image

Windows:

image

If this is the expected outcome we can merge the PR in.

jmezach:
@aaronpowell I believe the end state is correct, it is just the intermediate state while we're waiting for the SQL Server to become healthy that is somehow not showing.

ErikEJ (approved):

ErikEJ:
@jmezach I tested this - what happens is that the initial state of pending is never set, the child seems to inherit the parent state. 

Then the state changes to "Publishing" and then "Finished" - so agree with @aaronpowell that this works as intended!

LGTM!

aaronpowell:
Feel free to hit merge too - you don't have to wait for me 😉

Issue #444, opened 2025-02-03
OTEL export broken in GO Feature Flags


aaronpowell:
Did you get the OTEL exporter to work? @aaronpowell

_Originally posted by @Odonno in https://github.com/CommunityToolkit/Aspire/issues/425#issuecomment-2632393457_

aaronpowell:
I think the problem might be caused by the OTLE endpoint being served on a self-signed certificate from the Aspire app host, which then isn't trusted in the container. This is mostly a problem with container resources because executable resources run on the host which the aspnet dev cert is trusted, whereas inside the container it hasn't been trusted.

Odonno:
I trust you on that. I am also unsure if it is related to the private IP address, can't the OTEL endpoint be consumed via its service name?

aaronpowell:
The way the networking is configured with Docker the endpoint that it access it via, `host.docker.internal:<some port>`, is available within the container, but the SSL certificate for the OTEL endpoint isn't valid so it fails.

danegsta:
Just tested this with the new Aspire 13 support for automatically trusting the dev cert and telemetry is flowing automatically, even over https when `WithOtlpSupport()` is added to a go feature flags resource.

danegsta:
@aaronpowell this should only require adding `WithOtlpSupport()` to the defaults for `AddGoFeatureFlag`.

Odonno:
I still have the issue unfortunately. But now I strongly believe this might be due to Podman in some way.

Issue #388, opened 2025-01-14
Data API Builder OTEL doc


tommasodotNET:
### Page URL

https://learn.microsoft.com/en-us/dotnet/aspire/community-toolkit/hosting-data-api-builder

### Describe the issue

Since [this PR](https://github.com/Azure/data-api-builder/pull/2449) has been merged, we could improve DAB sample and add some documentation on how to correctly instrument the integration with OTEL.

### Suggestion

Example should include OTEL instrumentation and doc should explain how it works

### Help us help you

Yes, I'd like to be assigned to work on this item

tommasodotNET:
@aaronpowell should we start updating OUR readme and then align the official one on ms learn?

aaronpowell:
Shouldn't the integration just set that up without you having to do anything?

tommasodotNET:
Yes, but you need to configure the dab-config.json to use otel

aaronpowell:
Given that DAB supports layering config files, would it make sense for us to generate that config and add it as "the last" config file that would get loaded?

That way it's not something you have to worry about configuring, it just gets added for you "magically" (and maybe we add an opt-out path if people don't want OTEL enabled)

aaronpowell:
We need to work out what the process for doing OTEL into containers is, need to check the internal discussion as it seems there's some challenges still with the SSL certificate.

aaronpowell:
Yep, that'll be blocked by the SSL issue - I need to follow up with the Aspire team to see what the guidance is on that.

The problem is that the OTEL server running from Aspire uses the .NET development certificate, which is self-signed, so to container doesn't trust it, and in turn rejects it as an endpoint.

RobertK66:
Is there a possible workaround available yet? Would like to have a working example on that....

tommasodotNET:
> Is there a possible workaround available yet? Would like to have a working example on that....

the workaround i found is to run dab as an executable via aspire

danegsta:
I'm able to get telemetry flowing over https from a container with the new support for automatically trusting the dev cert in Aspire 13. Not sure how the container telemetry stacks up against the output when run from the CLI though.

danegsta:
Telemetry should just work with when properly configured in dab-config.json (with a variation on the config from https://github.com/CommunityToolkit/Aspire/issues/388#issuecomment-2746178745 to automatically set the service name):
```json
"telemetry": {
    "open-telemetry": {
        "enabled": true,
[...]

PR #603, opened 2025-03-26
Add OpenTelemetry Collector extension/component

### LINKED ISSUE #602: Add OpenTelemetry Collector extension
### .NET Aspire issue link

_No response_

### Overview

Add an integration to add the OpenTelemetry Collector into the pipeline between the apps and the dashboard.

### Usage example

```csharp
var collector = builder.AddOpenTelemetryCollector("collector" new OpenTelemetryCollectorSettings {
   Version = "latest"
   Image = "ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contrib"
})
.WithConfig("./config.yaml");
```

### Additional context

It needs to support SSL for the receivers.

### Help us help you

martinjt:
**Closes #602 **

## PR Checklist

- [x] New integration
  - [x] Docs are written


martinjt:
@dotnet-policy-service agree

martinjt:
I'm not sure how ready this is yet. I've been playing with the API a bit to give it enough flexibility around the certificates, images and config files.

Feedback very welcome, I wanted to get this up as is before I added any tests and worked on the documentation, etc.

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
```
aaronpowell, line 8:
```suggestion
```

This will all be provided by the props files above.

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/EnvironmentVariableHook.cs
```diff
@@ -0,0 +1,53 @@
...
+namespace CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector;
+
+internal class EnvironmentVariableHook : IDistributedApplicationLifecycleHook
+{
+    private readonly ILogger<EnvironmentVariableHook> _logger;
+
+    public EnvironmentVariableHook(ILogger<EnvironmentVariableHook> logger)
+    {
+        _logger = logger;
+    }
```
aaronpowell, line 14: Use a primary constructor

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/EnvironmentVariableHook.cs
```diff
@@ -0,0 +1,53 @@
...
+    public Task AfterEndpointsAllocatedAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken)
+    {
+        var resources = appModel.GetProjectResources();
+        var collectorResource = appModel.Resources.OfType<CollectorResource>().FirstOrDefault();
+
+        if (collectorResource == null)
```
aaronpowell, line 20:
```suggestion
        if (collectorResource is null)
```

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/EnvironmentVariableHook.cs
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        var endpoint = collectorResource!.GetEndpoint(collectorResource!.GRPCEndpoint.EndpointName);
+        if (endpoint == null)
+        {
+            _logger.LogWarning("No endpoint for the collector");
+            return Task.CompletedTask;
+        }
+
+        if (resources.Count() == 0)
```
aaronpowell, line 33:
```suggestion
        if (!resources.Any())
```

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/EnvironmentVariableHook.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            _logger.LogInformation("No resources to add Environment Variables to");
+        }
+
+        foreach (var resourceItem in resources)
+        {
+            _logger.LogDebug($"Forwarding Telemetry for {resourceItem.Name} to the collector");
+            if (resourceItem == null) continue;
```
aaronpowell, line 41:
```suggestion
            if (resourceItem is null) continue;
```

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/PublicAPI.Shipped.txt
aaronpowell, line 1: File no longer needed

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/PublicAPI.Unshipped.txt
aaronpowell, line 1: File no longer needed

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/DevCertHostingExtensions.cs
aaronpowell, line 1: We can likely use this in a few places, it should help unblock #388 and #444, so could it go into the `Shared` folder and be added to the csproj.

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/OpenTelemetryCollectorSettings.cs
```diff
@@ -0,0 +1,27 @@
...
+
+/// <summary>
+/// Settings for the OpenTelemetry Collector
+/// </summary>
+public class OpenTelemetryCollectorSettings
+{
+    /// <summary>
+    /// The version of the collector, defaults to latest
+    /// </summary>
+    public string CollectorVersion { get; set; } = "latest";
```
aaronpowell, line 11: We should pin to a version so it's known stable per the package version

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/OpenTelemetryCollectorSettings.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    /// <summary>
+    /// The image of the collector, defaults to ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contrib
+    /// </summary>
+    public string CollectorImage { get; set; } = "ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contrib";
```
aaronpowell, line 16: I feel like this should be split to registry/image like we do with other container resources. And is there a reason you're using ghcr.io not Docker Hub?

aaronpowell (commented):
Would it make sense that, rather than having a default collector image set, that it's a required property to be provided and there's a method that will set the defaults for you?

So, `AddOpenTelemetryCollector` is used if you want to specify a collector, whereas `AddOpenTelemetryDefaultCollection` uses the one that you have currently specified as a default.

martinjt:
Interesting approach, I'm honestly not sure, but maybe a `.WithDefaultImage()` might be better?

I was thinking that the approach I showed was more common, but happy if not.

I kinda wanted a minimal default command though.

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/OpenTelemetryCollectorSettings.cs
martinjt: I'm not convinced that pinning is the right approach for that, especially since there isn't a v1 yet. It's generally accepted that people don't pin even in production.

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/OpenTelemetryCollectorSettings.cs
martinjt: ghcr is the recommended default since the limits went into dockerhub.

Do you have an example of the pattern other integrations use for the registry split.

aaronpowell:
> Interesting approach, I'm honestly not sure, but maybe a `.WithDefaultImage()` might be better?
> 
> I was thinking that the approach I showed was more common, but happy if not.
> 
> I kinda wanted a minimal default command though.

The problem with a method like `WithDefaultImage` is that you could setup one without any image specified which would be a runtime error, which would be a less than desirable outcome.

martinjt:
I agree it's not ideal, but I'd prefer to not have them need to know what the image link is to collector. So I believe the best option is to use the default image we recommend in Otel right now (which is `latest`) when the extension is called without parameters.

martinjt:
I'm coming back at this now @aaronpowell, with a hope to move it forward so I don't need to maintain a separate one (happy to be codeowner here).

On the devcerts, it seems those issues haven't got a resolution, would it be ok to leave the devcerts in this project for now, and consolidate/move later? This project would actually fix both of those issues as the users could run the non-ssl collector and forward their data through the collector as a way to unblock them.

The big outstanding issue is the collector image model, can you let me know what you'd suggest there. It should be `latest` since there isn't a v1 of the collector yet, and the advice from the collector team is to always use latest.

aaronpowell:
> On the devcerts, it seems those issues haven't got a resolution, would it be ok to leave the devcerts in this project for now, and consolidate/move later? This project would actually fix both of those issues as the users could run the non-ssl collector and forward their data through the collector as a way to unblock them.

If you can pop it in the [`Shared` folder](https://github.com/CommunityToolkit/Aspire/tree/main/src/Shared) we can then add it as a linked file into several projects that could benefit from it in the CT as well.

> The big outstanding issue is the collector image model, can you let me know what you'd suggest there. It should be `latest` since there isn't a v1 of the collector yet, and the advice from the collector team is to always use latest.

What's the release frequency that they are going with? The main reason that we pin to a version is so that we don't risk any unexpected changes to command line args or other behaviour getting introduced that impact someone who hasn't opt-ed in to using a newer version. If they are releasing new versions daily, then pinned isn't ideal, but generally speaking, pinning a version is going to be preferred, and people can always specify a version override if we don't release an updated package quick enough (Ollama is a good example of where we get caught on that front).

martinjt:
The release cycle for the collector is every 2 weeks. The compatibility issues wouldn't be commandline args, its more likely the config of collector components that might change, but we haven't done a breaking change in about 6 months and those are expected to be the last.

The biggest issue with pinning a version is security issues, since the amount of components in the collector means that the updates contain those updates regularly.

I'll happily take guidance on what the best way to get this published is.

martinjt:
I've done a few major updates, so it probably needs a full re-review.

* Implemented inbound SSL, which can also be forced off (this helps fix some of the clients that can't use HTTPS because of the certificate issues)
* Outbound SSL was removed, it won't work until the Certificate supports `host.docker.internal` or the `*.localhost` stuff works.
* Moved to using `.WithConfig("config.yaml")`, I'm not sure how I'd fail startup if they don't provide that, I'm happy to implement it.
* `.WithConfig()` supports multiple configs
* Moved to using Random ports supplied by Aspire
* Added SSL certificates for HTTP and gRPC
* Made the Environment variables detect the protocol that the App wants and use that (useful as Node, by default, supports HTTP not gRPC)
* Support for both of the endpoint environment variable names `ASPIRE_DASHBOARD_OTLP_ENDPOINT_URL` and `DOTNET_DASHBOARD_OTLP_ENDPOINT_URL`
* Updated DevCerts to not try and add the certificates twice (this will only be useful when the Outbound SSL functionality is available.

This does require the user to implement `insecure_skip_verify`, but I can't find a way around that until the new certificate work to make the authority match.

src/Shared/DevCertHostingExtensions.cs
```diff
@@ -0,0 +1,199 @@
...
+    /// <summary>
+    /// Injects the ASP.NET Core HTTPS developer certificate into the resource via the specified environment variables when
+    /// <paramref name="builder"/>.<see cref="IResourceBuilder{T}.ApplicationBuilder">ApplicationBuilder</see>.<see cref="IDistributedApplicationBuilder.ExecutionContext">ExecutionContext</see>.<see cref="DistributedApplicationExecutionContext.IsRunMode">IsRunMode</see><c> == true</c>.<br/>
+    /// If the resource is a <see cref="ContainerResource"/>, the certificate files will be bind mounted into the container.
+    /// </summary>
+    /// <remarks>
+    /// This method <strong>does not</strong> configure an HTTPS endpoint on the resource.
+    /// Use <see cref="ResourceBuilderExtensions.WithHttpsEndpoint{TResource}"/> to configure an HTTPS endpoint.
+    /// </remarks>
+    public static IResourceBuilder<TResource> RunWithHttpsDevCertificate<TResource>(
```
davidfowl, line 34: We don't ship this in any of the intergrations as yet and I think we'd rewrite it at this point to behave a bit better given the advances in the model.

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/CollectorExtensions.cs
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /// <summary>
+    /// Force all apps to forward to the collector instead of the dashboard directly
+    /// </summary>
+    /// <param name="builder"></param>
+    /// <returns></returns>
+    public static IResourceBuilder<CollectorResource> WithAppForwarding(this IResourceBuilder<CollectorResource> builder)
+    {
+        builder.ApplicationBuilder.Services.TryAddLifecycleHook<EnvironmentVariableHook>();
```
aaronpowell: Lifecycle hooks are mostly considered legacy with the event system now in place. The hook should be replacable with the `OnAfterEndpointsAllocated` callback.

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/EnvironmentVariableHook.cs
```diff
@@ -0,0 +1,59 @@
...
+        var resources = appModel.GetProjectResources();
+        var collectorResource = appModel.Resources.OfType<CollectorResource>().FirstOrDefault();
+
+        if (collectorResource is null)
+        {
+            logger.LogWarning("No collector resource found");
+            return Task.CompletedTask;
+        }
+
+        var grpcEndpoint = collectorResource!.GetEndpoint(collectorResource!.GRPCEndpoint.EndpointName);
```
aaronpowell: Shouldn't need to force non-null here, the `if` clause above should catch that.

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/EnvironmentVariableHook.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            logger.LogDebug("Forwarding Telemetry for {name} to the collector", resourceItem.Name);
+            if (resourceItem is null) continue;
+
+            resourceItem.Annotations.Add(new EnvironmentCallbackAnnotation((EnvironmentCallbackContext context) =>
+            {
+                var protocol = context.EnvironmentVariables.GetValueOrDefault("OTEL_EXPORTER_OTLP_PROTOCOL", "");
+                var endpoint = protocol.ToString() == "http/protobuf" ? httpEndpoint : grpcEndpoint;
+
+                if (endpoint == null)
```
aaronpowell:
```suggestion
                if (endpoint is null)
```

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/OpenTelemetryCollectorSettings.cs
aaronpowell: https://github.com/CommunityToolkit/Aspire/blob/0c0afa121b5c4078310f1f13eca49cbe3c77c609/src/CommunityToolkit.Aspire.Hosting.Ollama/OllamaContainerImageTags.cs#L5-L6

src/Shared/DevCertHostingExtensions.cs
aaronpowell: Is there something "wrong" in the design of how this has been implemented that should be revised? Having a way to bootstrap the dev certificate into a container has been a pain point with a few integrations and having them working with OTEL (DAB and Go Feature Flags off the top of my head).

src/Shared/DevCertHostingExtensions.cs
```diff
@@ -0,0 +1,199 @@
...
+        this IResourceBuilder<TResource> builder, string certFileEnv, string certKeyFileEnv, Action<string, string>? onSuccessfulExport = null)
+        where TResource : IResourceWithEnvironment
+    {
+        if (builder.ApplicationBuilder.ExecutionContext.IsRunMode && builder.ApplicationBuilder.Environment.IsDevelopment())
+        {
+            builder.ApplicationBuilder.Eventing.Subscribe<BeforeStartEvent>(async (e, ct) =>
```
aaronpowell: Let's use the [`OnBeforeStart`](https://learn.microsoft.com/en-us/dotnet/api/aspire.hosting.lifecycle.idistributedapplicationlifecyclehook.beforestartasync?view=dotnet-aspire-9.0) callback as it is a cleaner API

aaronpowell (changes requested):
The main change here would be to use the event callbacks of 9.4.

Then would you be able to introduce an example project so I can get a better idea of how this works.

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/CollectorExtensions.cs
```diff
@@ -0,0 +1,105 @@
...
+using Aspire.Hosting.Lifecycle;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Hosting;
+
+namespace Aspire.Hosting;
+
+/// <summary>
+/// Extension methods to add the collector resource
+/// </summary>
+public static class CollectorExtensions
```
davidfowl: OpenTelemetryCollectorExtensions

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/CollectorResource.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Aspire.Hosting.ApplicationModel;
+
+namespace Aspire.Hosting;
+
+/// <summary>
+/// The collector resource
+/// </summary>
+/// <param name="name">Name of the resource</param>
+public class CollectorResource(string name) : ContainerResource(name)
```
davidfowl: OpenTelemetryCollectorResource

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/CollectorResource.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    internal static string GRPCEndpointName = "grpc";
+    internal static string HTTPEndpointName = "http";
+
+    /// <summary>
+    /// gRPC Endpoint
+    /// </summary>
+    public EndpointReference GRPCEndpoint => new(this, GRPCEndpointName);
```
davidfowl:
```suggestion
    public EndpointReference GrpcEndpoint => new(this, GRPCEndpointName);
```

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/CollectorResource.cs
```diff
@@ -0,0 +1,23 @@
...
+
+    /// <summary>
+    /// HTTP Endpoint
+    /// </summary>
+    public EndpointReference HTTPEndpoint => new(this, HTTPEndpointName);
```
davidfowl:
```suggestion
    public EndpointReference HttpEndpoint => new(this, HTTPEndpointName);
```

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/EnvironmentVariableHook.cs
```diff
@@ -0,0 +1,59 @@
...
+using Aspire.Hosting.Lifecycle;
+using Microsoft.Extensions.Logging;
+
+namespace Aspire.Hosting;
+
+/// <summary>
+/// Hooks to add the OTLP environment variables to the various containers
+/// </summary>
+/// <param name="logger"></param>
+public class EnvironmentVariableHook(ILogger<EnvironmentVariableHook> logger) : IDistributedApplicationLifecycleHook
```
davidfowl: On top of moving this implementation to the eventing model, we should also use this [OtlpExporterAnnotation](https://learn.microsoft.com/en-us/dotnet/api/aspire.hosting.applicationmodel.otlpexporterannotation?view=dotnet-aspire-9.0) to detect which resources should be impacted.

src/Shared/DevCertHostingExtensions.cs
martinjt: I can self-contain this inside of the Collector to reduce the impact? This was code taken from an example built by the Aspire team with a few amendments to allow some resilience.

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/CollectorExtensions.cs
martinjt: I'm trying to gather from the documentation whether the `OnAfterEndpointsAllocated` is resource specific, or part of the wider DistributedAppModel. My concern being that this needs to be applied after the Aspire code has added the dashboard environment variables.

martinjt:
I've done everything suggested other than moving the EnvironmentVariableHook. Since this needs to occur after all the resources had endpoints allocated, I'm struggling to see how I might achieve that.

I did have a thought of using OnAfterEndpointsAllocatedEvent, but that's deprecated, and then thought of using `OnResourceEndpointsAllocated` event on the collector resource, then access each resource to add that same event to add the endpoint, but I'm not sure how to access the eventing model from there?

src/Shared/DevCertHostingExtensions.cs
aaronpowell: It's `internal`, so it'll only be available in the assemblies it's packaged with, and I want to use it for others, so I'm find with where it is situated.

I'm more questioning @davidfowl's comment, as I'm not sure what he's looking for or suggesting.

aaronpowell:
Yes, sorry, I was mistaken, `OnResourceEndpointsAllocated` is the event that you'd want on the collector resource.

Is the logic that you'd have multiple collectors registered, or is it a singleton sort of thing?

martinjt:
The collector is a service that intercepts the OTLP telemetry from all the other services, so it intercepts all the other project resources and overrides the OTLP environment variable to point it to the collector instead.

For that reason, I thought that using the resource focused events wasn't something I could do.

This is instead of telling the user to access every resource and reference it to the collector.

martinjt:
@davidfowl let me know what you think of the new DevCerts code. I used yours as a base and updated it with optional parameters.

If thats OK, I'll build out the examples.

src/Shared/DevCertHostingExtensions.cs
```diff
@@ -0,0 +1,141 @@
...
+            }
+            if (!string.IsNullOrEmpty(certKeyFileEnv))
+            {
+                builder.WithEnvironment(certKeyFileEnv, certKeyFileDest);
+            }
+        }
+        else
+        {
+
+            // For non-container resources, set the file paths directly
```
davidfowl, line 137: Should this branch check certKeyFileEnv and certFileEnv as well?

martinjt:
Other than tests, I think this is pretty much there now @aaronpowell @davidfowl 

I've struggled with the tests because the main thing to test is the AppForwarding (testing that all OTHER resources have their OTEL endpoint environment variables updated to point to the collector). This means that the endpoints need to be allocated in the collector before the event is fired.

The only other tests I can think of that would be beneficial are the environment variables for the Dashboard endpoints, and the devcerts, I'm not sure how we can test those. 

If you've got pointers on the tests then please let me know!

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/EnvironmentVariableEvent.cs
```diff
@@ -0,0 +1,95 @@
...
+/// Hooks to add the OTLP environment variables to the various resources
+/// </summary>
+public static class EnvironmentVariableEventExtention
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <param name="builder"></param>
+    /// <returns></returns>
+    public static IResourceBuilder<OpenTelemetryCollectorResource> WithFirstStartup(this IResourceBuilder<OpenTelemetryCollectorResource> builder)
```
davidfowl: Why is this public API?

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/EnvironmentVariableEvent.cs
```diff
@@ -0,0 +1,95 @@
+using Aspire.Hosting.ApplicationModel;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+namespace Aspire.Hosting;
+
+/// <summary>
+/// Hooks to add the OTLP environment variables to the various resources
+/// </summary>
+public static class EnvironmentVariableEventExtention
```
davidfowl: These methods should be internal.

martinjt:
Looks like it should all be resolved now.

aaronpowell:
Tests need to be added to the workflow file, since we run a job-per-test. Check https://github.com/CommunityToolkit/Aspire/actions/runs/17362288128/job/49289169742?pr=603#step:4:11

martinjt:
I've added some more tests. I don't think we'll be able to test the DevCerts because it requires run mode. This is because the annotations being added are based on the output from the DevCerts extraction.

I suppose we could change the code to make that do something "dummy" during non-run mode? However, that doesn't feel like the right thing to do?

tests/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector.Tests/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector.Tests.csproj
```diff
@@ -6,6 +6,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <ProjectReference Include="../CommunityToolkit.Aspire.Testing/CommunityToolkit.Aspire.Testing.csproj" />
```
davidfowl, line 9: ..\\
../

😄

aaronpowell: Go complain to the editor teams - I added the reference via devkit on my devcontainer 😝

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/OpenTelemetryCollectorExtensions.cs
```diff
@@ -0,0 +1,178 @@
...
+        if (settings.EnableGrpcEndpoint)
+            resourceBuilder.WithEndpoint(targetPort: 4317, name: OpenTelemetryCollectorResource.GrpcEndpointName, scheme: isHttpsEnabled ? "https" : "http");
+        if (settings.EnableHttpEndpoint)
+            resourceBuilder.WithEndpoint(targetPort: 4318, name: OpenTelemetryCollectorResource.HttpEndpointName, scheme: isHttpsEnabled ? "https" : "http");
+
+
+        if (!settings.ForceNonSecureReceiver && isHttpsEnabled && builder.ExecutionContext.IsRunMode && builder.Environment.IsDevelopment())
+        {
+            resourceBuilder.RunWithHttpsDevCertificate();
+            var certFilePath = Path.Combine(DevCertHostingExtensions.DEV_CERT_BIND_MOUNT_DEST_DIR, DevCertHostingExtensions.CERT_FILE_NAME);
```
Worstjibs, line 56: I believe there is an issue with these two lines. On Windows, Path.Combine uses two backslashes, which causes the collector container to fall-over on startup, complaining about yaml format. I've tried replacing the line with a `string.Join('/', ...)` to force it to use forward slashes, which works correctly, e.g.

`var certFilePath = string.Join('/', DevCertHostingExtensions.DEV_CERT_BIND_MOUNT_DEST_DIR, DevCertHostingExtensions.CERT_FILE_NAME);`

Not the most elegant approach, but it has fixed the issue.

PR #630, opened 2025-04-13
Remove implicit `WaitFor` in `WithDbGate`

Alirexaa:
Let users explicitly use `WaitFor`.


src/CommunityToolkit.Aspire.Hosting.SqlServer.Extensions/CommunityToolkit.Aspire.Hosting.SqlServer.Extensions.csproj
```diff
@@ -17,6 +17,7 @@
   <!-- Workaround for https://github.com/dotnet/aspire/issues/7779 -->
   <ItemGroup Condition="'$(TargetFramework)' == 'net9.0'">
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" />
+    <PackageReference Include="Microsoft.Extensions.Options"></PackageReference>
```
Alirexaa, line 20: The build fails locally without this change.
build error

I wonder why CI does not fail.

CC @aaronpowell

aaronpowell: That's a weird one. Would have to run the nuget why command to trace the package refs probably and find the conflict.

Or maybe it's a VS quirk. I only dev on the repo in VS Code, which _should_ behave the same as VS, but I know there are quirks between them.

aaronpowell (approved):

PR #646, opened 2025-04-20
Update packages baseline

Alirexaa:
Should be merge after 9.4 release.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -37,8 +37,8 @@
 
   <!-- Versioning -->
   <PropertyGroup>
-    <ToolkitMinorVersion>5</ToolkitMinorVersion>
-    <ToolkitPatchVersion>0</ToolkitPatchVersion>
+    <ToolkitMinorVersion>4</ToolkitMinorVersion>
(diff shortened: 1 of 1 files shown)


Directory.Build.props
```diff
@@ -37,7 +37,7 @@
 
   <!-- Versioning -->
   <PropertyGroup>
-    <ToolkitMinorVersion>4</ToolkitMinorVersion>
+    <ToolkitMinorVersion>5</ToolkitMinorVersion>
```
aaronpowell, line 40: In the past I've just bumped the `ToolkitPatchVersion`, rather than the minor.

"Minor version" bumps are reserved for either new integrations or Aspire releases, whereas bugfixes/etc. go via patch versions (essentially we freeze the major with Aspire and then don't really follow semver)

aaronpowell (approved):

aaronpowell:
We've got two integrations that didn't publish with 9.4 (forgot to get them setup on NuGet properly), so once that's fixed I'll do an updated deployment

PR #706, opened 2025-05-31
Add PowerShell scripting support in Aspire AppHost

oising:
**Closes #<ISSUE_NUMBER>**

## PR Checklist

- [ ] New integration
  - [ ] Docs are written

### DIFF
--- a/examples/powershell/CommunityToolkit.Aspire.PowerShell.AppHost/Program.cs
+++ b/examples/powershell/CommunityToolkit.Aspire.PowerShell.AppHost/Program.cs
@@ -1,0 +1,39 @@
+using CommunityToolkit.Aspire.Hosting.PowerShell;
+
+var builder = DistributedApplication.CreateBuilder(args);
+
+var storage = builder.AddAzureStorage("storage").RunAsEmulator();
+var blob = storage.AddBlobs("myblob");
+
+var ps = builder.AddPowerShell("ps")
+    .WithReference(blob)
+    .WaitFor(storage);
+
+// uploads the script in scripts/
+var script1 = ps.AddScript("script1", """
+    param($name)
+
+    write-information "Hello, $name"
+
+    # uncommenting this will hang the script if you don't attach the pwsh debugger
+    # wait-debugger
+
+    write-information "`$myblob is $myblob"
+
+    az storage container create --connection-string $myblob -n demo
+    az storage blob upload --connection-string $myblob -c demo --file ./scripts/script.ps1
+    
+    write-information $pwd
+
+    write-information "Blob uploaded"
+""").WithArgs("world");
+
+// outputs "the sum of 2 and 3 is 5"
+var script2 = ps.AddScript("script2", """
+    & ./scripts/script.ps1 @args
+    """)
+    .WithArgs(2, 3)
+    .WaitForCompletion(script1);
+
+builder.Build().Run();
+
--- a/src/CommunityToolkit.Aspire.Hosting.PowerShell/CommunityToolkit.Aspire.Hosting.PowerShell.csproj
+++ b/src/CommunityToolkit.Aspire.Hosting.PowerShell/CommunityToolkit.Aspire.Hosting.PowerShell.csproj
@@ -1,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <AdditionalPackageTags>powershell; pwsh; scripting; script; hosting</AdditionalPackageTags>
+    <Description>Run powershell scripts in-process with your Aspire AppHost, injecting aspire resources and/or object instances as variables, using the command lines tools of your choice like azure cli, azd, or any other terminal tools.</Description>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Aspire.Hosting" />
+    <PackageReference Include="Microsoft.PowerShell.SDK" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <InternalsVisibleTo Include="CommunityToolkit.Aspire.Hosting.PowerShell.Tests" />
+  </ItemGroup>
+</Project>
--- a/src/CommunityToolkit.Aspire.Hosting.PowerShell/DistributedApplicationBuilderExtensions.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.PowerShell/DistributedApplicationBuilderExtensions.cs
@@ -1,0 +1,99 @@
+﻿using Aspire.Hosting;
+using Aspire.Hosting.ApplicationModel;
+using Microsoft.Extensions.DependencyInjection;
+using System.Diagnostics;
+using System.Globalization;
+using System.Management.Automation;
+using System.Management.Automation.Runspaces;
+
+namespace CommunityToolkit.Aspire.Hosting.PowerShell;
+
+/// <summary>
+/// Extensions for the <see cref="IDistributedApplicationBuilder"/> to add PowerShell runspace pool resources.
+/// </summary>
+public static class DistributedApplicationBuilderExtensions
+{
+    /// <summary>
+    /// Adds a PowerShell runspace pool resource to the distributed application.
+    /// </summary>
+    /// <param name="builder"></param>
+    /// <param name="name"></param>
+    /// <param name="languageMode"></param>
+    /// <param name="minRunspaces"></param>
+    /// <param name="maxRunspaces"></param>
+    /// <returns></returns>
+    /// <exception cref="ArgumentException"></exception>
+    /// <exception cref="DistributedApplicationException"></exception>
+    public static IResourceBuilder<PowerShellRunspacePoolResource> AddPowerShell(
+        this IDistributedApplicationBuilder builder,
+        [ResourceName] string name,
+        PSLanguageMode languageMode = PSLanguageMode.ConstrainedLanguage,
+        int minRunspaces = 1,
+        int maxRunspaces = 5)
+    {
+        if (string.IsNullOrWhiteSpace(name))
+            throw new ArgumentException("Value cannot be null or whitespace.", nameof(name));
+
+        if (builder.Resources.OfType<PowerShellRunspacePoolResource>().Any(
+            rsp => rsp.Name == name))
+        {
+            throw new DistributedApplicationException("AddPowerShell failed",
+                new InvalidOperationException(
+                    string.Format(
+                        CultureInfo.InvariantCulture,
+                        "A PowerShell resource with the name '{0}' already exists.", name)));
+        }
+
+        var pool = new PowerShellRunspacePoolResource(name, languageMode, minRunspaces, maxRunspaces);
+
+        builder.Eventing.Subscribe<AfterEndpointsAllocatedEvent>(async (e, ct) =>
+        {
+            var pools = e.Model.Resources.OfType<PowerShellRunspacePoolResource>().ToList();
+
+            foreach (var poolResource in pools)
+            {
+                Debug.Assert(poolResource is not null);
+
+                var loggerService = e.Services.GetRequiredService<ResourceLoggerService>();
+                var notificationService = e.Services.GetRequiredService<ResourceNotificationService>();
+
+                var sessionState = InitialSessionState.CreateDefault();
+
+                foreach (var annotation in poolResource.Annotations)
+                {
+                    if (annotation is PowerShellVariableReferenceAnnotation<ConnectionStringReference> reference)
+                    {
+                        var connectionString = await reference.Value.Resource.GetConnectionStringAsync(ct);
+                        sessionState.Variables.Add(
+                            new SessionStateVariableEntry(reference.Name, connectionString,
+                                $"ConnectionString for {reference.Value.Resource.GetType().Name} '{reference.Name}'",
+                                ScopedItemOptions.ReadOnly | ScopedItemOptions.AllScope));
+                    }
+                }
+
+                var poolName = poolResource.Name;
+                var poolLogger = loggerService.GetLogger(poolName);
+
+                _ = notificationService.WaitForDependenciesAsync(poolResource, ct)
+                    .ContinueWith(
+                        async _ =>
+                            await poolResource.StartAsync(sessionState, notificationService, poolLogger, ct),
+                        ct);
+            }
+        });
+
+        return builder.AddResource(pool)
+            .WithInitialState(new()
+            {
+                ResourceType = "PowerShellRunspacePool",
+                State = KnownResourceStates.NotStarted,
+                Properties = [
+
+                    new ("LanguageMode", pool.LanguageMode.ToString()),
+                    new ("MinRunspaces", pool.MinRunspaces.ToString()),
+                    new ("MaxRunspaces", pool.MaxRunspaces.ToString())
+                ]
+            })
+            .ExcludeFromManifest();
+    }
+}
--- a/src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellRunspacePoolResourceBuilderExtensions.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellRunspacePoolResourceBuilderExtensions.cs
@@ -1,0 +1,124 @@
+﻿using Aspire.Hosting;
+using Aspire.Hosting.ApplicationModel;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using System.Diagnostics.CodeAnalysis;
+using System.Management.Automation;
+
+namespace CommunityToolkit.Aspire.Hosting.PowerShell;
+
+/// <summary>
+/// Extensions for the PowerShellRunspacePoolResourceBuilder.
+/// </summary>
+public static class PowerShellRunspacePoolResourceBuilderExtensions
+{
+    /// <summary>
+    /// Adds a PowerShell script resource to the distributed application.
+    /// </summary>
+    /// <param name="builder"></param>
+    /// <param name="name"></param>
+    /// <param name="script"></param>
+    /// <returns></returns>
+    public static IResourceBuilder<PowerShellScriptResource> AddScript(
+        this IResourceBuilder<PowerShellRunspacePoolResource> builder,
+        [ResourceName] string name,
+        [StringSyntax("PowerShell")] string script)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(name);
+        ArgumentException.ThrowIfNullOrWhiteSpace(script);
+
+        // parse to force an early exception if the script is invalid
+        var scriptBlock = ScriptBlock.Create(script);
+
+        var scriptResource = new PowerShellScriptResource(name, scriptBlock, builder.Resource);
+
+        builder.ApplicationBuilder.Eventing.Subscribe<InitializeResourceEvent>(scriptResource, (e, ct) =>
+        {
+            var loggerService = e.Services.GetRequiredService<ResourceLoggerService>();
+            var notificationService = e.Services.GetRequiredService<ResourceNotificationService>();
+
+            var scriptName = scriptResource.Name;
+            var scriptLogger = loggerService.GetLogger(scriptName);
+            try
+            {
+                // TODO: capture script streams and log them
+                scriptLogger.LogInformation("Starting script '{ScriptName}'", scriptName);
+
+                _ = notificationService
+                    .WaitForDependenciesAsync(scriptResource, ct)
+                    .ContinueWith(
+                        async (_) => await scriptResource.StartAsync(scriptLogger, notificationService, ct),
+                        ct);
+            }
+            catch (Exception ex)
+            {
+                scriptLogger.LogError(ex, "Failed to start script '{ScriptName}'", scriptName);
+            }
+
+            return Task.CompletedTask;
+        });
+
+        return builder.ApplicationBuilder
+            .AddResource(scriptResource)
+            .WaitFor(builder) // wait for pool resource
+            .WithParentRelationship(builder.Resource) // owned by pool
+            .WithInitialState(new()
+            {
+                ResourceType = "PowerShellScript",
+                State = KnownResourceStates.NotStarted,
+                Properties = [
+                    new ("Script", script),
+                    new("RunspacePool", builder.Resource.Name)
+                ]
+            })
+            .ExcludeFromManifest()
+            .WithCommand("break", "Stop script execution",
+            async _ =>
+            {
+                await scriptResource.BreakAsync();
+                return CommandResults.Success();
+            },
+            new CommandOptions
+            {
+                ConfirmationMessage = "Are you sure you want to stop the script?",
+                Description = "Stop script execution",
+                IconName = "Stop",
+                IsHighlighted = true,
+                IconVariant = IconVariant.Filled,
+                UpdateState = updateContext =>
+                    updateContext.ResourceSnapshot.State?.Text != KnownResourceStates.Running ?
+                        ResourceCommandState.Disabled :
+                        ResourceCommandState.Enabled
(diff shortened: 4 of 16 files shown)


examples/powershell/CommunityToolkit.Aspire.PowerShell.AppHost/Program.cs
```diff
@@ -0,0 +1,39 @@
...
+    az storage blob upload --connection-string $myblob -c demo --file ./scripts/script.ps1
+    
+    write-information $pwd
+
+    write-information "Blob uploaded"
+""").WithArgs("world");
+
+// outputs "the sum of 2 and 3 is 5"
+var script2 = ps.AddScript("script2", """
+    & ./scripts/script.ps1 @args
```
aaronpowell, line 42: I think this example is a little brittle, as it's making some assumptions about paths that might not hold true. Instead, should the path to `script.ps1` be a resource that you pass in

src/CommunityToolkit.Aspire.Hosting.PowerShell/DistributedApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// <exception cref="DistributedApplicationException"></exception>
+    public static IResourceBuilder<PowerShellRunspacePoolResource> AddPowerShell(
+        this IDistributedApplicationBuilder builder,
+        [ResourceName] string name,
+        PSLanguageMode languageMode = PSLanguageMode.ConstrainedLanguage,
+        int minRunspaces = 1,
+        int maxRunspaces = 5)
+    {
+        if (string.IsNullOrWhiteSpace(name))
+            throw new ArgumentException("Value cannot be null or whitespace.", nameof(name));
```
aaronpowell:
```suggestion
        ArgumentNullException.ThrowIfNullOrEmpty(name);
```

src/CommunityToolkit.Aspire.Hosting.PowerShell/DistributedApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        if (builder.Resources.OfType<PowerShellRunspacePoolResource>().Any(
+            rsp => rsp.Name == name))
+        {
+            throw new DistributedApplicationException("AddPowerShell failed",
+                new InvalidOperationException(
+                    string.Format(
+                        CultureInfo.InvariantCulture,
+                        "A PowerShell resource with the name '{0}' already exists.", name)));
+        }
```
aaronpowell: We don't need to write this code, Aspire will take care of ensuring resources have unique names.

src/CommunityToolkit.Aspire.Hosting.PowerShell/DistributedApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        var pool = new PowerShellRunspacePoolResource(name, languageMode, minRunspaces, maxRunspaces);
+
+        builder.Eventing.Subscribe<AfterEndpointsAllocatedEvent>(async (e, ct) =>
```
aaronpowell: There needs to be some de-duping happening here, as this event handler will be subscribed multiple times if you do multiple `AddPowerShell` calls, and that means that you execute this logic more than it needs to.

src/CommunityToolkit.Aspire.Hosting.PowerShell/DistributedApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,99 @@
...
+                    }
+                }
+
+                var poolName = poolResource.Name;
+                var poolLogger = loggerService.GetLogger(poolName);
+
+                _ = notificationService.WaitForDependenciesAsync(poolResource, ct)
+                    .ContinueWith(
+                        async _ =>
+                            await poolResource.StartAsync(sessionState, notificationService, poolLogger, ct),
```
aaronpowell:
```suggestion
                    .ContinueWith(_ => poolResource.StartAsync(sessionState, notificationService, poolLogger, ct),
```

Since this expects you to return a `Task`, you don't need to `await` here, delegate the state machine higher up.

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellRunspacePoolResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,124 @@
...
+            var scriptLogger = loggerService.GetLogger(scriptName);
+            try
+            {
+                // TODO: capture script streams and log them
+                scriptLogger.LogInformation("Starting script '{ScriptName}'", scriptName);
+
+                _ = notificationService
+                    .WaitForDependenciesAsync(scriptResource, ct)
+                    .ContinueWith(
+                        async (_) => await scriptResource.StartAsync(scriptLogger, notificationService, ct),
```
aaronpowell:
```suggestion
                        _ => scriptResource.StartAsync(scriptLogger, notificationService, ct),
```

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellScriptResource.cs
```diff
@@ -0,0 +1,269 @@
...
+using Microsoft.Extensions.Logging;
+using System.Diagnostics;
+using System.Management.Automation;
+
+namespace CommunityToolkit.Aspire.Hosting.PowerShell
+{
+    /// <summary>
+    /// Represents a PowerShell script resource.
+    /// </summary>
+    public class PowerShellScriptResource : Resource, IDisposable,
```
aaronpowell, line 11: This should implement `IResourceWithParent<PowerShellRunspacePoolResource>`

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellRunspacePoolResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,124 @@
...
+            {
+                scriptLogger.LogError(ex, "Failed to start script '{ScriptName}'", scriptName);
+            }
+
+            return Task.CompletedTask;
+        });
+
+        return builder.ApplicationBuilder
+            .AddResource(scriptResource)
+            .WaitFor(builder) // wait for pool resource
```
aaronpowell: You don't need to do this if the `PowerShellScriptResource` is defined as a `IResourceWithParent` (see other comment) as this resource will know it's a child resource and it'll wait until the parent is ready.

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellRunspacePoolResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,124 @@
...
+                ]
+            })
+            .ExcludeFromManifest()
+            .WithCommand("break", "Stop script execution",
+            async _ =>
+            {
+                await scriptResource.BreakAsync();
+                return CommandResults.Success();
+            },
+            new CommandOptions
```
aaronpowell, line 79: Would it be worthwhile having a command to rerun the script too?

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellScriptResource.cs
```diff
@@ -0,0 +1,269 @@
...
+
+                        if (KnownResourceStates.TerminalStates.Contains(knownState))
+                        {
+                            state = state with
+                            {
+                                StopTimeStamp = DateTime.Now,
+                            };
+                        }
+
+                        return state;
```
aaronpowell: We can simplify this a little with a switch expression and some pattern matching.

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellScriptResource.cs
```diff
@@ -0,0 +1,269 @@
...
+            {
+                await notifications.PublishUpdateAsync(this, state => state with
+                {
+                    State = KnownResourceStates.Finished,
+                    StopTimeStamp = DateTime.Now,
+                    ExitCode = _ps.HadErrors ? 1 : 0,
+                });
+
+                logger.LogInformation("Output completed");
+            };
```
aaronpowell, line 219: If you invert the order of the methods in here you can return the call to `PublishUpdateAsync` and skip the async/await usage.

aaronpowell (changes requested):
Looks like a good start, there's some code changes I've got in there.

Can you also look at the unit tests as well as the integration tests, to match the test design of other integrations.

Also, it doesn't appear that the projects are in the `slnx` so they'll need to be added to ensure they are built.

Lastly, the PowerShell expects the Azure CLI to be installed, maybe we should have simpler scripts without those dependencies, otherwise the devcontainer + CI will need to be updated to ensure that the Azure CLI is installed.

examples/powershell/CommunityToolkit.Aspire.PowerShell.AppHost/Program.cs
oising: I don't understand what you mean -- I think you omitted a word or two :)

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellRunspacePoolResourceBuilderExtensions.cs
oising: @davidfowl This information (IResourceWithParent is implicit WaitFor) would be a good tidbit for your megagist.

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellRunspacePoolResourceBuilderExtensions.cs
oising: I did think about it, but there are some things to be aware of - namely that a runspace pool is somewhat stateful in terms of loaded modules, asssemblies etc. In 99.9% of cases this won't matter. That said, it might require a bit of a rearchitecture should we decide to add that feature (which I would like, tbh) so that I might incorporate a runspace reset.

https://github.com/PowerShell/PowerShell/issues/7131

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellScriptResource.cs
oising: Is this the general guideline now within MSFT? I tend to prefer using await at the callsite so my stacktraces are more readable, but I get it with respect to avoiding allocations of needless nested state machines.

examples/powershell/CommunityToolkit.Aspire.PowerShell.AppHost/Program.cs
oising: If you mean allow passing in a path to a ps1 file instead, I'd rather add that in the next point release. The point of this is to show that you're not blocked from doing so if you use pwsh native methods.

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellScriptResource.cs
oising: Ah, so KnownResourceStates isn't an enum. Switch expressions want constants on the LHS. I simplified it nonetheless.

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellScriptResource.cs
oising: btw, I think I neglected this because the UI nesting didn't work for custom resources before 9.3

examples/powershell/CommunityToolkit.Aspire.PowerShell.AppHost/Program.cs
oising: The reason being is that when you accept a file, you've now ventured into command invocation, so you must ensure your API surface allows binding named and unnamed parameters for completeness - right now, I'm fine with subsuming WithArgs for basic support for unbound and positional args - the reason being that one has full control of inline script parameterization.

aaronpowell: Ah ok - I'm not super deep in PowerShell.

My only concern is that how do you ensure that the `./scripts/script.ps1` is a valid path relative to the app host startup

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellScriptResource.cs
aaronpowell: Yep, that's true, I raised https://github.com/dotnet/aspire/issues/8020 and it was fixed for 9.3

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellScriptResource.cs
aaronpowell: No, I don't think it's explicit guidance from MS, more a preference on my part. When methods are really just doing a pass-through to another async method, I prefer the simplicity of not having extra state machines created.

But looking at this one again, it's probably needed, because I'm not sure if you can assign the lambda to the event handler without an async.

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellScriptResource.cs
oising: Hmm, okay, so I decided not to do this since scripts should be able to wait on each other. They need independent lifecycles.

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellRunspacePoolResourceBuilderExtensions.cs
oising: See other comment. Won't fix.

examples/powershell/CommunityToolkit.Aspire.PowerShell.AppHost/Program.cs
oising: By default, this example runs from the apphost root. Devs are free to use `cd` and/or use fully qualified paths if they want. It's in their hands.

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellScriptResource.cs
```diff
@@ -0,0 +1,258 @@
...
+            {
+                throw new ObjectDisposedException(nameof(PowerShellScriptResource));
+            }
+
+            if (_ps.InvocationStateInfo.State != PSInvocationState.Running)
+            {
+                return false;
+            }
+
+            await _cts.CancelAsync();
```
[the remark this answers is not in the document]
oising: I think you need a refresher course on C#, @copilot

https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.cancelasync

src/CommunityToolkit.Aspire.Hosting.PowerShell/DistributedApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(name);
+
+        var pool = new PowerShellRunspacePoolResource(name, languageMode, minRunspaces, maxRunspaces);
+
+        if (!_eventingSubscribed)
+        {
+            _eventingSubscribed = true;
+
+            builder.Eventing.Subscribe<InitializeResourceEvent>(pool, async (e, ct) =>
```
aaronpowell: Given this now uses a resource-scoped event, indicated by passing in the `pool` resource on the subscription, rather than a global event as previously done, I don't think the conditional check and `eventingSubscribed` field are useful.

If I was to do this:

```csharp
var ps1 = builder.AddPowerShell("ps1");
var ps2 = builder.AddPowerShell("ps2");
```

Only a single `InitializeResourceEvent` will be subscribed, and that'd be against `ps1`. This would mean that the second resource, `ps2`, won't have a session state created for it and it'll never have `StartAsync` invoked on the resource, which will mean that the associated scripts would fail right?

src/CommunityToolkit.Aspire.Hosting.PowerShell/DistributedApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+    public static IResourceBuilder<PowerShellRunspacePoolResource> AddPowerShell(
+        this IDistributedApplicationBuilder builder,
+        [ResourceName] string name,
+        PSLanguageMode languageMode = PSLanguageMode.ConstrainedLanguage,
+        int minRunspaces = 1,
+        int maxRunspaces = 5)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(name);
+
+        var pool = new PowerShellRunspacePoolResource(name, languageMode, minRunspaces, maxRunspaces);
```
aaronpowell, line 35: Since you're implementing everything from the ground up, you are going to need to handle the life cycle yourself, which does add a bit of overhead.

Another thing that you may want to add is setting the initial state of the resource, which can be done when creating the resource using this:

```csharp
            .WithInitialState(new CustomResourceSnapshot()
            {
                Properties = [new ResourcePropertySnapshot(CustomResourceKnownProperties.Source, "PowerShell")],
                ResourceType = "PowerShell",
                State = KnownResourceStates.Waiting
            })
```

src/CommunityToolkit.Aspire.Hosting.PowerShell/DistributedApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+                Debug.Assert(poolResource is not null);
+
+                var loggerService = e.Services.GetRequiredService<ResourceLoggerService>();
+                var notificationService = e.Services.GetRequiredService<ResourceNotificationService>();
+
+                var sessionState = InitialSessionState.CreateDefault();
+
+                // This will block until explicit and implied WaitFor calls are completed
+                await builder.Eventing.PublishAsync(
+                    new BeforeResourceStartedEvent(poolResource, e.Services), ct);
```
aaronpowell: I haven't considered raising event manually, but it possibly makes sense since you're having to manually control the life cycle.

Normally I use the `ResourceNotificationService` and publish status updates on there, but is the event going to perform the same sort of thing?

src/CommunityToolkit.Aspire.Hosting.PowerShell/DistributedApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+
+                foreach (var annotation in poolResource.Annotations)
+                {
+                    if (annotation is PowerShellVariableReferenceAnnotation<ConnectionStringReference> reference)
```
aaronpowell: This could be combined as:

```csharp
foreach (var annotation in poolResource.Anotations.OfType<PowerShellVariableReferenceAnnotation<ConnectionStringReference>>)
```

Couldn't it? that'd collapse the indentation block

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellScriptResource.cs
aaronpowell: While true that they could wait on each other, don't they also need to wait for the pool to be available, and they are always attached to a pool resource?

oising: Yes, but this is not the purpose of IResourceWithParent<T>. If the script resource implements that, then they lose the ability to be waited on since any waits are deferred to the parent instead.

src/CommunityToolkit.Aspire.Hosting.PowerShell/DistributedApplicationBuilderExtensions.cs
oising: Good catch!

src/CommunityToolkit.Aspire.Hosting.PowerShell/DistributedApplicationBuilderExtensions.cs
oising: Fairly sure I'm already doing that :)

src/CommunityToolkit.Aspire.Hosting.PowerShell/DistributedApplicationBuilderExtensions.cs
oising: The BeforeResourceStartedEvent is pretty much a mandatory task for custom resources in order for WaitFor to function correctly. The publish action for this event is blocking, and won't return until all WaitFor conditions on the context resource have resolved. If we don't do this, the InitializeResourceEvent fires too early and the scripts will try to run before the runspace is opened. Myself and afschrome had a long chat on discord about this. Migrating from lifecycle hooks to pure eventing was harder than expected.

davidfowl: Useful to read https://gist.github.com/davidfowl/b408af870d4b5b54a28bf18bffa127e1

src/CommunityToolkit.Aspire.Hosting.PowerShell/DistributedApplicationBuilderExtensions.cs
aaronpowell: Ah, I see it's at the end of the method - sorry, I tend to group everything on resource creation up top.

src/CommunityToolkit.Aspire.Hosting.PowerShell/DistributedApplicationBuilderExtensions.cs
aaronpowell: > Useful to read https://gist.github.com/davidfowl/b408af870d4b5b54a28bf18bffa127e1

PresidentNotToReadGIF

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellScriptResource.cs
```diff
@@ -0,0 +1,259 @@
...
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        /// <exception cref="ArgumentOutOfRangeException"></exception>
+        public async Task StartAsync(ILogger scriptLogger,
+            ResourceNotificationService notificationService,
+            CancellationToken cancellationToken = default)
+        {
+            scriptLogger.LogInformation("Starting PowerShell script '{ScriptName}'", Name);
+           
+            Debug.Assert(scriptLogger != null);
```
aaronpowell:
```suggestion
            Debug.Assert(scriptLogger is not null);
```

src/CommunityToolkit.Aspire.Hosting.PowerShell/PowerShellScriptResource.cs
```diff
@@ -0,0 +1,259 @@
...
+            _scriptLogger = scriptLogger;
+
+            Debug.Assert(Parent.Pool != null);
```
aaronpowell:
```suggestion
            Debug.Assert(Parent.Pool is not null);
```

aaronpowell (approved):

src/CommunityToolkit.Aspire.Hosting.PowerShell/CommunityToolkit.Aspire.Hosting.PowerShell.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
```
aaronpowell:
```suggestion
```

Don't need these are we set them in the `Directory.Build.props` - should have picked that up earlier.

PR #733, opened 2025-06-16
Add Go Build Tags Support

### LINKED ISSUE #730: Support build tags for Go programs

Yes

### Existing integration

Golang

### Overview

I am using Aspire to run a Go program that serves a Vue.js application and implements the API for the Vue.js application. I am using the [`embed`](https://pkg.go.dev/embed) package to embed the static Vue.js files into my Go application and then using the [`FileServeFS`](https://pkg.go.dev/net/http#FileServerFS) function to serve the files from the embedded file system.

When running my service with Aspire, instead of serving from the static file system, I would like to run the Vue.js development server as a child process and use Go's [`ReverseProxy`](https://pkg.go.dev/net/http/httputil#ReverseProxy) to forward the requests to the NPM development server. By doing this, during development and testing, I can make changes to my Vue.js application and the changes can be rebuilt and served without me having to stop and restart everything.

I am implementing this in my Go program using Go's [build tags](https://medium.com/@sealofthetime/go-build-tags-pros-cons-and-all-the-useful-things-you-can-do-with-them-91eaeebdc27b). In my case, I am using the `dev` build tag to indicate that I want to run the development server instead of embedding the static files in the program.

Currently, there is no way to send a build tag to the `go run` command that `GolangAppHostingExtension` executes. The optional `args` is used to pass command line arguments to the Go program that iss being executed. I would like to extend the `GolangAppHostingExtension.AddGolangApp` method to add an optional `buildTags` string array. If present, the tags would be joined into a comma-separated list and passed to the `go run` command using the `-tags` command line argument.

### Usage example

```go

builder.AddGolangApp("web", "../../web", buildTags: ["dev"])
```

### Breaking change?

No

### Alternatives

I am not aware of any alternative approaches.

### Additional context

_No response_

### Help us help you

mfcollins3:
**Closes #730**

I updated `GolangAppHostingExtension` to support specifying Go build tags to be used when building and running Go applications. Go build tags can be used to include or exclude files from the build process based On specific conditions. For example, when using Aspire for development, a Go developer may use a build tag to enable special code for OpenTelemetry instrumentation or enable debugging features.

I created a unit test to verify that the build tag was being passed in the command line to the `go` command.

## PR Checklist

- [ ] ~New integration~
  - [ ] ~Docs are written~
  - [ ] ~Added description of major feature to project description for NuGet package (4000 total character limit, so don't push entire description over that)~

## Other information

I accomplished my change by adding a new argument to the `GolangAppHostingExtension.AddGolangApp` method, but the new argument is marked as optional and has a default null value. This should not break existing consumers of the `AddGolandApp` method or make them alter how they call `AddGolangApp` in their code.

### DIFF
--- a/src/CommunityToolkit.Aspire.Hosting.Golang/GolangAppHostingExtension.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.Golang/GolangAppHostingExtension.cs
@@ -23,18 +23,19 @@ public static class GolangAppHostingExtension
         ArgumentException.ThrowIfNullOrWhiteSpace(name, nameof(name));
         ArgumentException.ThrowIfNullOrWhiteSpace(workingDirectory, nameof(workingDirectory));
 
-        string[] allArgs;
+        var allArgs = new List<string> { "run" };
+
         if (buildTags is { Length: > 0 })
         {
-            allArgs = args is { Length: > 0 }
-                ? ["run", "-tags", string.Join(",", buildTags), ".", .. args ?? []]
-                : ["run", "-tags", string.Join(",", buildTags), "."];
+            allArgs.Add("-tags");
+            allArgs.Add(string.Join(",", buildTags));
         }
-        else
+
+        allArgs.Add(".");
+
+        if (args is { Length: > 0 })
         {
-            allArgs = args is { Length: > 0 }
-                ? ["run", ".", .. args]
-                : ["run", ".",];
+            allArgs.AddRange(args);
         }
 
         workingDirectory = Path.Combine(builder.AppHostDirectory, workingDirectory).NormalizePathForCurrentPlatform();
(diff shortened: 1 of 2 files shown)


src/CommunityToolkit.Aspire.Hosting.Golang/GolangAppHostingExtension.cs
```diff
@@ -15,16 +15,27 @@ public static class GolangAppHostingExtension
...
+            allArgs = args is { Length: > 0 }
+                ? ["run", "-tags", string.Join(",", buildTags), ".", .. args ?? []]
+                : ["run", "-tags", string.Join(",", buildTags), "."];
+        }
+        else
+        {
+            allArgs = args is { Length: > 0 }
+                ? ["run", ".", .. args]
+                : ["run", ".",];
+        }
```
tommasodotNET: This check is good but could be improve using a different logic? It'd require to make allArgs a List, but would benefit in readability.
```csharp
var allArgs = new List<string> { "run" };

if (buildTags is { Length: > 0 })
{
    allArgs.Add("-tags");
    allArgs.Add(string.Join(",", buildTags));
}

allArgs.Add(".");

if (args is { Length: > 0 })
{
    allArgs.AddRange(args);
}
```

mfcollins3: Thanks. I made the change you requested.

PR #740, opened 2025-06-25
Node installer tweaks

### LINKED ISSUE #738: WithNpmPackageInstallation - easily add flags to pass to npm
### Describe the bug

In AspiriFridays 4 around 54 min, we wanted to add WithNpmPackageInstallation but it required a flag for legacy packages. There isn't a simple way to pass flags into the method.

### Regression

_No response_

### Steps to reproduce

```text
WithNpmPackageInstallation with a package.json with legacy deps
```

### Expected behavior

I can pass flags into withnpmpackageinstallation - maybe like (args: "--legacy")

### Screenshots

_No response_

### IDE and version

VS Code

### IDE version

_No response_

### Nuget packages

```text

```

### Additional context

_No response_

### Help us help you

aaronpowell:
Changed to using a callback rather than an array of args.

Fixes #738


aaronpowell:
@davidfowl What do you think of this - ditch the array and just give you a callback.

PR #770, opened 2025-07-30
Update to Aspire 9.4

aaronpowell:


src/CommunityToolkit.Aspire.Hosting.PostgreSQL.Extensions/PostgresBuilderExtensions.cs
```diff
@@ -141,22 +141,30 @@ private static void ConfigureDbGateContainer(EnvironmentCallbackContext context,
...
         var new_servers = postgresInstances.ToDictionary(
              postgresServer => postgresServer.Name,
-             postgresServer =>
+             async postgresServer =>
              {
-                 var user = postgresServer.UserNameParameter?.Value ?? "postgres";
+                 var user = postgresServer.UserNameParameter switch
+                 {
+                     null => "postgres",
+                     _ => await postgresServer.UserNameParameter.GetValueAsync(default)
```
davidfowl: context.CancellationToken

ErikEJ:
@aaronpowell Looks like we could do some changes with event handling (SQL Projects) - assume we can address that later?

aaronpowell:
> @aaronpowell Looks like we could do some changes with event handling (SQL Projects) - assume we can address that later?

Yeah, those sorts of changes are not going to impact cutting a release, but just clean up the API a bit, so no rush on them

Alirexaa (approved):

ErikEJ (approved):

PR #780, opened 2025-08-02
Support auth token on MCP Inspector

### LINKED ISSUE #779: McpInspector resource should not disable token authentication
### Describe the bug

In the current McpInspector resource there is an environment variable that disables the authentication via `DANGEROUSLY_OMIT_AUTH` -- this is not best practice and we should not have that be the default.

### Regression

_No response_

### Steps to reproduce

```text
Run the examples/mcp-inspector apphost project
```

### Expected behavior

The resource should:

- Provide a default token to the `MCP_PROXY_AUTH_TOKEN` environment variable before the resource starts
- Enable a `WithProxyToken` method on the builder that would enable someone to pass in the value for that token as a password parameter
- The dashboard URIs that are exposed would have the proxy token appended to them in the URI with a query string of the same name as the environment variable name

### Screenshots

_No response_

### IDE and version

VS 2022 Preview

### IDE version

_No response_

### Nuget packages

```text

```

### Additional context

_No response_

### Help us help you

timheuer:
**Closes #779**

Enhance MCP Inspector with proxy token support
- ✨ Add ProxyTokenParameter to McpInspectorResource
- 🔧 Update AddMcpInspector method to accept proxy token
- ✅ Implement authenticated health checks for server proxy endpoint
- 🧪 Add tests for proxy token generation and usage

## PR Checklist

- [ ] New integration
  - [ ] Docs are written


src/CommunityToolkit.Aspire.Hosting.McpInspector/McpInspectorResourceBuilderExtensions.cs
```diff
@@ -34,8 +36,46 @@ public static IResourceBuilder<McpInspectorResource> AddMcpInspector(this IDistr
...
+            .WithUrls(async context =>
+            {
+                var token = await proxyTokenParameter.GetValueAsync(CancellationToken.None);
+                
+                foreach (var url in context.Urls)
+                {
+                    if (url.Endpoint is not null)
+                    {
+                        var uriBuilder = new UriBuilder(url.Url);
+                        uriBuilder.Query = $"MCP_PROXY_AUTH_TOKEN={Uri.EscapeDataString(token!)}";
```
davidfowl: Hmmmmm, interesting. This might not be the best practice right?

cc @DamianEdwards

timheuer: It definitely feels dirty. No link and defer to getting it (the link) rom console log?

src/CommunityToolkit.Aspire.Hosting.McpInspector/McpInspectorResourceBuilderExtensions.cs
```diff
@@ -15,16 +16,17 @@ public static class McpInspectorResourceBuilderExtensions
     /// <param name="clientPort">The port for the client application. Defaults to 6274.</param>
     /// <param name="serverPort">The port for the server proxy application. Defaults to 6277.</param>
     /// <param name="inspectorVersion">The version of the Inspector app to use</param>
-    public static IResourceBuilder<McpInspectorResource> AddMcpInspector(this IDistributedApplicationBuilder builder, [ResourceName] string name, int clientPort = 6274, int serverPort = 6277, string inspectorVersion = McpInspectorResource.InspectorVersion)
+    /// <param name="proxyToken">The parameter used to provide the proxy authentication token for the MCP Inspector resource. If <see langword="null"/> a random token will be generated.</param>
+    public static IResourceBuilder<McpInspectorResource> AddMcpInspector(this IDistributedApplicationBuilder builder, [ResourceName] string name, int clientPort = 6274, int serverPort = 6277, string inspectorVersion = McpInspectorResource.InspectorVersion, IResourceBuilder<ParameterResource>? proxyToken = null)
```
aaronpowell: We're getting to the point that there's a lot of parameters being passed in now, so I think we should look to use an options object instead and deprecate the individual args.

src/CommunityToolkit.Aspire.Hosting.McpInspector/McpInspectorResourceBuilderExtensions.cs
timheuer: However, this is the URI that is emitted in the console...so it's not like this surfaces a secret of sorts that isn't surfaced already.

src/CommunityToolkit.Aspire.Hosting.McpInspector/McpInspectorResourceBuilderExtensions.cs
timheuer: Updated to options pattern in updated commit...obsoleted the verbose method with params)

aaronpowell (approved):
Thanks for the method refactor @timheuer - looks a lot cleaner now.

I'm good on the review front, just waiting on the comment from @davidfowl to see if there's any action we should take.

PR #790, opened 2025-08-06
feat(surrealdb): ensures ns/db are created in strict mode

Odonno:
Given that in strict mode, SurrealDB requires everything to be defined before use including Namespace, Database and Table definitions.

* Follow the new method definition called `WithCreationScript`
* Creates namespace and database when strict mode enabled (not required when disabled)

## PR Checklist


src/CommunityToolkit.Aspire.Hosting.SurrealDb/SurrealDbBuilderExtensions.cs
```diff
@@ -80,6 +82,59 @@ public static IResourceBuilder<SurrealDbServerResource> AddSurrealServer(
             : SurrealDbContainerImageTags.Tag;
         
         var surrealServer = new SurrealDbServerResource(name, userName?.Resource, passwordParameter);
+        
+        string? connectionString = null;
+
+        builder.Eventing.Subscribe<ConnectionStringAvailableEvent>(surrealServer, async (_, ct) =>
```
aaronpowell: Check out the new approach to resource life cycle events in 9.4 - https://learn.microsoft.com/en-us/dotnet/aspire/whats-new/dotnet-aspire-9.4#-resource-lifecycle-events

It should help simplify this code.

src/CommunityToolkit.Aspire.Hosting.SurrealDb/SurrealDbBuilderExtensions.cs
```diff
@@ -80,6 +82,59 @@ public static IResourceBuilder<SurrealDbServerResource> AddSurrealServer(
...
+        {
+            connectionString = await surrealServer.GetConnectionStringAsync(ct).ConfigureAwait(false);
+
+            if (connectionString == null)
```
aaronpowell:
```suggestion
            if (connectionString is null)
```

Issue #794, opened 2025-08-11
SurrealDB hosting tests unstable


aaronpowell:
After merging #790 the new tests seem to be failing on the `main` branch - https://github.com/CommunityToolkit/Aspire/actions/runs/16868590705

The `AddDatabaseCreatesDatabaseWithCustomScript` failed with this exception in the logs:

```
[2025-08-11 01:30:19Z] fail: CommunityToolkit.Aspire.Testing.Resources.surreal[0]
      83: 2025-08-11T01:30:19.4022291Z Failed to create namespace 'ns1'
      Websocket.Client.Exceptions.WebsocketException: Failed to start Websocket client , error: 'Unable to connect to the remote server'
       ---&gt; System.Net.WebSockets.WebSocketException (0x80004005): Unable to connect to the remote server
       ---&gt; System.Net.Http.HttpRequestException: An error occurred while sending the request.
       ---&gt; System.IO.IOException: Unable to read data from the transport connection: Connection reset by peer.
       ---&gt; System.Net.Sockets.SocketException (104): Connection reset by peer
         --- End of inner exception stack trace ---
         at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)
         at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.System.Threading.Tasks.Sources.IValueTaskSource&lt;System.Int32&gt;.GetResult(Int16 token)
         at System.Net.Http.HttpConnection.InitialFillAsync(Boolean async)
         at System.Net.Http.HttpConnection.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
         --- End of inner exception stack trace ---
         at System.Net.Http.HttpConnection.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
         at System.Net.Http.HttpConnectionPool.SendWithVersionDetectionAndRetryAsync(HttpRequestMessage request, Boolean async, Boolean doRequestAuth, CancellationToken cancellationToken)
         at System.Net.Http.RedirectHandler.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
         at System.Net.WebSockets.WebSocketHandle.ConnectAsync(Uri uri, HttpMessageInvoker invoker, CancellationToken cancellationToken, ClientWebSocketOptions options)
[...]

PR #795, opened 2025-08-11
fix(surrealdb): improve surrealdb healthchecks

Odonno:
#794 

Improve SurrealDB resources healthchecks to avoid consumption before initialized.

## PR Checklist

## Other information

Note: there is also an issue with some password generation (not correctly handled by the client). I am also trying to figure out a solution.


Odonno:
Not sure how to properly escape password but I suppose now it should work for a majority of scenario.

aaronpowell (approved):

aaronpowell:
Looks like there's still a failing test

Odonno:
I can't reproduce this, and I don't see the reason why it could fail. Isn't it a random error, what if we rerun the failed jobs?

aaronpowell:
Randomly failing tests are a concern because they either a) represent tests that aren't robust or b) represent a feature that isn't stable and could cause runtime errors to users.

Looking through the logs, it appears that the health check for databases is not passing, so I've pushed a commit with some additional log messages in the health check to see if we can get more insights.

aaronpowell:
The new logs are reporting this:

```
fail: CommunityToolkit.Aspire.SurrealDb.SurrealDbHealthCheck[0]
      SurrealDB health check raised an exception. Health check had previously reported: Healthy.
      System.TimeoutException: The operation has timed out.
         at SurrealDb.Net.Internals.SurrealDbWsEngine.SendRequestAsync(String method, Object[] parameters, SurrealDbWsRequestPriority priority, CancellationToken cancellationToken) in /_/SurrealDb.Net/Internals/SurrealDbEngine.Ws.cs:line 1535
         at SurrealDb.Net.Internals.SurrealDbWsEngine.SendRequestAsync(String method, Object[] parameters, SurrealDbWsRequestPriority priority, CancellationToken cancellationToken) in /_/SurrealDb.Net/Internals/SurrealDbEngine.Ws.cs:line 1546
         at SurrealDb.Net.Internals.SurrealDbWsEngine.SendRequestAsync(String method, Object[] parameters, SurrealDbWsRequestPriority priority, CancellationToken cancellationToken) in /_/SurrealDb.Net/Internals/SurrealDbEngine.Ws.cs:line 1546
         at SurrealDb.Net.Internals.SurrealDbWsEngine.SendRequestAsync(String method, Object[] parameters, SurrealDbWsRequestPriority priority, CancellationToken cancellationToken) in /_/SurrealDb.Net/Internals/SurrealDbEngine.Ws.cs:line 1546
         at SurrealDb.Net.Internals.SurrealDbWsEngine.RawQuery(String query, IReadOnlyDictionary`2 parameters, CancellationToken cancellationToken) in /_/SurrealDb.Net/Internals/SurrealDbEngine.Ws.cs:line 798
         at CommunityToolkit.Aspire.SurrealDb.SurrealDbHealthCheck.CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken) in /home/runner/work/Aspire/Aspire/src/CommunityToolkit.Aspire.SurrealDb/SurrealDbHealthCheck.cs:line 29
```

This would suggest that `_surrealdbClient.Health` returns `true`, but then the query line, `_surrealdbClient.RawQuery` is timing out, which is highly confusing 🤣

aaronpowell:
The stack trace says that the error is coming from here: https://github.com/surrealdb/surrealdb.net/blob/6b9057dd004b085276fe0c9019cd7d6950a30741/SurrealDb.Net/Internals/SurrealDbEngine.Ws.cs#L1535

Unfortunately, we don't have a stack trace on the `OperationCancelledException` that was the root cause of landing there, but ultimately, something in the stack doesn't like the raw query being run despite it "being healthy".

Odonno:
Ok, it seems the Namespace is not really created when the query is executed, kind of a delay between the command taking order and the action really fully executed. I made a simple loop to wait until it works, not sure what could be the best pattern here.

PR #810, opened 2025-08-26
feat(dapr): add IValueProvider support for component metadata configuration

### LINKED ISSUE #604: Dapr sidecar WaitFor

Yes

### Existing integration

CommunityToolkit.Aspire.Hosting.Dapr

### Overview

The dapr sidecar will start immediately while the redis container supporting pubsub for example is still starting.

The application has a WaitFor configured but the sidecar will fail the on pubsub component loading as redis is not done starting. When redis is started, the application will start and then fail because the sidecar has failed.

### Usage example

To make this play more nicely, the sidecar starting should either respect the application WaitFor configuration or be expanded to have its own.

additionally, I'm running the dapr scheduler and placement services using the dapr container images as well. Using this scenario, it would make more sense to have an additional WaitFor configuration on the sidecar configuration.

### Breaking change?

I'm not sure

### Alternatives

use persistent container runs and restart aspire host

### Additional context

_No response_

### Help us help you

No, just wanted to propose this

gabynevada:
## Summary
- Adds support for using `IValueProvider` (including `EndpointReference`) in Dapr component metadata configuration
- Enables dynamic endpoint resolution for Dapr components at runtime
- Maintains backward compatibility with existing string and ParameterResource overloads

## Changes
- Added new `WithMetadata` overload accepting `IValueProvider` for flexible metadata configuration
- Implemented deferred value resolution through environment variables and secretKeyRef mechanism
- Added `DaprComponentValueProviderAnnotation` to track value providers requiring resolution
- Updated lifecycle hook to resolve and inject endpoint values at runtime
- Enhanced `DaprComponentSchema` with async value resolution capabilities

## Test Plan
- [x] Added unit tests for EndpointReference metadata configuration
- [x] Added tests for value provider resolution in component schema
- [x] Added tests for multiple endpoint references
- [x] Updated example to demonstrate Redis endpoint usage with the new feature
- [x] Verified backward compatibility with existing metadata configuration methods

## Example Usage
```csharp
var redis = builder.AddRedis("redis");

var redisHost= redis.Resource.PrimaryEndpoint.Property(EndpointProperty.Host);
var redisTargetPort = redis.Resource.PrimaryEndpoint.Property(EndpointProperty.TargetPort);

var pubsub = builder
  .AddDaprPubSub("pubsub")
  .WithMetadata(
    "redisHost",
    ReferenceExpression.Create(
      $"{redisHost}:{redisTargetPort}"
    )
  );
```

This allows Dapr components to dynamically reference Aspire resource endpoints that are only known at runtime.

Closes #604


FullStackChef (approved):

src/Shared/Dapr/Core/IDaprSidecarResource.cs
```diff
@@ -8,6 +8,6 @@ namespace CommunityToolkit.Aspire.Hosting.Dapr;
 /// <summary>
 /// Represents a Dapr sidecar resource.
 /// </summary>
-public interface IDaprSidecarResource : IResource
+public interface IDaprSidecarResource : IResource, IResourceWithWaitSupport
```
davidfowl, line 11: Did you implement the waiting? To make this work, the resource needs a lifecycle, see https://github.com/dotnet/aspire/blob/main/docs/specs/appmodel.md#built-in-resources--lifecycle.

When you implement IResourceWithWaitSupport, at some point you have to raise https://github.com/dotnet/aspire/blob/364382724b4cd3590ec75010525ce36ef0b5e769/src/Aspire.Hosting/ConnectionStringBuilderExtensions.cs#L92

gabynevada: Implemented it, don't know if it's the correct approach but I followed the examples. Also implemented on the IDaprComponentResource as it seems it was missing support

gabynevada:
Here's the example:

https://github.com/user-attachments/assets/56e9284c-e333-485c-8753-1d3c6f5a2107

FullStackChef:
@gabynevada just checking the state of the PR - are you wanting this re-reviewed? is it ready to merge?

gabynevada:
@FullStackChef was occupied on work so could not do the final changes. Pushed the final changes so that the dapr component resources have a lifecycle and thus other resources can wait for them to get to the running state.

Let me know if anything needs to be changed!

PR #811, opened 2025-08-26
**Add Keycloak with Postgres integration**

max2020204:
This PR introduces a new integration project that makes it easier to configure Keycloak to run with a Postgres database in Aspire-based applications.

Summary of changes

Added new project CommunityToolkit.Aspire.Keycloak.Postgres.

Implemented an extension method AddPostgres for IResourceBuilder<KeycloakResource> to configure Keycloak with Postgres database resources:

Sets required environment variables (KC_DB, KC_DB_URL_*, KC_DB_USERNAME, KC_DB_PASSWORD, KC_TRANSACTION_XA_ENABLED).

Includes validation for missing or invalid Postgres server configuration.

Updated solution and package references to align with Aspire 9.4.1 dependencies.

- [X] New integration
  - [X] Docs are written

Other information

Project name normalized to Postgres for correctness and consistency.

Extension method returns IResourceBuilder<KeycloakResource> to allow chaining.

Uses proper IValueExpression instead of ToString() for connection strings.

Ensures JDBC-compatible environment variables for Keycloak.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -4,12 +4,13 @@
   </PropertyGroup>
   <ItemGroup Label="Aspire Packages">
     <!-- Aspire packages -->
-    <PackageVersion Include="Aspire.Hosting" Version="$(AspireVersion)" />
+    <PackageVersion Include="Aspire.Hosting" Version="9.4.1" />
     <PackageVersion Include="Aspire.Hosting.AppHost" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.Azure.Storage" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.Dapr" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.Azure.AppContainers" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.Azure.Redis" Version="$(AspireVersion)" />
+    <PackageVersion Include="Aspire.Hosting.Keycloak" Version="9.4.1-preview.1.25408.4" />
     <PackageVersion Include="Aspire.Hosting.NodeJS" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.PostgreSQL" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.Python" Version="$(AspireVersion)" />
--- a/src/CommunityToolkit.Aspire.Keycloak.Postgress/CommunityToolkit.Aspire.Keycloak.Postgress.csproj
+++ b/src/CommunityToolkit.Aspire.Keycloak.Postgress/CommunityToolkit.Aspire.Keycloak.Postgress.csproj
@@ -1,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net9.0</TargetFramework>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <PackageReference Include="Aspire.Hosting" />
+      <PackageReference Include="Aspire.Hosting.Keycloak" />
+      <PackageReference Include="Aspire.Hosting.PostgreSQL" />
+    </ItemGroup>
+
+</Project>
--- a/src/CommunityToolkit.Aspire.Keycloak.Postgress/KeycloakPostgresExtension.cs
+++ b/src/CommunityToolkit.Aspire.Keycloak.Postgress/KeycloakPostgresExtension.cs
@@ -1,0 +1,56 @@
+﻿using Aspire.Hosting;
+using Aspire.Hosting.ApplicationModel;
+
+namespace CommunityToolkit.Aspire.Keycloak.Postgress;
+
+/// <summary>
+/// Provides extension methods for configuring Keycloak with a Postgres database
+/// in a resource builder.
+/// </summary>
+public static class KeycloakPostgresExtension
+{
+    /// <summary>
+    /// Configures Keycloak to use a Postgres database as its underlying storage
+    /// by adding the necessary environment variables to the resource builder.
+    /// </summary>
+    /// <param name="builder">
+    /// The resource builder for the Keycloak resource being configured.
+    /// </param>
+    /// <param name="database">
+    /// The resource builder representing the Postgres database resource to be used.
+    /// </param>
+    /// <param name="transaction">
+    /// A boolean flag indicating whether XA transaction support should be enabled.
+    /// Defaults to true.
+    /// </param>
+    /// <exception cref="ArgumentNullException">
+    /// Thrown when the <paramref name="builder"/> or <paramref name="database"/> is null.
+    /// </exception>
+    /// <exception cref="ArgumentException">
+    /// Thrown when the parent resource of the Postgres database is not a valid
+    /// Postgres server or when username/password parameters are not defined for the
+    /// Postgres server.
+    /// </exception>
+    public static void AddPostgres(this IResourceBuilder<KeycloakResource> builder,
+        IResourceBuilder<PostgresDatabaseResource> database, bool transaction = true)
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentNullException.ThrowIfNull(database);
+
+        if (database.Resource.Parent is null)
+            throw new ArgumentException("The Postgres database must be hosted by a Postgres server.",
+                nameof(database.Resource.Parent));
+        if (database.Resource.Parent.UserNameParameter is null)
+            throw new ArgumentException("The Postgres database must have a username parameter.",
+                nameof(database.Resource.Parent.UserNameParameter));
+        if (database.Resource.Parent.PasswordParameter is null)
+            throw new ArgumentException("The Postgres database must have a password parameter.",
+                nameof(database.Resource.Parent.PasswordParameter));
+
+        builder.WithEnvironment("KC_DB", "postgres")
+            .WithEnvironment("KC_DB_URL", database.Resource.ConnectionStringExpression)
+            .WithEnvironment("KC_DB_USERNAME", database.Resource.Parent.UserNameParameter.ValueExpression)
+            .WithEnvironment("KC_DB_PASSWORD", database.Resource.Parent.PasswordParameter.ValueExpression)
+            .WithEnvironment("KC_TRANSACTION_XA_ENABLED", transaction.ToString().ToLowerInvariant());
+    }
+}
--- a/CommunityToolkit.Aspire.slnx
+++ b/CommunityToolkit.Aspire.slnx
@@ -145,8 +145,8 @@
   </Folder>
   <Folder Name="/examples/surrealdb/">
     <Project Path="examples/surrealdb/CommunityToolkit.Aspire.Hosting.SurrealDb.ApiService/CommunityToolkit.Aspire.Hosting.SurrealDb.ApiService.csproj" />
-    <Project Path="examples/surrealdb/CommunityToolkit.Aspire.Hosting.SurrealDb.ServiceDefaults/CommunityToolkit.Aspire.Hosting.SurrealDb.ServiceDefaults.csproj" />
     <Project Path="examples/surrealdb/CommunityToolkit.Aspire.Hosting.SurrealDb.AppHost/CommunityToolkit.Aspire.Hosting.SurrealDb.AppHost.csproj" />
+    <Project Path="examples/surrealdb/CommunityToolkit.Aspire.Hosting.SurrealDb.ServiceDefaults/CommunityToolkit.Aspire.Hosting.SurrealDb.ServiceDefaults.csproj" />
   </Folder>
   <Folder Name="/src/">
     <Project Path="src/CommunityToolkit.Aspire.EventStore/CommunityToolkit.Aspire.EventStore.csproj" />
@@ -191,6 +191,7 @@
     <Project Path="src/CommunityToolkit.Aspire.OllamaSharp/CommunityToolkit.Aspire.OllamaSharp.csproj" />
     <Project Path="src/CommunityToolkit.Aspire.RavenDB.Client/CommunityToolkit.Aspire.RavenDB.Client.csproj" />
     <Project Path="src/CommunityToolkit.Aspire.SurrealDb/CommunityToolkit.Aspire.SurrealDb.csproj" />
+    <Project Path="src\CommunityToolkit.Aspire.Keycloak.Postgress\CommunityToolkit.Aspire.Keycloak.Postgress.csproj" Type="Classic C#" />
   </Folder>
   <Folder Name="/src/Dapr/">
     <Project Path="src/CommunityToolkit.Aspire.Hosting.Azure.Dapr.Redis/CommunityToolkit.Aspire.Hosting.Azure.Dapr.Redis.csproj" />


davidfowl:
Did you test this?

axies20:
I have some problems, i can't access host, can you help me?

axies20:
Now it working

davidfowl:
Its worth reading this https://github.com/dotnet/aspire/blob/main/docs/specs/appmodel.md#values-and-references

Directory.Packages.props
```diff
@@ -4,12 +4,13 @@
   </PropertyGroup>
   <ItemGroup Label="Aspire Packages">
     <!-- Aspire packages -->
-    <PackageVersion Include="Aspire.Hosting" Version="$(AspireVersion)" />
+    <PackageVersion Include="Aspire.Hosting" Version="9.4.1" />
```
aaronpowell: I assume this is just done when you were testing, but ensure you roll that back and use the MSBuild variable

aaronpowell:
Once there are tests and samples, we can get started on a review of the work.

davidfowl:
This implemenaton needs work:
- You don't need to handle the event to make it work
- This will only work when running but not when publishing.

See https://github.com/dotnet/aspire/issues/8034 for more details.

src/CommunityToolkit.Aspire.Keycloak.Postgress/CommunityToolkit.Aspire.Keycloak.Postgress.csproj
Alirexaa, line 1: The package name should be `CommunityToolkit.Aspire.Keycloak.Extensions`, like other extension packages.

src/CommunityToolkit.Aspire.Keycloak.Postgress/KeycloakPostgresExtension.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// The resource builder for the Postgres database resource. It provides connection string
+    /// information and other database-related configurations.
+    /// </param>
+    /// <param name="transaction">
+    /// A boolean indicating whether transactions should be enabled in the configuration. Default is true.
+    /// </param>
+    /// <param name="port">
+    /// The port number on which the Postgres database is running. Default is 5432.
+    /// </param>
+    public static void AddPostgrctes(this IResourceBuilder<KeycloakResource> builder,
```
Alirexaa: This should be `With` instead of `Add`; we don't add a resource to the appmodel we are using an existing one. Also, there is a typo in the method name

```suggestion
    public static void WithPostgres(this IResourceBuilder<KeycloakResource> builder,
```

Directory.Packages.props
axies20: I had error
0>CommunityToolkit.Aspire.Keycloak.Postgres.csproj: Error NU1605 : Warning As Error: Detected package downgrade: Aspire.Hosting from 9.4.1 to 9.4.0. Reference the package directly from the project to select a different version. 
 CommunityToolkit.Aspire.Keycloak.Postgres -> Aspire.Hosting.Keycloak 9.4.1-preview.1.25408.4 -> Aspire.Hosting (>= 9.4.1) 
 CommunityToolkit.Aspire.Keycloak.Postgres -> Aspire.Hosting (>= 9.4.0)
0>------- Finished building project: CommunityToolkit.Aspire.Keycloak.Postgres. Succeeded: False. Errors: 1. Warnings: 0

src/CommunityToolkit.Aspire.Keycloak.Postgress/CommunityToolkit.Aspire.Keycloak.Postgress.csproj
axies20: Can i name it like `CommunityToolkit.Aspire.Keycloak.Extensions.Postgres`?

axies20: I think there may be other providers here.

axies20:
I updated, add event as dev and add parameters for production

src/CommunityToolkit.Aspire.Keycloak.Extensions.Postgres/KeycloakPostgresExtension.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// <param name="database">
+    /// The resource builder for the Postgres database resource.
+    /// </param>
+    /// <param name="app">
+    /// The distributed application builder for adding parameters like username and password for the Postgres database.
+    /// </param>
+    /// <returns>
+    /// The updated resource builder for the Keycloak resource.
+    /// </returns>
+    private static void WithPostgres(this IResourceBuilder<KeycloakResource> builder,
```
davidfowl: This will work with dev *and* production.

src/CommunityToolkit.Aspire.Keycloak.Extensions.Postgres/KeycloakPostgresExtension.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// <param name="database">
+    /// The resource builder for the Postgres database resource.
+    /// </param>
+    /// <param name="port">
+    /// The port for connecting to the Postgres database. Defaults to 5432.
+    /// </param>
+    /// <returns>
+    /// The updated resource builder for the Keycloak resource.
+    /// </returns>
+    public static IResourceBuilder<KeycloakResource> WithPostgresDev(this IResourceBuilder<KeycloakResource> builder,
```
davidfowl: You don't need this overload at all.

src/CommunityToolkit.Aspire.Keycloak.Extensions.Postgres/KeycloakPostgresExtension.cs
```diff
@@ -0,0 +1,104 @@
...
+        PostgresServerResource pgServer = database.Resource.Parent;
+        EndpointReference ep = pgServer.GetEndpoint("tcp");
+
+        string dbName = database.Resource.Name;
+
+        ReferenceExpression jdbcUrl = ReferenceExpression.Create(
+            $"jdbc:postgresql://{ep.Property(EndpointProperty.Host)}:" +
+            $"{ep.Property(EndpointProperty.Port)}/{dbName}");
+
+        builder.WithEnvironment("KC_DB", "postgres")
```
davidfowl: Is this correct? Why is this hard coded?

davidfowl: What happens if postgres is password protected (which is the default).

src/CommunityToolkit.Aspire.Keycloak.Extensions.Postgres/KeycloakPostgresExtension.cs
axies20: You mean void? After that, you can't add other methods, so I returned IResourceBuilder<KeycloakResource>

src/CommunityToolkit.Aspire.Keycloak.Extensions.Postgres/KeycloakPostgresExtension.cs
axies20: You said that event will not work in prod, so I thought that this i will leave for the development environment

src/CommunityToolkit.Aspire.Keycloak.Extensions.Postgres/KeycloakPostgresExtension.cs
axies20: `builder.WithEnvironment("KC_DB", "postgres")`
You about this? Yeah, it's correct. There are other database vendors, but I'm writing for postgres
All environment variables: https://www.keycloak.org/server/all-config
I wanna make for other vendors such as `dev-file (default), dev-mem, mariadb, mssql, mysql, oracle, postgres`  (all vendors that support)

axies20: > What happens if postgres is password protected (which is the default).

```csharp
public static IResourceBuilder<KeycloakResource> WithPostgres(this IResourceBuilder<KeycloakResource> builder,
        IResourceBuilder<PostgresDatabaseResource> database, ParameterResource username, ParameterResource password)
    {
        ArgumentNullException.ThrowIfNull(username);
        ArgumentNullException.ThrowIfNull(password);
        WithPostgres(builder, database);
        builder.WithEnvironment("KC_DB_USERNAME", username)
            .WithEnvironment("KC_DB_PASSWORD", password);
        return builder;
    }
```

Here I'm setting by parameter that will set up from user, or what you mean?

src/CommunityToolkit.Aspire.Keycloak.Extensions.Postgres/KeycloakPostgresExtension.cs
davidfowl: No, I mean that this approach using expressions works in both cases and you don’t need 2 implementations for run and publish

src/CommunityToolkit.Aspire.Keycloak.Extensions.Postgres/KeycloakPostgresExtension.cs
davidfowl: Right so you don’t need this

src/CommunityToolkit.Aspire.Keycloak.Extensions.Postgres/CommunityToolkit.Aspire.Keycloak.Extensions.Postgres.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net9.0</TargetFramework>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
```
aaronpowell: Not required as this is inherited.

There are some required properties for description and tags.

src/CommunityToolkit.Aspire.Keycloak.Extensions.Postgres/KeycloakPostgresExtension.cs
aaronpowell: There can be conditional logic to check for the different modes rather than having to call different methods.

tests/CommunityToolkit.Aspire.Keycloak.Extensions.Postgres.Tests/CommunityToolkit.Aspire.Keycloak.Extensions.Postgres.Tests.csproj
```diff
@@ -0,0 +1,26 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net9.0</TargetFramework>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+        <IsPackable>false</IsPackable>
+    </PropertyGroup>
```
aaronpowell: These aren't needed as they are inherited. Check the other test projects and follow their patterns

tests/CommunityToolkit.Aspire.Keycloak.Extensions.Postgres.Tests/CommunityToolkit.Aspire.Keycloak.Extensions.Postgres.Tests.csproj
```diff
@@ -0,0 +1,26 @@
...
+
+    <ItemGroup>
+        <PackageReference Include="Aspire.Hosting.Keycloak" />
+        <PackageReference Include="coverlet.collector" Version="6.0.2"/>
+        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.12.0"/>
+        <PackageReference Include="xunit" Version="2.9.2"/>
+        <PackageReference Include="xunit.runner.visualstudio" Version="2.8.2"/>
```
aaronpowell: These will fail the build as CPM dictates versions

Directory.Packages.props
aaronpowell: The `main` branch is now updated to 9.4.1 Aspire.

Directory.Packages.props
```diff
@@ -4,12 +4,13 @@
...
     <PackageVersion Include="Aspire.Hosting.AppHost" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.Azure.Storage" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.Dapr" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.Azure.AppContainers" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.Azure.Redis" Version="$(AspireVersion)" />
+    <PackageVersion Include="Aspire.Hosting.Keycloak" Version="9.4.1-preview.1.25408.4" />
```
aaronpowell: We'll need to introduce a MSBuild variable for the preview version of Aspire packages.

Directory.Packages.props
```diff
@@ -18,6 +19,7 @@
     <PackageVersion Include="Aspire.Hosting.MongoDB" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.MySql" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.SqlServer" Version="$(AspireVersion)" />
+    <PackageVersion Include="Moq" Version="4.20.72" />
```
aaronpowell: This should be under the `Testing` item group

aaronpowell (changes requested):
This is a hosting integration by the looks of it, so the project should reflect that.

Also, let's just make it `Keycloak.Extensions` so it's not PG specific, as that'll allow for other extensions on Keycloack to be included in here rather than going through a bunch of micro packages.

axies20:
```csharp
 public static IResourceBuilder<KeycloakResource> WithPostgres(this IResourceBuilder<KeycloakResource> builder,
        IDistributedApplicationBuilder appBuilder, bool xaEnabled = false,
        string usernameParameter = "keycloak-username", string databaseName = "keycloak-db",
        string postgrsName = "keycloak-postgres")
    {
        var username = appBuilder.AddParameter(usernameParameter, "keycloak-db-user");
        var pwd = ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(appBuilder, "pg-pass");
        var password = appBuilder.CreateResourceBuilder(pwd);
        var keycloakPostgres = appBuilder.AddPostgres(postgrsName, username, password);
        var db = keycloakPostgres.AddDatabase(databaseName);

        builder.WithPostgres(db, username, password, xaEnabled)
            .WithReference(db)
            .WaitFor(keycloakPostgres);
        return builder;
    }
```
What you think about this method for easy dev environment?

tests/CommunityToolkit.Aspire.Keycloak.Extensions.Postgres.Tests/CommunityToolkit.Aspire.Keycloak.Extensions.Postgres.Tests.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  
+  
+
```
davidfowl:
```suggestion
```

src/CommunityToolkit.Aspire.Keycloak.Extensions/KeycloakPostgresExtension.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        ReferenceExpression jdbcUrl = ReferenceExpression.Create(
+            $"jdbc:postgresql://{ep.Property(EndpointProperty.Host)}:" +
+            $"{ep.Property(EndpointProperty.Port)}/{dbName}");
+
+        builder.WithEnvironment("KC_DB", "postgres")
+            .WithEnvironment("KC_DB_URL", jdbcUrl);
+
+        builder.WithEnvironment("KC_DB_USERNAME", username.Resource);
+        builder.WithEnvironment("KC_DB_PASSWORD", password.Resource);
```
davidfowl: Can you get these from the database.Parent instead? This overload is about using an existing postgres right?

axies20: Yeah, but i need to make task method with async/await. Will it be fine?

davidfowl: You don't need a task. Look at how the existing postgres integrations work in this project e.g. https://github.com/CommunityToolkit/Aspire/blob/c7dedf100938dcdcc84f8348ea76032f1fd180b5/src/CommunityToolkit.Aspire.Hosting.PostgreSQL.Extensions/PostgresBuilderExtensions.cs#L38

You can see this dbgate extension resolving the username password and endpoints:

https://github.com/CommunityToolkit/Aspire/blob/c7dedf100938dcdcc84f8348ea76032f1fd180b5/src/CommunityToolkit.Aspire.Hosting.PostgreSQL.Extensions/PostgresBuilderExtensions.cs#L108-L119

Since you care about this working when deployed as well, you don't need to change the endpoint resolution logic, but you can get the username and password from the postgres server.

axies20: Oh, i see

Alirexaa (commented):
This PR needs an example (like other integrations) and integration tests.

axies20:
> This PR needs an example (like other integrations) and integration tests.

I added integration tests and there readme file with examples already exists

davidfowl:
@max2020204 I think @Alirexaa means a runnable sample in the playground, not just a readme.

axies20:
> @max2020204 I think @Alirexaa means a runnable sample in the playground, not just a readme.

Oh, Okey

max2020204:
> @max2020204 I think @Alirexaa means a runnable sample in the playground, not just a readme.

Can you help? Unfortunately, I don't know why my project's web APIs failed to run.

tests/CommunityToolkit.Aspire.Keycloak.Extensions.Tests/KeycloakExtensionTests.cs
```diff
@@ -0,0 +1,126 @@
...
+namespace CommunityToolkit.Aspire.Keycloak.Extensions.Tests;
+
+public class KeycloakExtensionTests
+{
+    private readonly IDistributedApplicationBuilder _app;
+
+    public KeycloakExtensionTests()
+    {
+        _app = DistributedApplication.CreateBuilder();
+    }
```
aaronpowell: No need to create this in the constructor, just create it at the point it's needed.

Directory.Packages.props
```diff
@@ -18,6 +19,7 @@
     <PackageVersion Include="Aspire.Hosting.MongoDB" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.MySql" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.SqlServer" Version="$(AspireVersion)" />
+    <PackageVersion Include="Aspire.Keycloak.Authentication" Version="9.4.1-preview.1.25408.4" />
```
aaronpowell: We'll need to introduce a MSBuild variable for the preview version of Aspire packages.

axies20:
> Since this is a hosting integration, can you correct the library name to reflect that - it should be `CommunityToolkit.Aspire.Keycloak.Hosting.Extensions`.

However, since I’m integrating a Postgres database with Keycloak, it might be clearer if the name also included the word Postgres.

aaronpowell:
> > Since this is a hosting integration, can you correct the library name to reflect that - it should be `CommunityToolkit.Aspire.Keycloak.Hosting.Extensions`.
> 
> However, since I’m integrating a Postgres database with Keycloak, it might be clearer if the name also included the word Postgres.

I think it's best we keep it as generic extensions, so if there are other things people want to add for extending Keycloak in the future, we don't end up with a heap of micro packages.

axies20:
> > > Since this is a hosting integration, can you correct the library name to reflect that - it should be `CommunityToolkit.Aspire.Keycloak.Hosting.Extensions`.
> > 
> > However, since I’m integrating a Postgres database with Keycloak, it might be clearer if the name also included the word Postgres.
> 
> I think it's best we keep it as generic extensions, so if there are other things people want to add for extending Keycloak in the future, we don't end up with a heap of micro packages.

Okay

axies20:
@davidfowl Will it be implemented?

axies20:
@aaronpowell
Is everything alright?

aaronpowell:
The integration is still incorrectly named and the namespace doesn't match [the contributing guide](https://github.com/CommunityToolkit/Aspire/blob/main/docs/create-integration.md)

src/CommunityToolkit.Aspire.Keycloak.Hosting.Extensions/CommunityToolkit.Aspire.Keycloak.Hosting.Extensions.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <AssemblyName>CommunityToolkit.Aspire.Keycloak.Hosting.Extensions</AssemblyName>
+    <RootNamespace>CommunityToolkit.Aspire.Keycloak.Hosting.Extensions</RootNamespace>
```
aaronpowell: Not required, this is inferred from the project name

src/CommunityToolkit.Aspire.Keycloak.Hosting.Extensions/KeycloakPostgresExtension.cs
aaronpowell, line 1: This needs to be pluralised so it is consistent with all other projects.

src/CommunityToolkit.Aspire.Keycloak.Hosting.Extensions/KeycloakPostgresExtension.cs
```diff
@@ -0,0 +1,80 @@
...
+namespace Aspire.Hosting;
+
+/// <summary>
+/// Provides extension methods for integrating Keycloak resources with PostgreSQL.
+/// </summary>
+public static class KeycloakPostgresExtension
+{
+    private static IResourceBuilder<KeycloakResource> WithPostgresData(
+        this IResourceBuilder<KeycloakResource> builder,
+        IResourceBuilder<PostgresDatabaseResource> database, bool xaEnabled = false)
```
aaronpowell: If parameters are going to be wrapped, can you wrap all for readability.

src/CommunityToolkit.Aspire.Keycloak.Hosting.Extensions/KeycloakPostgresExtension.cs
```diff
@@ -0,0 +1,80 @@
...
+    /// </summary>
+    /// <param name="builder">The builder for the Keycloak resource.</param>
+    /// <param name="database">The builder for the PostgreSQL database resource to use for Keycloak.</param>
+    /// <param name="username">The builder for the parameter resource representing the username for PostgreSQL authentication.</param>
+    /// <param name="password">The builder for the parameter resource representing the password for PostgreSQL authentication.</param>
+    /// <param name="xaEnabled">Indicates whether XA transactions are enabled for PostgreSQL. Defaults to false.</param>
+    /// <returns>The updated Keycloak resource builder configured with PostgreSQL integration and explicit credentials.</returns>
+    public static IResourceBuilder<KeycloakResource> WithPostgres(this IResourceBuilder<KeycloakResource> builder,
+        IResourceBuilder<PostgresDatabaseResource> database, IResourceBuilder<ParameterResource> username,
+        IResourceBuilder<ParameterResource> password, bool xaEnabled = false)
```
aaronpowell: If parameters are going to be wrapped, can you wrap all for readability.

src/CommunityToolkit.Aspire.Keycloak.Hosting.Extensions/KeycloakPostgresExtension.cs
```diff
@@ -0,0 +1,80 @@
...
+    /// <summary>
+    /// Configures a Keycloak resource to use PostgreSQL, explicitly setting credentials and other related values.
+    /// </summary>
+    /// <param name="builder">The builder for the Keycloak resource.</param>
+    /// <param name="database">The builder for the PostgreSQL database resource to use for Keycloak.</param>
+    /// <param name="username">The builder for the parameter resource representing the username for PostgreSQL authentication.</param>
+    /// <param name="password">The builder for the parameter resource representing the password for PostgreSQL authentication.</param>
+    /// <param name="xaEnabled">Indicates whether XA transactions are enabled for PostgreSQL. Defaults to false.</param>
+    /// <returns>The updated Keycloak resource builder configured with PostgreSQL integration and explicit credentials.</returns>
+    public static IResourceBuilder<KeycloakResource> WithPostgres(this IResourceBuilder<KeycloakResource> builder,
```
aaronpowell: If the username and password are made nullable, then the two methods can be combined, with the values using the order of precedence of `username` -> `database.Username` -> created in method.

Directory.Packages.props
```diff
@@ -1,6 +1,8 @@
 <Project>
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    <AspirePreviewVersion>9.4.1-preview.1.25408.4</AspirePreviewVersion>
```
aaronpowell: This belongs in the `Directory.Build.props` file, and it should be just the suffix, so that when used we would do `<PackageVersion Include="Aspire.Hosting.Keycloak" Version="$(AspireVersion)-$(AspirePreviewSuffix)" />`, which ensures that we sync the `Major.Minor.Patch` across all packages.

Directory.Packages.props
```diff
@@ -1,6 +1,8 @@
...
+    <MoqVersion>4.20.72</MoqVersion>
```
aaronpowell: This belongs in `Directory.Build.props`

Directory.Packages.props
```diff
@@ -18,6 +20,9 @@
     <PackageVersion Include="Aspire.Hosting.MongoDB" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.MySql" Version="$(AspireVersion)" />
     <PackageVersion Include="Aspire.Hosting.SqlServer" Version="$(AspireVersion)" />
+    <PackageVersion Include="Aspire.Hosting.Keycloak" Version="$(AspirePreviewVersion)" />
+    <PackageVersion Include="Aspire.Keycloak.Authentication" Version="$(AspirePreviewVersion)" />
+    <PackageVersion Include="Moq" Version="$(MoqVersion)" />
```
aaronpowell: This belongs in the `<ItemGroup Label="Testing">` section.

aaronpowell (changes requested):
Can you please follow the naming conventions [from the contributing guide](https://github.com/CommunityToolkit/Aspire/blob/main/docs/create-integration.md#-getting-started). Hosting integrations should be `CommunityToolkit.Aspire.Hosting.<integration name parts>`.

axies20:
> Since this is a hosting integration, can you correct the library name to reflect that - it should be `CommunityToolkit.Aspire.Keycloak.Hosting.Extensions`.

I followed your recommendation, okay, I will renamed

Issue #800, opened 2025-08-11
Add OpenFeature OFREP for feature flags support


askpt:
### Overview

OpenFeature provides standardisation across feature flag providers, allowing developers to switch between different systems without rewriting application code. As a CNCF project it would be a great addition to the Aspire toolset. More information here: https://openfeature.dev

I am planning on bringing support of OpenFeature using the OFREP protocol (https://www.nuget.org/packages/OpenFeature.Providers.Ofrep) and a container integration (flagd).
- OFREP: https://github.com/open-feature/protocol
- flagd: https://flagd.dev

The idea behind is dotnet aspire would be able to use locally the container (flagd) using OFREP as a communication method and in production, that endpoint would be overwritten by the production endpoint. Similar on how the OTEL and Aspire dashboard currently works.

### Benefits
This integration would provide the .NET Aspire community with:
- Standardised Feature Management: Consistent API across different feature flag providers (flagd, goff or others that use the OFREP endpoint)
- Enhanced Development Workflows: Local development support with containerised services
- Future Flexibility: Easy migration between OFREP compatible providers

### Components
- `CommunityToolkit.Aspire.Hosting.OpenFeature.OFREP` <- contains the OpenFeature.OFREP specific implementation 
- `CommunityToolkit.Aspire.FlagD` <- contains the container for flagd (please note, we should not restrain from using other providers that also support OFREP like GoFeatureFlag)

### Usage example

AppHost

```csharp
var builder = DistributedApplication.CreateBuilder(args);

var flagd = builder.AddFlagd("feature-flags")
    .WithFlagConfiguration("./flags.json")
    .WithOfrepPort(8016);

var api = builder.AddProject<Projects.MyApi>("api")
    .WithReference(ofrep);
```

ServiceDefaults

```csharp
        builder.Services.AddOpenFeature(ofBuilder =>
        {
            ofBuilder
                .AddHostedFeatureLifecycle()
                .AddOfrepProvider() // Service url should be retrieved from the configuration
                .AddHook<TraceEnricherHook>()
                .AddHook<MetricsHook>();
        });
```

Implementation

```csharp
var service = sp.GetService<IFeatureClient>(); // This is injected. Just adding the service here to showcase
var isEnabled = await featureClient.GetBooleanValueAsync( "new-feature", false);
```

### Additional context

[...]

PR #827, opened 2025-09-01
Add flagd Aspire hosting

### LINKED ISSUE #737: Add integration for OpenFeature/Flagd
### .NET Aspire issue link

_No response_

### Overview

[OpenFeature](https://openfeature.dev/) is an open specification for feature flag APIs and providers.

It offers a sub-project called [flagd](https://flagd.dev/) as a default provider.

Integrating this tooling helps developers implement feature flags and control them in their local development environment while reusing feature flag definitions from their live environments.

### Usage example

```c#
// Hosting
var flagd = builder.AddOpenFeature("openfeature")
    .AddFlagdProvider("flagd")
flagd.AddFile("flags.json");
flagd.AddAzureBlob("https://mystorage.blob.core.windows.net/blobs/flags.json")
flagd.WithOverride(/* Parameter */);

// Client
builder.AddOpenFeature();

public class ExampleService(IFeatureClient featureClient)
{
}
```

### Additional context

Given OpenFeature is a client-side SDK, it may make more sense to instead define the interface as simply adding Flagd on the hosting side and instead providing a `builder.AddFlagdProvider()` method in the client which can be used by the existing OpenFeature SDK. However, the presence of Flagd implies the usage of OpenFeature, so it's a bit awkward to ask the user to do some of the setup on their own with OpenFeature.

### Help us help you

askpt:
**Closes #737**

## PR Checklist

- [ ] New integration
  - [ ] Docs are written


src/CommunityToolkit.Aspire.Hosting.Flagd/FlagdBuilderExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+        ArgumentException.ThrowIfNullOrEmpty(fileSource, nameof(fileSource));
+        ArgumentOutOfRangeException.ThrowIfLessThan(port, 1, nameof(port));
+        ArgumentOutOfRangeException.ThrowIfGreaterThan(port, 65535, nameof(port));
+
+        var resource = new FlagdResource(name);
+
+        return builder.AddResource(resource)
+            .WithImage(FlagdContainerImageTags.Image, FlagdContainerImageTags.Tag)
+            .WithImageRegistry(FlagdContainerImageTags.Registry)
+            .WithHttpEndpoint(port: 8013, targetPort: port, name: FlagdResource.HttpEndpointName)
```
Odonno: todo: create a constant `FlagdPort`

src/CommunityToolkit.Aspire.Hosting.Flagd/FlagdBuilderExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// <param name="builder">The <see cref="IDistributedApplicationBuilder"/>.</param>
+    /// <param name="name">The name of the resource. This name will be used as the connection string name when referenced in a dependency.</param>
+    /// <param name="fileSource">The path to the flag configuration file on the host. The flags configuration should be stored in a file named flagd.json</param>
+    /// <param name="port">The host port for flagd HTTP endpoint. If not provided, a random port will be assigned.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{FlagdResource}"/>.</returns>
+    public static IResourceBuilder<FlagdResource> AddFlagd(
+        this IDistributedApplicationBuilder builder,
+        [ResourceName] string name,
+        string fileSource,
+        int port = 8013)
```
Odonno: todo: the other integration uses `int? port = null` so that it provides a dynamic port by default, allowing the possibility to override the port only when needed.

src/CommunityToolkit.Aspire.Hosting.Flagd/FlagdBuilderExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// </summary>
+    /// <param name="builder">The <see cref="IDistributedApplicationBuilder"/>.</param>
+    /// <param name="name">The name of the resource. This name will be used as the connection string name when referenced in a dependency.</param>
+    /// <param name="fileSource">The path to the flag configuration file on the host. The flags configuration should be stored in a file named flagd.json</param>
+    /// <param name="port">The host port for flagd HTTP endpoint. If not provided, a random port will be assigned.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{FlagdResource}"/>.</returns>
+    public static IResourceBuilder<FlagdResource> AddFlagd(
+        this IDistributedApplicationBuilder builder,
+        [ResourceName] string name,
+        string fileSource,
```
Odonno: suggestion: if you expect a default value for this, why not set it here?

```suggestion
        string fileSource = "flagd.json",
```

src/CommunityToolkit.Aspire.Hosting.Flagd/FlagdBuilderExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+            .WithArgs("start", "--uri", "file:./flags/flagd.json");
+    }
+
+    /// <summary>
+    /// Configures logging level for flagd.
+    /// </summary>
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="level">The logging level (debug, info, warn, error).</param>
+    /// <returns>The <see cref="IResourceBuilder{FlagdResource}"/>.</returns>
+    public static IResourceBuilder<FlagdResource> WithLogging(
```
Odonno: suggestion: It may be me but I would have expect `WithLogLevel(builder, FlagdLogLevel level)` with `FlagdLogLevel` being a specifc `enum` of possible values.

Odonno: cc. @aaronpowell

Odonno (changes requested):

examples/flagd/CommunityToolkit.Aspire.Hosting.Flagd.AppHost/Program.cs
```diff
@@ -0,0 +1,20 @@
...
+builder.Eventing.Subscribe<AfterResourcesCreatedEvent>(static async (@event, cancellationToken) =>
+    {
+        // When the resources are created, set the OpenFeature provider to use Flagd (debug purposes)
+        await OpenFeature.Api.Instance.SetProviderAsync(new FlagdProvider());
+
+        var flagClient = OpenFeature.Api.Instance.GetClient();
+        var welcomeBanner = await flagClient.GetBooleanDetailsAsync("welcome-banner", false);
+        var backgroundColor = await flagClient.GetStringDetailsAsync("background-color", "000000");
+        var apiVersion = await flagClient.GetStringDetailsAsync("api-version", "0.1");
+    });
```
aaronpowell: Is this for illustrating something specific in the example apphost, or is this the kind of thing that we could roll up into an extension method?

src/CommunityToolkit.Aspire.Hosting.Flagd/FlagdBuilderExtensions.cs
aaronpowell: yep, should be defaulting to `null` and leave that to Aspire to assign a port.

src/CommunityToolkit.Aspire.Hosting.Flagd/FlagdBuilderExtensions.cs
aaronpowell: Agreed, an enum would be better as that is more discoverable to the end user.

Alternatively, shouldn't it just inherit the logging level from the standard logger? Is there a reason you'd configure flagd to log a different verbosity level to the whole app?

tests/CommunityToolkit.Aspire.Hosting.Flagd.Tests/AppHostTests.cs
```diff
@@ -0,0 +1,16 @@
...
+[RequiresDocker]
+public class AppHostTests(AspireIntegrationTestFixture<Projects.CommunityToolkit_Aspire_Hosting_Flagd_AppHost> fixture) : IClassFixture<AspireIntegrationTestFixture<Projects.CommunityToolkit_Aspire_Hosting_Flagd_AppHost>>
+{
+    [Fact]
+    public async Task ResourceStartsAndRespondsCorrectly()
+    {
+        var resourceName = "flagd";
+        await fixture.ResourceNotificationService.WaitForResourceAsync(resourceName).WaitAsync(TimeSpan.FromMinutes(1));
+        var httpClient = fixture.CreateHttpClient(resourceName);
+    }
```
aaronpowell: Is there anything we can assert against on the resource to verify that it's running beyond "it didn't time out"?

aaronpowell (changes requested):

src/CommunityToolkit.Aspire.Hosting.Flagd/FlagdBuilderExtensions.cs
Odonno: I am not sure what the default logger would be. If it's like goff, the default log level is INFO. You can still override it within the config file and if needed by the environment variable. I do think that overriding with an env variable like this is the best solution. What did you have in mind?

src/CommunityToolkit.Aspire.Hosting.Flagd/FlagdBuilderExtensions.cs
askpt: Fixed with https://github.com/CommunityToolkit/Aspire/pull/827/commits/6317cd6f962119343752291adfaa20684fecfe53

src/CommunityToolkit.Aspire.Hosting.Flagd/FlagdBuilderExtensions.cs
askpt: Fixed with https://github.com/CommunityToolkit/Aspire/pull/827/commits/f468d420f4e472c7ddb62387ff360f19b12cd91c

src/CommunityToolkit.Aspire.Hosting.Flagd/FlagdBuilderExtensions.cs
askpt: Fixed with https://github.com/CommunityToolkit/Aspire/pull/827/commits/a03c6e0171287413e697d109eb596323e77585a7

This was still code from the initial scaffolding the copilot did when I used the agent in this repo to create the structure for this hosting project. Now it is fixed to add extra logging when a flag is evaluated: https://flagd.dev/troubleshooting/

examples/flagd/CommunityToolkit.Aspire.Hosting.Flagd.AppHost/Program.cs
askpt: @aaronpowell I am using this as a testing project to see if the integration is working properly. With this I can test a flag evaluation when I start this aspire project.

Ideally, I will add later this Service extension: https://github.com/CommunityToolkit/Aspire/issues/800, which will deal with the DI and so on. Similar to what I have here: https://github.com/askpt/openfeature-aspire-welcome/blob/550eb169399da489916ebf9360503df99cf9befe/Todo.ServiceDefaults/Extensions.cs#L32

tests/CommunityToolkit.Aspire.Hosting.Flagd.Tests/AppHostTests.cs
askpt: I am still implementing this tests and cleanup some code. I am planning on adding a flag evaluation here, similar to what is done in the `examples` project.

src/CommunityToolkit.Aspire.Hosting.Flagd/FlagdBuilderExtensions.cs
askpt: I am still thinking on what to do here. The main reason is flagd supports multiple sources and at this moment I am limiting just to file system. Thinking on if I should add a minimal working hosting that works with just files or adding now all the possible syncs. Check: https://flagd.dev/concepts/syncs/

askpt:
@aaronpowell could you please reopen this PR? I am still working on it

src/CommunityToolkit.Aspire.Hosting.Flagd/README.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+### Usage
+
+In your `AppHost` project, call the `AddFlagd` method to add flagd to your application with a flag configuration file:
+
+```csharp
+var builder = DistributedApplication.CreateBuilder(args);
+
+var flagd = builder.AddFlagd("flagd", "./flags");
```
kylejuliandev: I think the usage example needs updating after the changes in [fc2b7a8](https://github.com/CommunityToolkit/Aspire/pull/827/commits/fc2b7a8de152df6715428656fe160aee2992c225) 👍

askpt: Good catch! https://github.com/CommunityToolkit/Aspire/pull/827/commits/c899e0cf26b4cbbdaaf3b77cc90fec178b5f9ad7

src/CommunityToolkit.Aspire.Hosting.Flagd/FlagdBuilderExtensions.cs
Odonno: Yeah, I checked and I don't see any possibility to set the log level (like `warning` or `info`). It's either DEBUG or not.

I would go for a more generic solution/naming that can be reused for other Hosting integration like goff, I suggest this:

```cs
enum FlagdLogLevel
{
  Error = 0,
  Debug
}

public static IResourceBuilder<FlagdResource> WithLoglevel(
  this IResourceBuilder<FlagdResource> builder,
  FlagdLogLevel logLevel)
{
  // ..
  return builder.WithEnvironment("FLAGD_DEBUG", logLevel == FlagdLogLevel.Error ? "true" : "false");
}
```

Even better would be reusing this https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel?view=net-9.0-pp

```cs
public static IResourceBuilder<FlagdResource> WithLoglevel(
  this IResourceBuilder<FlagdResource> builder,
  LogLevel logLevel)
{
  // ..
  if (logLevel == LogLevel.Debug)
  {
    return builder.WithEnvironment("FLAGD_DEBUG", "true");
  }
  else if (logLevel == LogLevel.Error)
  {
    return builder;
  }
  else
  {
    // throw error as not expected? 
  }
}
```

What do you think @aaronpowell ?

askpt: It definitely gives a more .NET-like look and feel. Waiting on feedback to make that change since it feels a design we might want to replicate in other integrations.

aaronpowell: Yeah let's go idiomatic with the .NET logging levels. I think that raising an exception if someone uses a log level that's unsupported would be the right way to go as that avoids a false sense of what logging can achieve, and allows for flexibility if flagd adds additional logging levels in the future.

aaronpowell:
@askpt can you add the test project to the list so that the tests are run.

askpt:
> @askpt can you add the test project to the list so that the tests are run.

@aaronpowell I fixed the test list and the log level suggestion. Should be now good to go 👍

aaronpowell (approved):

PR #831, opened 2025-09-03
Adds API RunAsExecutable for DAB integration

### LINKED ISSUE #639: Data API Builder RunAsExecutable

Yes

### Existing integration

DataAPIBuilder

### Overview

Running DAB as an executable could be a valuable option for developers to avoid dealing with a container. There is a [known issue](https://github.com/dotnet/aspire/issues/6635) at the moment in Aspire when sending telemetry from a container.

### Usage example

```csharp
var dab = builder.AddDataAPIBuilder("dab").RunAsExecutable();
```

Where RunAsExecutable would check for the `Context.IsPublishMode` and if the app is running locally, it would run the same dab config using the local CLI.

### Breaking change?

No

### Alternatives

Current alternative is to run DAB as an executable using the standard Aspire `AddExectuable` method.
```csharp
var dab = builder.AddExecutable("dab", dabCLIPath, "../../dab/", "start")
    .WithReference(sqlDatabase)
    .WithHttpEndpoint(targetPort: 5000)
    .WaitFor(sqlServer)
    .WithOtlpExporter();
```

### Additional context

_No response_

### Help us help you

tommasodotNET:
**Closes #639**

DAB can now be run as executable locally. This allows support for https and OTEL in the local dev experience:
```csharp
var dab = builder.AddDataAPIBuilder("dab")
    .RunAsExecutable()
    .WaitFor(sqlServer)
    .WithReference(sqlDatabase);
```

When the application is in Publish mode, the `RunAsExecutable` extension will be ignored and the resource will be deployed using the DAB docker image.

## PR Checklist

- [ ] New integration
  - [ ] Docs are written


aaronpowell (commented):
Before doing a detailed review, I want to ask a question about the design.

My assumption is that the intent of `RunAsExecutable` would be to treat the exe version as a local dev only solution, essentially an emulator right?

Working on that assumption, I think it'd be best to use the pattern we see with `RunAsEmulator` on things like the Azure Storage resource, or `RunAsContainer` for databases.

The primary difference in the design would be that you don't return the "emulated" resource, but instead you provide it as a callback to the `RunAsEmulator` method (or in this case `RunAsExecutable`). That way, if there's stuff you want to do that is specific to the "emulated" resource, you do it in the callback, and everything else is bubbled up from the real resource.

ErikEJ:
@tommasodotNET Does this also address #388 ?

aaronpowell:
@ErikEJ it would _in a sense_ as you don't have to inject the certificates into the container so you can get OTEL since it's running on the host that DCP runs on.

But with #602 there is a new method that will allow injecting certificates into containers, which would unblock it in a different manner.

tommasodotNET:
> Before doing a detailed review, I want to ask a question about the design.
> 
> My assumption is that the intent of `RunAsExecutable` would be to treat the exe version as a local dev only solution, essentially an emulator right?
> 
> Working on that assumption, I think it'd be best to use the pattern we see with `RunAsEmulator` on things like the Azure Storage resource, or `RunAsContainer` for databases.
> 
> The primary difference in the design would be that you don't return the "emulated" resource, but instead you provide it as a callback to the `RunAsEmulator` method (or in this case `RunAsExecutable`). That way, if there's stuff you want to do that is specific to the "emulated" resource, you do it in the callback, and everything else is bubbled up from the real resource.

I reversed the logic of `PublishAsDockerfile`. But that does not yield a great result - we have to declare `RunAsExecutable()` before calling `WithReference` and `WaitFor`. I'll try changing the design logic to match the `RunAsEmulator`. Great call, @aaronpowell, thanks

tommasodotNET:
Let's wait for #602 and see if that solves the issue for both otel and https support in dab locally. If it does, this extension might not be needed.

tommasodotNET:
in any case, we should update the dab image tag to at least [1.4.26](https://github.com/Azure/data-api-builder/releases/tag/v1.4.26) since it introduces support for otel.

PR #853, opened 2025-09-14
Fix Otel Collector on Windows

### LINKED ISSUE #852: Open Telemetry Collector doesn't work on Windows
### Describe the bug

Using the Otel Collector on windows results in the collector entering the "FailedToStart" state.  This happens because the dev cert certificate paths end up with invalid characters due to `Path.Combine()` using windows directory separators rather than unix ones.

### Regression

_No response_

### Steps to reproduce

```text
Run the open telemetry sample project on windows.


dotnet run --project .\examples\opentelemetry-collector\CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector.AppHost\
```

### Expected behavior

Open Telemetry Collector should start properly.

### Screenshots

_No response_

### IDE and version

VS Code, Other, VS 2022

### IDE version

.NET SDK: 9.0.304 f12f5f689e

### Nuget packages

```text
latest main (b98f4c2eb41ae7b0a217465e9429820aedca3248)
```

### Additional context

https://discord.com/channels/1361488941836140614/1375617984512528567/1414908067518222369

### Help us help you

afscrome:
**Closes #852**
Fixes: https://discord.com/channels/1361488941836140614/1375617984512528567/1414908067518222369
Fixes: https://discord.com/channels/1361488941836140614/1375617984512528567/1415839400901738616
Partly Fixes: https://discord.com/channels/1361488941836140614/1375617984512528567/1414909111987998753, although the fix may not fully work until dotnet/aspire#6547 is resolved

- Avoid using `Path.Combine` when calculating container paths for certs - `Path.Combine` uses the host OS's path separator, but these values MUST use unix style path separators. 
- Use `HostUrl` to resolve the Aspire Dashboard url in the context of the otel container (i.e. handling any `host.docker.internal` translation)
- Refactored `WithAppForwarding` to use the global `BeforeStartEvent` rather than resource specific events - this avoids race conditions of other resources starting up whilst the otel collector starts up.

## PR Checklist

### DIFF
--- a/tests/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector.Tests/ResourceCreationTests.cs
+++ b/tests/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector.Tests/ResourceCreationTests.cs
@@ -1,16 +1,16 @@
+using Aspire.Components.Common.Tests;
 using Aspire.Hosting;
 using Aspire.Hosting.Utils;
-using Microsoft.Extensions.FileProviders;
-using Microsoft.Extensions.Hosting;
+using Xunit.Abstractions;
 
 namespace CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector.Tests;
 
-public class ResourceCreationTests
+public class ResourceCreationTests(ITestOutputHelper testOutputHelper)
 {
     [Fact]
     public void CanCreateTheCollectorResource()
     {
-        var builder = DistributedApplication.CreateBuilder();
+        var builder = TestDistributedApplicationBuilder.Create();
 
         builder.AddOpenTelemetryCollector("collector")
             .WithConfig("./config.yaml")
@@ -150,33 +150,33 @@ public class ResourceCreationTests
     }
 
     [Fact]
-    public void ContainerHasAspireEnvironmentVariables()
+    [RequiresDocker]
+    public async Task ContainerHasAspireEnvironmentVariables()
     {
-        var builder = DistributedApplication.CreateBuilder();
+        using var builder = TestDistributedApplicationBuilder.Create()
+            .WithTestAndResourceLogging(testOutputHelper);
+        builder.Configuration["APPHOST:ContainerHostname"] = "what.ever";
 
-        builder.AddOpenTelemetryCollector("collector")
+        var collector = builder.AddOpenTelemetryCollector("collector")
             .WithAppForwarding();
 
         using var app = builder.Build();
         var appModel = app.Services.GetRequiredService<DistributedApplicationModel>();
-        var collectorResource = appModel.Resources.OfType<OpenTelemetryCollectorResource>().SingleOrDefault();
-        Assert.NotNull(collectorResource);
 
-        var envs = collectorResource.Annotations.OfType<EnvironmentCallbackAnnotation>().ToList();
-        Assert.NotEmpty(envs);
+        var resourceNotificationService = app.Services.GetRequiredService<ResourceNotificationService>();
 
-        var context = new EnvironmentCallbackContext(new DistributedApplicationExecutionContext(new DistributedApplicationExecutionContextOptions(DistributedApplicationOperation.Run)));
-        foreach (var env in envs)
-        {
-            env.Callback(context);
-        }
+        await app.StartAsync();
+        await resourceNotificationService.WaitForResourceHealthyAsync(collector.Resource.Name);
 
-        Assert.Contains("ASPIRE_ENDPOINT", context.EnvironmentVariables.Keys);
-        Assert.Contains("ASPIRE_API_KEY", context.EnvironmentVariables.Keys);
+        Assert.True(resourceNotificationService.TryGetCurrentState(collector.Resource.Name, out var resourceEvent));
 
-        var url = Assert.IsType<HostUrl>(context.EnvironmentVariables["ASPIRE_ENDPOINT"]);
-        Assert.Equal("http://localhost:18889", url.Url);
-        Assert.NotNull(context.EnvironmentVariables["ASPIRE_API_KEY"]);
+        var envVars = resourceEvent.Snapshot.EnvironmentVariables.ToDictionary(k => k.Name, v => v.Value);
+
+        var endpoint = Assert.Contains("ASPIRE_ENDPOINT", envVars);
+        var apiKey = Assert.Contains("ASPIRE_API_KEY", envVars);
+
+        Assert.Equal($"http://what.ever:18889", endpoint);
+        Assert.NotNull(apiKey);
     }
 
     [Fact]


src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/OpenTelemetryCollectorExtensions.cs
```diff
@@ -79,8 +79,21 @@ public static IResourceBuilder<OpenTelemetryCollectorResource> AddOpenTelemetryC
...
+        {
+            var logger = evt.Services.GetRequiredService<ResourceLoggerService>().GetLogger(builder.Resource);
+            var otelSenders = evt.Model.Resources
+                .OfType<IResourceWithEnvironment>()
+                .Where(x => x.HasAnnotationOfType<OtlpExporterAnnotation>());
+
+            foreach (var otelSender in otelSenders)
+            {
+                var otelSenderBuilder = builder.ApplicationBuilder.CreateResourceBuilder(otelSender);
+                otelSenderBuilder.WithOpenTelemetryCollectorRouting(builder);
```
martinjt, line 92: Is using the builder from outside the event the right option here? or should we be thinking about looking up the collector? I'm not sure if they actually change when the builder is built or not.

tests/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector.Tests/ResourceCreationTests.cs
```diff
@@ -173,7 +173,9 @@ public void ContainerHasAspireEnvironmentVariables()
 
         Assert.Contains("ASPIRE_ENDPOINT", context.EnvironmentVariables.Keys);
         Assert.Contains("ASPIRE_API_KEY", context.EnvironmentVariables.Keys);
-        Assert.Equal("http://host.docker.internal:18889", context.EnvironmentVariables["ASPIRE_ENDPOINT"]);
+
+        var url = Assert.IsType<HostUrl>(context.EnvironmentVariables["ASPIRE_ENDPOINT"]);
```
martinjt: We could add a test here that uses `AppHost:ContainerHostname"` to ensure that it's using the container hostname and it's not hardcoded.

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/OpenTelemetryCollectorExtensions.cs
afscrome: This should be ok - it's actually somewhat recommended see https://github.com/dotnet/aspire/issues/9146 .

> This event is the last chance to add or remove resources from the DistributedApplicationModel before they're processed by consumers (note you'll likely want to capture the IDistributedApplicationBuilder before this so that you can get instances of IResourceBuilder<TResource> if adding resources)

tests/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector.Tests/ResourceCreationTests.cs
afscrome: I'm going to punt on this to a follow up PR - `HostUrl` resolution happens deep within aspire start up - https://github.com/dotnet/aspire/blob/36509b5734d6e1984c2dc346b559edbbef2d398d/src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs#L73

We can test this by using `DistributedApplicationTestingBuilder` to build & run the app, but that makes it more of an integration test than a unit test (Taking ~10 seconds).  It's doable, but nothing in this project has done that yet so there are a whole load of demons that can crop up when getting it running in Github Actions for the first time, so I'd rather get it into it's own PR.  

```cs
    public async Task ContainerHasAspireEnvironmentVariables()
    {
        var builder = DistributedApplicationTestingBuilder.Create();
        builder.Configuration["APPHOST:ContainerHostname"] = "what.ever";

        var collector = builder.AddOpenTelemetryCollector("collector")
            .WithAppForwarding();

        using var app = builder.Build();
        var appModel = app.Services.GetRequiredService<DistributedApplicationModel>();

        var resourceNotificationService = app.Services.GetRequiredService<ResourceNotificationService>();

        await app.StartAsync();
        await resourceNotificationService.WaitForResourceHealthyAsync(collector.Resource.Name);

        Assert.True(resourceNotificationService.TryGetCurrentState(collector.Resource.Name, out var resourceEvent));

        var envVars = resourceEvent.Snapshot.EnvironmentVariables.ToDictionary(k => k.Name, v => v.Value);


        var endpoint = Assert.Contains("ASPIRE_ENDPOINT", envVars);
        var apiKey = Assert.Contains("ASPIRE_API_KEY", envVars);

        Assert.Equal($"http://what.ever:{collector.Resource.GrpcEndpoint.TargetPort}", endpoint);
        Assert.NotNull(apiKey);
    }
```

martinjt: The DevCerts tests do this FWIW.

afscrome: Ah, I couldn't find any references to `DistributedApplicationTestingBuilder`, so didn't realise there was a custom test framework.  Pushed the test update.

tests/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector.Tests/ResourceCreationTests.cs
```diff
@@ -150,31 +150,33 @@ public void CanDisableBothEndpoints()
     }
 
     [Fact]
-    public void ContainerHasAspireEnvironmentVariables()
+    [RequiresDocker]
```
martinjt, line 153: Do we actually need to run this with docker to ensure that it's using the HostUrl conversion? My perception was that the resolving doesn't actually need all that. I'm happy to be wrong though

afscrome: `HostUrl` processing depends on context sensitive information which is only passed through when DCP properly starts the process up.  We may be able to do some things to not need to wait for the entire start process to complete, but then we're starting to relying on implementation details that'll be racy based on whether aspire does anything with the container runtime before evaluating env vars.  e.g. right now, DCP evaluates env vars before it starts pulling docker images, but that could change.

`await collector.Resource.GetEnvironmentVariableValuesAsync()` allows us to get env vars without going through DCP's start up, but it doesn't plumb `AppHost:ContainerHost` through the whole way, so falls back to `host.docker.internal ` - https://github.com/dotnet/aspire/blob/521ff0440f1433f1819c854c74eca7d2ef13fe75/src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs#L436 .

Alternatively we go back to the original test and just test as far as `The annotation has a value of type `HostUrl`, with our untranslated value`, and just trust that `HostUrl` gets translated correctly.

afscrome: @martinjt are you happy with this?  I'd like to get this merged into the toolkit release alongside 9.5

martinjt: Yup, it was only questions, this should already be approved.

afscrome: ah, I misunderstood.

afscrome:
@aaronpowell Can we merge this?

PR #866, opened 2025-09-20
Update hard coded version to "latest"

JerryNixon:
That old version is way-way outdated.


ErikEJ:
@JerryNixon How about the sample config files and the schema version inside them?

tommasodotNET:
yeah the need to update dab to latest has been mentioned here #831

tommasodotNET (approved):

PR #870, opened 2025-09-23
[9.5]: Add icons to the otel collector & dev cert export

afscrome:
Make use of the `WithIconName` API added in 9.5:

After:
image

Before:
image


aaronpowell:
PRETTY!

PR #914, opened 2025-10-22
Aspire 13 upgrade

aaronpowell:
Shortly the next major version of Aspire will be released and thing branch is going to act as the convergence point for all the work that needs to be done to update packages to work with Aspire 13, and any new things we want to add to support Aspire 13.


aaronpowell:
@copilot can you fix the failing test

tommasodotNET:
@aaronpowell should we mark all python extensions as deprecated?

aaronpowell:
> @aaronpowell should we mark all python extensions as deprecated?

I've got PR #945 open (that targets this branch) that deprecates the methods in it as I think everything in the Python extensions is covered by Aspire 13. The question is - are there new scenarios that would make sense to add to the CT Python extensions? Fully deprecating the NuGet package could be a pain if we in the future wanted to bring it back.

PR #956, opened 2025-11-08
Add Umami hosting integration

Odonno:
Add the Umami (https://umami.is/) hosting integration, a web analytics solution that can be self-hosted simply with a SQL backend storage.

Note: only PostgreSQL is currently supported as a backend storage.

## PR Checklist

- [x] New integration
  - [x] Docs are written

## Other information

* I have added a new `WithStorageBackend` extension method, not sure if that follows the current convention
* Tested locally with the example project, via the Blazor web app example
* I tweaked the code to make cross-container communication working with Podman. I am not sure what is the best practice at the moment (reading the connection string & supporting both Docker & Podman)
* I had some compilation errors with the newly created Blazor app. I am not sure why this is happening. I might go for a simple Vite app.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -25,6 +25,7 @@ jobs:
             Hosting.Azure.Dapr.Redis.Tests,
             Hosting.Azure.Dapr.Tests,
             Hosting.Azure.DataApiBuilder.Tests,
+            Hosting.Azure.Extensions.Tests,
             Hosting.Bun.Tests,
             Hosting.Dapr.Tests,
             Hosting.DbGate.Tests,
--- a/CommunityToolkit.Aspire.slnx
+++ b/CommunityToolkit.Aspire.slnx
@@ -8,6 +8,9 @@
   <Folder Name="/examples/adminer/">
     <Project Path="examples/adminer/CommunityToolkit.Aspire.Hosting.Adminer.AppHost/CommunityToolkit.Aspire.Hosting.Adminer.AppHost.csproj" />
   </Folder>
+  <Folder Name="/examples/azure-ext/">
+    <Project Path="examples/azure-ext/CommunityToolkit.Aspire.Azure.Extensions.AppHost/CommunityToolkit.Aspire.Azure.Extensions.AppHost.csproj" />
+  </Folder>
   <Folder Name="/examples/bun/">
     <Project Path="examples/bun/CommunityToolkit.Aspire.Hosting.Bun.AppHost/CommunityToolkit.Aspire.Hosting.Bun.AppHost.csproj" />
   </Folder>
@@ -187,6 +190,7 @@
     <Project Path="src/CommunityToolkit.Aspire.Hosting.ActiveMQ/CommunityToolkit.Aspire.Hosting.ActiveMQ.csproj" />
     <Project Path="src/CommunityToolkit.Aspire.Hosting.Adminer/CommunityToolkit.Aspire.Hosting.Adminer.csproj" />
     <Project Path="src/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder.csproj" />
+    <Project Path="src/CommunityToolkit.Aspire.Hosting.Azure.Extensions/CommunityToolkit.Aspire.Hosting.Azure.Extensions.csproj" />
     <Project Path="src/CommunityToolkit.Aspire.Hosting.Bun/CommunityToolkit.Aspire.Hosting.Bun.csproj" />
     <Project Path="src/CommunityToolkit.Aspire.Hosting.DbGate/CommunityToolkit.Aspire.Hosting.DbGate.csproj" />
     <Project Path="src/CommunityToolkit.Aspire.Hosting.Deno/CommunityToolkit.Aspire.Hosting.Deno.csproj" />
@@ -246,6 +250,7 @@
     <Project Path="tests/CommunityToolkit.Aspire.Hosting.ActiveMQ.Tests/CommunityToolkit.Aspire.Hosting.ActiveMQ.Tests.csproj" />
     <Project Path="tests/CommunityToolkit.Aspire.Hosting.Adminer.Tests/CommunityToolkit.Aspire.Hosting.Adminer.Tests.csproj" />
     <Project Path="tests/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder.Tests/CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder.Tests.csproj" />
+    <Project Path="tests/CommunityToolkit.Aspire.Hosting.Azure.Extensions.Tests/CommunityToolkit.Aspire.Hosting.Azure.Extensions.Tests.csproj" />
     <Project Path="tests/CommunityToolkit.Aspire.Hosting.Bun.Tests/CommunityToolkit.Aspire.Hosting.Bun.Tests.csproj" />
     <Project Path="tests/CommunityToolkit.Aspire.Hosting.DbGate.Tests/CommunityToolkit.Aspire.Hosting.DbGate.Tests.csproj" />
     <Project Path="tests/CommunityToolkit.Aspire.Hosting.Deno.Tests/CommunityToolkit.Aspire.Hosting.Deno.Tests.csproj" />
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -40,7 +40,7 @@
   <!-- Versioning -->
   <PropertyGroup>
     <ToolkitMinorVersion>1</ToolkitMinorVersion>
-    <ToolkitPatchVersion>1</ToolkitPatchVersion>
+    <ToolkitPatchVersion>2</ToolkitPatchVersion>
     <ToolkitPreviewVersion>preview.1</ToolkitPreviewVersion>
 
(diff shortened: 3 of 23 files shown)


examples/umami/CommunityToolkit.Aspire.Hosting.Umami.AppHost/AppHost.cs
```diff
@@ -0,0 +1,17 @@
...
+    .AddPostgres("postgres", password: password, port: 61118)
+    .WithLifetime(ContainerLifetime.Persistent);
+var postgresdb = postgres.AddDatabase("postgresdb");
+
+var umami = builder
+    .AddUmami("umami", port: 55932)
+    .WithStorageBackend(postgresdb);
+
+var front = builder
+    .AddJavaScriptApp("front", "../umami-vite-app");
```
aaronpowell: I feel like there needs to be some kind of relationship here so that the frontend app is able to get the access key and endpoint for the backend rather than having them hard coded in the HTML.

It also would be nice to have a way to provide the website endpoint to the unami resource, but reading their docs that doesn't look possible.

src/CommunityToolkit.Aspire.Hosting.Umami/UmamiBuilderExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+
+    /// <summary>
+    /// References a <see cref="PostgresDatabaseResource"/> as the storage backend for the <see cref="UmamiResource"/>.
+    /// </summary>
+    /// <param name="builder">The Umami resource builder.</param>
+    /// <param name="database">The PostgreSQL database resource builder.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<UmamiResource> WithStorageBackend(
+        this IResourceBuilder<UmamiResource> builder, 
+        IResourceBuilder<PostgresDatabaseResource> database
```
aaronpowell: It looks like they've just dropped support for MySQL and now only support PostgreSQL, so maybe the method name could be more like `WithPostgreSQL` or something rather than trying to be generic.

src/CommunityToolkit.Aspire.Hosting.Umami/UmamiBuilderExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+        builder.WaitFor(database);
+        builder.WithEnvironment(async (context) =>
+        {
+            var connectionString = await database.Resource.UriExpression.GetValueAsync(CancellationToken.None).ConfigureAwait(false);
+            if (connectionString is null)
+            {
+                throw new DistributedApplicationException($"Failed to retrieve the connection string of the '{database.Resource.Name}' resource.");
+            }
+            
+            // TODO : How to use container host?
```
aaronpowell: What do you mean?

aaronpowell (changes requested):
Just a few questions about the design, and it looks like the integration test isn't passing.

src/CommunityToolkit.Aspire.Hosting.Umami/UmamiBuilderExtensions.cs
Odonno: Cross-container communication does not seem to work with the given connection string I get. I need to rewrite it with the network IP address & port for it to work. I don't know what is the current practice for this to work. That's also why the test is failing.

examples/umami/CommunityToolkit.Aspire.Hosting.Umami.AppHost/AppHost.cs
Odonno: I first need to create a tracking website from Umami. So that would be the other way around. I will try to add a `WaitFor` and a custom logic for the JS app to register to umami dynamically.

src/CommunityToolkit.Aspire.Hosting.Umami/UmamiBuilderExtensions.cs
Odonno: I suppose the idea is to have an agnostic backend.

examples/umami/CommunityToolkit.Aspire.Hosting.Umami.AppHost/AppHost.cs
Odonno: Done. It's a lot of boilerplate code though. I don't know if we should refactor it and expose it to be used by other developers or if we should leave it as is.

src/CommunityToolkit.Aspire.Hosting.Umami/UmamiBuilderExtensions.cs
Odonno: To make it clear, I currently have to resort to this trick to have it work:
```cs
context.EnvironmentVariables[DatabaseStorageEnvVarName] = connectionString
                .Replace("localhost", "10.89.1.2")
                .Replace("61118", "5432");
```

src/CommunityToolkit.Aspire.Hosting.Umami/UmamiBuilderExtensions.cs
aaronpowell: When I read their docs (admittedly a brief read) it seemed like they only support PG as a backend, v3 dropped support for MySQL

src/CommunityToolkit.Aspire.Hosting.Umami/UmamiBuilderExtensions.cs
aaronpowell: You should just be relying on the endpoint reference in the same way we do things like Adminer and DbGate in the Postgres extensions. Since they are in a shared Docker network, the hostname would be `pg` that it can connect via.

Odonno: Done. It works perfectly.

src/CommunityToolkit.Aspire.Hosting.Umami/UmamiBuilderExtensions.cs
Odonno: At this moment, yes. I renamed it.

aaronpowell: If they bring it back, we'll handle the API breaking change then 😝

examples/umami/CommunityToolkit.Aspire.Hosting.Umami.AppHost/CommunityToolkit.Aspire.Hosting.Umami.AppHost.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <Sdk Name="Aspire.AppHost.Sdk" Version="$(AspireAppHostSdkVersion)"/>
```
aaronpowell:
```suggestion
﻿<Project Sdk="Aspire.AppHost.Sdk/13.0.0">
```

examples/umami/CommunityToolkit.Aspire.Hosting.Umami.AppHost/Program.cs
```diff
@@ -0,0 +1,73 @@
...
+
+var postgres = builder
+    .AddPostgres("postgres", password: password, port: 61118)
+    .WithLifetime(ContainerLifetime.Persistent);
+var postgresdb = postgres.AddDatabase("postgresdb");
+
+var umami = builder
+    .AddUmami("umami", port: 55932)
+    .WithPostgreSQL(postgresdb)
+    .OnResourceReady(async (resource, _, ct) =>
```
aaronpowell, line 18: I assume that this is only for something you want in this demo app and not something you'd need to do all the time right?

src/CommunityToolkit.Aspire.Hosting.Umami/README.md
```diff
@@ -0,0 +1,33 @@
+# CommunityToolkit.Aspire.Hosting.Umami library
+
+Provides extension methods and resource definitions for the .NET Aspire AppHost to support running [Umami](https://umami.is/) containers.
```
aaronpowell:
```suggestion
Provides extension methods and resource definitions for the Aspire AppHost to support running [Umami](https://umami.is/) containers.
```

Image

src/CommunityToolkit.Aspire.Hosting.Umami/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Getting started
+
+### Install the package
+
+In your AppHost project, install the .NET Aspire Umami Hosting library with [NuGet](https://www.nuget.org):
```
aaronpowell:
```suggestion
In your AppHost project, install the Aspire Umami Hosting library with [NuGet](https://www.nuget.org):
```

aaronpowell (changes requested):

examples/umami/CommunityToolkit.Aspire.Hosting.Umami.AppHost/Program.cs
Odonno: Yes, this is for demo purpose.

src/CommunityToolkit.Aspire.Hosting.Umami/README.md
Odonno: Just basic c/c without reading

aaronpowell (approved):

PR #960, opened 2025-11-11
xUnit v3 upgrade

### LINKED ISSUE #961: Upgrade xUnit to v3
The tests use xUnit 2, which is not actively developed compared to xUnit 3 (which is also what Aspire uses).

This will require moving to MTP, probably MTP 2.0 (no point starting at 1.0 as it's already been revved), so does result in an overhaul of the way tests are run in our CI pipeline.

We will be blocked due to https://github.com/Tyrrrz/GitHubActionsTestLogger/issues/41 and I don't want to drop that logger from CI, but there's an active PR https://github.com/Tyrrrz/GitHubActionsTestLogger/pull/53

aaronpowell:
Upgrading the tests to use xUnit v3, which is what is used by aspire now too.

Close #961

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -132,21 +132,21 @@ jobs:
 
       - name: Upload bin log artifact
         if: always()
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: binlog-${{ matrix.name }}-${{ matrix.os }}
           path: "**/*.binlog"
 
       - name: Upload test results artifact
         if: always()
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: testresults-${{ matrix.name }}-${{ matrix.os }}
           path: testresults/**
 
       - name: Upload npm logs if failed
         if: failure()
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: npm-logs-${{ matrix.name }}-${{ matrix.os }}
           path: ${{ runner.os == 'Windows' && 'C:\\npm\\cache\\_logs\\**' || '~/.npm/cache/_logs/**' }}
--- a/.github/agents/hosting-integration-creator.agent.md
+++ b/.github/agents/hosting-integration-creator.agent.md
@@ -238,7 +238,7 @@ Each hosting integration should have a corresponding sample application in the `
 Here is an example of the `csproj` file for the AppHost project for the Bun hosting integration:
 
 ```xml
-<Project Sdk="Aspire.AppHost.Sdk/13.2.0-preview.1.26126.8">
+<Project Sdk="Aspire.AppHost.Sdk/13.2.0-preview.1.26152.2">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -10,12 +10,12 @@ jobs:
       - uses: actions/checkout@v6
 
       - name: Download test results
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@v8
         with:
           pattern: testresults-*
 
       - name: Combine coverage reports
-        uses: danielpalme/ReportGenerator-GitHub-Action@5.5.1
+        uses: danielpalme/ReportGenerator-GitHub-Action@5.5.2
         with:
           reports: "**/*.cobertura.xml"
           targetdir: "${{ github.workspace }}/report"
@@ -28,7 +28,7 @@ jobs:
           assemblyfilters: "-*.AppHost;-CommunityToolkit.Aspire.Hosting.Azure.DataApiBuilder.BlazorApp;-*.ServiceDefaults;-CommunityToolkit.Aspire.Testing;-Aspire.Hosting;-Aspire.Hosting.JavaScript;-Aspire.Hosting.SqlServer;-Aspire.Hosting.Python;-Aspire.Hosting.RabbitMQ;-Aspire.Hosting.Redis;-Aspire.Hosting.PostgreSQL;-Aspire.Hosting.MongoDB"
 
       - name: Upload combined coverage XML
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: coverage
           path: ${{ github.workspace }}/report
@@ -48,7 +48,7 @@ jobs:
           thresholds: "60 80"
 
       - name: Upload combined coverage markdown
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: coverage-markdown
           path: ${{ github.workspace }}/code-coverage-results.md
--- a/.github/workflows/dotnet-ci.yml
+++ b/.github/workflows/dotnet-ci.yml
@@ -64,7 +64,7 @@ jobs:
       name: azure-artifacts
     steps:
       - name: Download package
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@v8
         with:
           name: nuget-packages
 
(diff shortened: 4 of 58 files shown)


.github/workflows/tests.yaml
```diff
@@ -92,11 +92,11 @@ jobs:
         env:
           CI: false
         run: |
-          dotnet build ${{ github.workspace }}/${{ env.TEST_PROJECT_PATH }} /bl --configuration ${{ env.DOTNET_CONFIGURATION }}
+          dotnet build ${{ github.workspace }}/${{ env.TEST_PROJECT_PATH }} /bl --configuration ${{ env.DOTNET_CONFIGURATION }} -o ${{ github.workspace }}/artifacts/${{ matrix.name }}
 
       - name: Run tests
         run: >-
-          dotnet test ${{ github.workspace }}/${{ env.TEST_PROJECT_PATH }}
+          dotnet run ${{ github.workspace }}/${{ env.TEST_PROJECT_PATH }}
```
Youssef1313: You need to:

1. Remove `--logger "console;verbosity=normal"`.
2. Replace `--logger trx` with `--report-trx` (and ensure Microsoft.Testing.Extensions.TrxReport is referenced)
3. Replace all the blame-related args with `--crashdump`, `--hangdump`, and `--hangdump-timeout 7m`
4. `--results-directory` is changing the meaning a little bit with this PR. Previously it would be relative to current working directory `${{ github.workspace }}` while with this PR it will be relative to the test executable. So either pass the full path, or keep using `dotnet test` which will handle the path transformation for you.
5. `--collect "XPlat Code Coverage"` needs to be replaced with `--coverage` (and ensure Microsoft.Testing.Extensions.CodeCoverage is referenced)

.github/workflows/tests.yaml
```diff
@@ -107,7 +107,15 @@ jobs:
           --results-directory testresults
           --collect "XPlat Code Coverage"
           --no-restore
-          --no-build -- RunConfiguration.CollectSourceInformation=true
+          --no-build
+          --filter-not-trait "category=failing"
+          --report-trx
+          --report-trx-filename "${{ matrix.name }}-${{ matrix.os }}.trx"
+          --ignore-exit-code 8
+          -- RunConfiguration.CollectSourceInformation=true
```
Youssef1313: This shouldn't be needed.

Directory.Packages.props
```diff
@@ -99,12 +100,14 @@
...
     <PackageVersion Include="OpenFeature.Contrib.Providers.Flagd" Version="0.3.3" />
     <PackageVersion Include="OpenFeature.Providers.Ofrep" Version="0.1.2" />
-    <PackageVersion Include="MartinCostello.Logging.XUnit" Version="0.6.0" />
+    <PackageVersion Include="MartinCostello.Logging.XUnit.v3" Version="0.6.0" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.14.1" />
-    <PackageVersion Include="xunit" Version="2.9.3" />
-    <PackageVersion Include="xunit.runner.visualstudio" Version="2.8.2" />
-    <PackageVersion Include="xunit.extensibility.execution" Version="2.9.3" />
-    <PackageVersion Include="Microsoft.DotNet.XUnitExtensions" Version="9.0.0-beta.24568.1" />
+    <PackageVersion Include="xunit.v3" Version="3.2.0" />
```
Youssef1313: nit: there is now `xunit.v3.mtp-v2` package.

Directory.Packages.props
```diff
@@ -99,12 +100,14 @@
...
+    <PackageVersion Include="xunit.v3.assert" Version="3.2.0" />
+    <PackageVersion Include="xunit.v3.extensibility.core" Version="3.2.0" />
+    <PackageVersion Include="xunit.runner.visualstudio" Version="3.1.5" />
```
Youssef1313, line 111: nit: no longer needed if you don't care about VSTest support.

tests/.runsettings
```diff
@@ -2,6 +2,6 @@
 <RunSettings>
   <RunConfiguration>
     <!-- Filter out failing (wrong framework, platform, runtime or activeissue) tests -->
-    <TestCaseFilter>category!=unsupported-platform</TestCaseFilter>
+    <TestCaseFilter>category!=failing</TestCaseFilter>
```
Youssef1313: runsettings isn't supported with MTP.

tests/Directory.Build.props
```diff
@@ -29,4 +35,13 @@
     <RunSettingsFilePath>$(MSBuildThisFileDirectory).runsettings</RunSettingsFilePath>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <TestRunnerAdditionalArguments>$(TestRunnerAdditionalArguments) --filter-not-trait &quot;category=failing&quot;</TestRunnerAdditionalArguments>
+    <!-- ignore exit code 8 (zero tests ran)
+         https://learn.microsoft.com/dotnet/core/testing/microsoft-testing-platform-exit-codes -->
+    <TestRunnerAdditionalArguments>$(TestRunnerAdditionalArguments) --ignore-exit-code 8</TestRunnerAdditionalArguments>
+
+    <!-- <UseMicrosoftTestingPlatformRunner>true</UseMicrosoftTestingPlatformRunner> -->
```
Youssef1313: Uncomment?

Evangelink:
@aaronpowell please feel free ping @Youssef1313 or I so we can move forward here.

Directory.Packages.props
Evangelink:
```suggestion
```

examples/powershell/CommunityToolkit.Aspire.PowerShell.AppHost/Program.cs
```diff
@@ -39,7 +39,7 @@ az storage blob upload --connection-string $myblob -c demo --file ./scripts/scri
 
 // outputs "the sum of 2 and 3 is 5"
 var script2 = ps.AddScript("script2", """
-    & ./scripts/script.ps1 @args
+    & ./Scripts/script.ps1 @args
```
Evangelink, line 61: I don't think that Powershell is case sensitive (cc @nohwnd)

tests/CommunityToolkit.Aspire.Hosting.McpInspector.Tests/AppHostTests.cs
```diff
@@ -3,7 +3,7 @@
 namespace CommunityToolkit.Aspire.Hosting.McpInspector.Tests;
 
 // Forcing linux only due to: https://github.com/modelcontextprotocol/inspector/issues/893
-[RequiresLinux]
+[RequiresWindows]
```
Evangelink: How was this working before?

examples/powershell/CommunityToolkit.Aspire.PowerShell.AppHost/Program.cs
nohwnd: This is up to the file system not up to powershell. If this needs to run on linux and the filesystem is case sensitive then ./scripts and ./Scripts are two different paths. On windows we typically don't use case sensitive file systems.

 https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_case-sensitivity?view=powershell-7.5

tests/CommunityToolkit.Aspire.Hosting.McpInspector.Tests/AppHostTests.cs
aaronpowell: I have no idea why I made that change, I'm going to revert it 🤣

Evangelink (commented):
@aaronpowell the latest errors are weird, it's complaining about a file that doesn't exist when I pull this branch.

I have also noticed the following missing changes:

1. In `global.json`, add:

```json
"test": {
    "runner": "Microsoft.Testing.Platform"
  }
```

1. In `Directory.Packages.props` and `tests/Directory.Build.props` remove:
- `coverlet.collector` -> not compatible with MTP, instead use MS coverage
- `Microsoft.NET.Test.Sdk` -> VSTest testhost part, no longer needed as you are moving to MTP
- `xunit.runner.visualstudio` -> VSTest adapter for xUnit, also no longer needed.

1. I'd suggest to be careful with the `--ignore-exit-code 8` set on all test projects because you are allowing all projects not to run tests without failure. I don't know if you have some other guard in place but if you don't then you would potentially no longer run any test without noticing.

PR #988, opened 2025-11-21
#940 upgrade papercut and changed portnumbers

### LINKED ISSUE #940: Upgrade Papercut SMTP tag
The following tags are available:
- 7.5
- 7.5.0

I guess that we should upgrade to 7.5.0?

anoordover:
**Closes #940**

## PR Checklist

- [ ] New integration
  - [ ] Docs are written

## Other information

@mguinness thanks for the pointer to the changed ports.
I don't know if this is a breaking change for this component.

### DIFF
--- a/src/CommunityToolkit.Aspire.Hosting.PapercutSmtp/README.md
+++ b/src/CommunityToolkit.Aspire.Hosting.PapercutSmtp/README.md
@@ -37,3 +37,16 @@ var xyz = builder.AddProject<Xyz>("application")
 
 builder.Build().Run();
 ```
+
+### Example 3: Get URI from connection-string using DbConnectionStringBuilder
+
+```csharp
+string? papercutConnectionString = builder.Configuration.GetConnectionString("papercut");
+DbConnectionStringBuilder connectionBuilder = new()
+{
+    ConnectionString = papercutConnectionString 
+};
+
+Uri endpoint = new(connectionBuilder["Endpoint"].ToString()!, UriKind.Absolute);
+builder.Services.AddScoped(_ => new SmtpClient(endpoint.Host, endpoint.Port));
+```


mguinness:
LGTM, I did some testing and revising the ports worked as expected.

The only problem I found was that I had to replace `httpPort` with `(int)HttpStatusCode.OK` in the following:

https://github.com/CommunityToolkit/Aspire/blob/4fc8f807ef9ea9107ce92f9b917fb4d72b75d322/src/CommunityToolkit.Aspire.Hosting.PapercutSmtp/PapercutSmtpHostingExtension.cs#L38

Also do you think it's worth adding some example code for the connection string to README?

```c#
var connStr = config.GetConnectionString("papercut");
var uri = new Uri(connStr[9..]);
var sc = new SmtpClient(uri.Host, uri.Port);
```

anoordover:
@mguinness thanks for pointing out this bug which might have been present from the start of this component.
I remove the status parameter from the call because the default value is 200.
@mguinness how did you notice this bug? Can I add a unittest to test this?
Do you have any pointers for me?

anoordover:
@mguinness about your second note: In my example project I used the DbConnectionStringBuilder and added this code to the readme.md.

mguinness:
I noticed the bug when I changed AddPapercutSmtp to specify ports.  

`var papercut = builder.AddPapercutSmtp("papercut", 8080, 2525)`

Then health check then failed, but I'm not sure why it didn't occur when `httpPort` variable was 80.

The message in [UriHealthCheck](
https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/846fbc99bcf60eafea23096c9aab05b8c8459b6b/src/HealthChecks.Uris/UriHealthCheck.cs#L56-L59) should be returned when status code 80 is incorrectly expected.  @sungam3r Any ideas?

sungam3r:
> The message in UriHealthCheck should be returned when status code 80 is incorrectly expected.

What do you mean?

anoordover:
Ah, now I see why my test doesn't fail. When you don't override the httpPort this parameter will be null on the line you mentioned @mguinness and the default httpStatus 200 will be used. Should I create two papercut resources in my example project and use them both in my unittest so as to make sure that setting the optional parameters won't break the healthcheck?
What do you think @mguinness ?

mguinness:
> What do you mean?

Apologies, I wasn't very clear.  I suppose the question should be whether bounds checking occur in `ExpectedHttpCodes` to ensure valid status code >= 100 and <=599 in [UriHealthCheckOptions](
https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/846fbc99bcf60eafea23096c9aab05b8c8459b6b/src/HealthChecks.Uris/UriHealthCheckOptions.cs#L91) before `httpClient.SendAsync` is called?

The check could also be done in [ResourceBuilderExtensions](https://github.com/dotnet/aspire/blob/a79afe3c1d0bc7d60b1905a36a2b27d4c2815080/src/Aspire.Hosting/ResourceBuilderExtensions.cs#L1820-L1828) and throw a `ArgumentOutOfRangeException` if statusCode is out of range.  Or more simply change the statusCode parameter from `int` to [HttpStatusCode](https://learn.microsoft.com/en-us/dotnet/api/system.net.httpstatuscode).

> Should I create two papercut resources in my example project and use them both in my unittest so as to make sure that setting the optional parameters won't break the healthcheck?

I don't think that is necessary since you have now removed the statusCode parameter from WithHttpHealthCheck.

anoordover:
@mguinness I have a local change ready to commit that I can push if we want to test whether setting the portnumbers on the papercut host will effect the healthcheck.

```csharp
var papercut = builder.AddPapercutSmtp(PapercutConstants.ResourceNameDefaultParameters);
var papercut2 = builder.AddPapercutSmtp(PapercutConstants.ResourceNameOverrideParameters, httpPort: 8080, smtpPort: 2525);
```

Now I can use the two resource-names in my unittest in a Theory. My previous code would fail on the second papercut-resource.
Shall I commit this?

Changes will be:
- Added a static class PapercutConstants containing the names of the two papercut resources;
- Added a second papercut resource in my example project;
- Changed my AppHostTests to run the test ResourceStartsAndMailShouldBeReceived as a Theory with the two resourcenames.

mguinness:
@aaronpowell might be better placed to answer.  I suppose you could test that a port number outside the range 1-65535 fails.

aaronpowell:
not sure I'm properly following - is the issue that if you set a port number outside a certain range things break unexpectedly?

mguinness:
I just tested and [AllocatedEndpoint](https://github.com/dotnet/aspire/blob/a79afe3c1d0bc7d60b1905a36a2b27d4c2815080/src/Aspire.Hosting/ApplicationModel/AllocatedEndpoint.cs#L58-L59) throws a `ArgumentOutOfRangeException`.

Arno, I think what you already have in the PR is sufficient for merging if it can be assigned for review.

aaronpowell:
Port range management should delegate to Aspire, unless a resource has very specific requirements

PR #989, opened 2025-11-22
Work around long path issue with VS on windows.

afscrome:
Without this, if you clone the repo to a path longer than ~25 chars, and try to build the toolkit in VS, it'll fail to build.

See https://developercommunity.visualstudio.com/t/GeneratedMsBuildEditorConfigeditorconf/11004277 .


ErikEJ:
Great, this has happened to me!

aaronpowell (approved):

PR #1021, opened 2025-12-01
feat: add Zitadel host integration

### LINKED ISSUE #1015: Hosting integration for Zitadel
### .NET Aspire issue link

https://github.com/dotnet/aspire/issues/6988

### Overview

There's been a previous integration request for Zitadel here: #350 but that eventually got closed due to stale.

I'd like to re-open the request and am willing to provide a PR (I already wrote most of the integration, but the PR template told me to open an issue)

### Usage example

ideally you'd be able to set up [zitadel](https://github.com/zitadel/zitadel) like this:

```csharp
builder.AddZitadel("zitadel")
```

### Additional context

Previous ticket: #350 
previous PR: #549 

both eventually have been closed due to lack of activity.

### Help us help you

dealloc:
**Closes #1015**

Adds host support for [Zitadel](https://github.com/zitadel/zitadel) to Aspire.

## PR Checklist

- [x] New integration
  - [x] Docs are written

## Other information

Currently in draft since I haven't figured out everything.
- HTTPS would be great if I could set this up by default, but I couldn't figure out how to get Aspire to share it's SSL
- Zitadel requires the hostname to be known, if you access from service discovery it would be `<name>` but if you open from the dashboard it's `localhost:<port>`. Currently it uses a semi-hardcoded `<name>.dev.localhost` as a middle ground.

I played around with something similar to Keycloak's `WithRealmImport` to "seed" the instance but AFAIK Zitadel doesn't support such things.

I need to update some of the docs that are missing and write the tests, not sure if the example I've included is sufficient for now

### DIFF
--- a/src/CommunityToolkit.Aspire.Hosting.Zitadel/ZitadelHostingExtensions.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.Zitadel/ZitadelHostingExtensions.cs
@@ -63,8 +63,6 @@ public static class ZitadelHostingExtensions
             )
             .WithHttpHealthCheck("/healthz")
             .WithEnvironment("ZITADEL_MASTERKEY", masterKeyParameter)
-            .WithEnvironment("ZITADEL_TLS_ENABLED", "false")
-            .WithEnvironment("ZITADEL_EXTERNALSECURE", "false")
             .WithUrlForEndpoint(ZitadelResource.HttpEndpointName, e => e.DisplayText = "Zitadel Dashboard");
 
 #pragma warning disable ASPIRECERTIFICATES001
--- a/tests/CommunityToolkit.Aspire.Hosting.Zitadel.Tests/AppHostTests.cs
+++ b/tests/CommunityToolkit.Aspire.Hosting.Zitadel.Tests/AppHostTests.cs
@@ -25,7 +25,7 @@ public class AppHostTests(
         // Test the health endpoint
         var request = new HttpRequestMessage(HttpMethod.Get, "/.well-known/openid-configuration");
         // Needs to match the external domain for Zitadel or we get a 404
-        request.Headers.Host = $"{resourceName}.dev.localhost";
+        request.Headers.Host = $"{fixture.App.GetEndpoint(resourceName, "http").Host}";
         var response = await httpClient.SendAsync(request);
 
         Assert.Equal(HttpStatusCode.OK, response.StatusCode);
@@ -46,7 +46,7 @@ public class AppHostTests(
         // Test the health endpoint
         var request = new HttpRequestMessage(HttpMethod.Get, "/");
         // Needs to match the external domain for Zitadel or we get a 404
-        request.Headers.Host = $"{resourceName}.dev.localhost";
+        request.Headers.Host = $"{fixture.App.GetEndpoint(resourceName, "http").Host}";
         var response = await httpClient.SendAsync(request);
 
         Assert.Equal(HttpStatusCode.OK, response.StatusCode);
--- a/src/CommunityToolkit.Aspire.Hosting.Zitadel/ZitadelContainerImageTags.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.Zitadel/ZitadelContainerImageTags.cs
@@ -8,6 +8,6 @@ internal static class ZitadelContainerImageTags
     /// <remarks>zitadel/zitadel</remarks>
     public const string Image = "zitadel/zitadel";
 
-    /// <remarks>v4.7.0</remarks>
-    public const string Tag = "v4.7.0";
+    /// <remarks>v4.9.0</remarks>
+    public const string Tag = "v4.9.0";
 }
--- a/tests/CommunityToolkit.Aspire.Hosting.Zitadel.Tests/ZitadelHostingExtensionsTests.cs
+++ b/tests/CommunityToolkit.Aspire.Hosting.Zitadel.Tests/ZitadelHostingExtensionsTests.cs
@@ -1,6 +1,7 @@
 using Aspire.Hosting;
 using Aspire.Hosting.ApplicationModel;
 using CommunityToolkit.Aspire.Hosting.Zitadel;
+using CommunityToolkit.Aspire.Testing;
 
 namespace CommunityToolkit.Aspire.Hosting.Zitadel.Tests;
 
@@ -47,7 +48,7 @@ public class ZitadelHostingExtensionsTests
 
         var zitadel = builder.AddZitadel("zitadel");
 
-        var env = await zitadel.Resource.GetEnvironmentVariableValuesAsync();
+        var env = await zitadel.Resource.GetEnvironmentVariablesAsync();
 
         Assert.Equal("false", env["ZITADEL_TLS_ENABLED"]);
         Assert.Equal("false", env["ZITADEL_EXTERNALSECURE"]);
@@ -63,7 +64,7 @@ public class ZitadelHostingExtensionsTests
 
         var zitadel = builder.AddZitadel("zitadel");
 
-        var env = await zitadel.Resource.GetEnvironmentVariableValuesAsync();
+        var env = await zitadel.Resource.GetEnvironmentVariablesAsync();
 
         Assert.NotNull(zitadel.Resource.AdminUsernameParameter);
         Assert.NotNull(zitadel.Resource.AdminPasswordParameter);
@@ -92,7 +93,7 @@ public class ZitadelHostingExtensionsTests
 
         var zitadel = builder.AddZitadel("zitadel", masterKey: masterKey);
 
-        var env = await zitadel.Resource.GetEnvironmentVariableValuesAsync();
+        var env = await zitadel.Resource.GetEnvironmentVariablesAsync();
 
         Assert.True(env.ContainsKey("ZITADEL_MASTERKEY"));
     }
@@ -129,18 +130,6 @@ public class ZitadelHostingExtensionsTests
     }
 
     [Fact]
-    public async Task AddZitadel_Uses_Custom_ExternalDomain()
-    {
-        var builder = DistributedApplication.CreateBuilder();
-
-        var zitadel = builder.AddZitadel("zitadel", externalDomain: "auth.example.com");
-
-        var env = await zitadel.Resource.GetEnvironmentVariableValuesAsync();
-
-        Assert.Equal("auth.example.com", env["ZITADEL_EXTERNALDOMAIN"]);
-    }
-
-    [Fact]
     public async Task WithExternalDomain_Overrides_Default()
     {
         var builder = DistributedApplication.CreateBuilder();
@@ -148,7 +137,7 @@ public class ZitadelHostingExtensionsTests
         var zitadel = builder.AddZitadel("zitadel")
             .WithExternalDomain("custom.domain.com");
 
-        var env = await zitadel.Resource.GetEnvironmentVariableValuesAsync();
+        var env = await zitadel.Resource.GetEnvironmentVariablesAsync();
 
         Assert.Equal("custom.domain.com", env["ZITADEL_EXTERNALDOMAIN"]);
     }
@@ -158,10 +147,10 @@ public class ZitadelHostingExtensionsTests
     {
         var builder = DistributedApplication.CreateBuilder();
 
-        var zitadel = builder.AddZitadel("zitadel", externalDomain: "first.example.com")
+        var zitadel = builder.AddZitadel("zitadel")
             .WithExternalDomain("second.example.com");
 
-        var env = await zitadel.Resource.GetEnvironmentVariableValuesAsync();
+        var env = await zitadel.Resource.GetEnvironmentVariablesAsync();
 
         // WithExternalDomain should override the parameter
         Assert.Equal("second.example.com", env["ZITADEL_EXTERNALDOMAIN"]);
--- a/tests/CommunityToolkit.Aspire.Hosting.Zitadel.Tests/ZitadelIntegrationTests.cs
+++ b/tests/CommunityToolkit.Aspire.Hosting.Zitadel.Tests/ZitadelIntegrationTests.cs
@@ -46,7 +46,7 @@ public class ZitadelIntegrationTests(
         var zitadelResource = appModel.Resources.OfType<ZitadelResource>()
             .Single(r => r.Name == zitadelName);
 
-        var env = await zitadelResource.GetEnvironmentVariableValuesAsync();
+        var env = await zitadelResource.GetEnvironmentVariablesAsync();
 
         Assert.True(env.ContainsKey("ZITADEL_DATABASE_POSTGRES_HOST"));
         Assert.True(env.ContainsKey("ZITADEL_DATABASE_POSTGRES_PORT"));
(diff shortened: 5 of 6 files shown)


dealloc:
@dotnet-policy-service agree

tests/CommunityToolkit.Aspire.Hosting.Zitadel.Tests/AppHostTests.cs
aaronpowell, line 1: Is there something more that could be done with the demo app that ensures everything is wired up properly and zitadel is doing what is expected of it? Calling the `healthz` endpoint is kind of implied by the resource becoming healthy.

dealloc:
> You mentioned a problem with hostnames, can you tag the line(s) where that's surfacing?

Currently that would be https://github.com/CommunityToolkit/Aspire/pull/1021/files/677452ebe38d7be79caf11268057182f5594136d#diff-427aa76f92a729429b7b134c3be02f0a722e8ab12d85a50a2bc822d9819cda33R35

Zitadel requires a "stable" external hostname or it'll return a 404 if the `Host` header does not match what's configured.

This means that if I configure the external hostname as `localhost` and I would call it from Aspire's service discovery (e.g `zitadel:<port>`) it would return a 404 as it can't "match" to an instance. On the other hand setting it to the resource name would mean you can't open it in the browser from Aspire's dashboard.

Currently I'm setting it to `<name>.dev.localhost` which seemed a good middle ground but I can't seem to override the URL shown in the dashboard, and I'm not sure if this is the "idiomatic" way to handle this in Aspire either

tests/CommunityToolkit.Aspire.Hosting.Zitadel.Tests/AppHostTests.cs
dealloc: Updated it so that it calls the OpenID Connect Discovery Endpoint and checks if the dashboard is being served

dealloc:
rebased changes on `main`

aaronpowell:
Is there a reason that the `host:port` that Aspire will assign the container couldn't be provided for the external domain?

dealloc:
> Is there a reason that the `host:port` that Aspire will assign the container couldn't be provided for the external domain?

How would I retrieve that?

aaronpowell:
> > Is there a reason that the `host:port` that Aspire will assign the container couldn't be provided for the external domain?
> 
> How would I retrieve that?

Use the endpoint reference and get the allocated endpoint, that will have all the parts of the endpoint on it

dealloc:
@aaronpowell one final tweak I'd like to make is that the Aspire dashboard generates the URL to the dashboard with `localhost:<port>` while Zitadel will expect the same hostname as passed to `ZITADEL_EXTERNALDOMAIN` (which now properly resolves to the endpoint's hostname like `zitadel`)

it's a minor thing as this only affects the link shown in the dashboard, but I feel it would be a lot more polished if I managed to get that working as well

aaronpowell:
I think I'm still not properly understanding the role of the `EXTERNALDOMAIN` environment variable.

The endpoint reference host is going to be pointing to the host that the resource is exposed within the container network, so `zitadel` in this case, which is how other containers can address it (which is how the zitadel resource talks to the PG database using `postgres` as the host).

Resources outside the network, say your browser, would use the host `localhost` and the port that was assigned via aspire, not `zitadel:8080` since you're not in the same container network.

So then, shouldn't the `EXTERNALDOMAIN` be `localhost`? Looking at the sample [`docker-compose.yml`](https://raw.githubusercontent.com/zitadel/zitadel/main/docs/docs/self-hosting/deploy/docker-compose.yaml) that is what they have set.

ahmedhemdan21:
> I think I'm still not properly understanding the role of the `EXTERNALDOMAIN` environment variable.
> 
> The endpoint reference host is going to be pointing to the host that the resource is exposed within the container network, so `zitadel` in this case, which is how other containers can address it (which is how the zitadel resource talks to the PG database using `postgres` as the host).
> 
> Resources outside the network, say your browser, would use the host `localhost` and the port that was assigned via aspire, not `zitadel:8080` since you're not in the same container network.
> 
> So then, shouldn't the `EXTERNALDOMAIN` be `localhost`? Looking at the sample [`docker-compose.yml`](https://raw.githubusercontent.com/zitadel/zitadel/main/docs/docs/self-hosting/deploy/docker-compose.yaml) that is what they have set.

Agreed, whenever I run without setting the external domain I get 

`unable to set instance using origin &{localhost:28080  http} (ExternalDomain is zitadel)`
while adding `WithExternalDomain("localhost")` make it work without any issues.

dealloc:
> I think I'm still not properly understanding the role of the `EXTERNALDOMAIN` environment variable.
> 
> The endpoint reference host is going to be pointing to the host that the resource is exposed within the container network, so `zitadel` in this case, which is how other containers can address it (which is how the zitadel resource talks to the PG database using `postgres` as the host).
> 
> Resources outside the network, say your browser, would use the host `localhost` and the port that was assigned via aspire, not `zitadel:8080` since you're not in the same container network.
> 
> So then, shouldn't the `EXTERNALDOMAIN` be `localhost`? Looking at the sample [`docker-compose.yml`](https://raw.githubusercontent.com/zitadel/zitadel/main/docs/docs/self-hosting/deploy/docker-compose.yaml) that is what they have set.

A single Zitadel instance can have multiple "virtual" instances running and it uses the `Host` header on incoming requests to differentiate which instance to select.

This means if we set `EXTERNALDOMAIN` to `localhost` it won't "recognize" which instance to select if a resource calls `zitadel` (even though Aspire would resolve to the same service), which in turn results in Zitadel responding with the `unable to set instance using origin &{localhost:28080  http} (ExternalDomain is zitadel)` error

Let's assume the following setup in Aspire:
```csharp

var database = builder.AddPostgres("postgres");

builder.AddZitadel("zitadel")
    .WithDatabase(database);

builder.AddProject<Example>("my-example")
    .WithReference(zitadel);

builder.Build().Run();
```

if we set `EXTERNALDOMAIN` to `localhost`:
- the user navigating in his browser to the instance would **WORK** (because the browser passes `localhost` as the hostname)
- the example application (which would call `zitadel`) would **NOT WORK** because it sets the hostname as `zitadel`

As a middle ground I set `zitadel.dev.localhost` since it would work for both scenarios.

I hope that helps clarify the role of `EXTERNALDOMAIN` and the issue with `localhost` / `<resourcename>`

danegsta:
@dealloc APIs to make the .NET dev cert available to resources (and configure them to use it) are in Aspire main and will ship with 13.1 as early as next week, so you'd be able to get HTTPS working once that's available.

danegsta:
Also, as a heads up, `*.internal` addresses work reliably in every major browser, but support isn't consistent at the OS level. You can visit `zitadel.dev.internal` in a browser and it'll resolve correctly to localhost, but if you try to do the same outside a browser (say with curl) it won't always work. Depends on the particular system DNS setup.

src/CommunityToolkit.Aspire.Hosting.Zitadel/ZitadelHostingExtensions.cs
```diff
@@ -0,0 +1,141 @@
...
+                name: ZitadelResource.HttpEndpointName
+            )
+            .WithHttpHealthCheck("/healthz")
+            .WithEnvironment("ZITADEL_MASTERKEY", masterKeyParameter)
+            .WithEnvironment("ZITADEL_TLS_ENABLED", "false")
+            .WithEnvironment("ZITADEL_EXTERNALSECURE", "false")
+            .WithUrlForEndpoint(ZitadelResource.HttpEndpointName, e => e.DisplayText = "Zitadel Dashboard");
+
+        // Use ReferenceExpression for the port to avoid issues with endpoint allocation
+        var endpoint = resource.GetEndpoint(ZitadelResource.HttpEndpointName);
```
danegsta: This will get a reference to the endpoint from the perspective of the host, so it'll resolve the localhost domain and port instead of the internal container address and port. That being said there's a bug that'll need to be fixed that's preventing the network identifier from actually applying: https://github.com/dotnet/aspire/issues/13440

```suggestion
        var endpoint = resource.GetEndpoint(ZitadelResource.HttpEndpointName, KnownNetworkIdentifiers.LocalhostNetwork);
```

src/CommunityToolkit.Aspire.Hosting.Zitadel/ZitadelHostingExtensions.cs
```diff
@@ -0,0 +1,141 @@
...
+            .WithArgs("start-from-init", "--masterkeyFromEnv")
+            .WithHttpEndpoint(
+                targetPort: 8080,
+                port: port,
+                name: ZitadelResource.HttpEndpointName
+            )
+            .WithHttpHealthCheck("/healthz")
+            .WithEnvironment("ZITADEL_MASTERKEY", masterKeyParameter)
+            .WithEnvironment("ZITADEL_TLS_ENABLED", "false")
+            .WithEnvironment("ZITADEL_EXTERNALSECURE", "false")
```
danegsta: Once Aspire 13.1 is available, this will get you HTTPS support. It'll set the appropriate config if the resource is configured with a certificate (it'll get the ASP.NET development certificate by default):
```suggestion
            .WithEnvironment("ZITADEL_TLS_ENABLED", "false")
            .WithEnvironment("ZITADEL_EXTERNALSECURE", "false")
            .WithHttpsCertificateConfiguration(ctx =>
            {
                ctx.EnvironmentVariables["ZITADEL_EXTERNALSECURE"] = "true";
                ctx.EnvironmentVariables["ZITADEL_TLS_ENABLED"] = "true";
                ctx.EnvironmentVariables["ZITADEL_TLS_CERTPATH"] = ctx.CertificatePath;
                ctx.EnvironmentVariables["ZITADEL_TLS_KEYPATH"] = ctx.KeyPath;

                return Task.CompletedTask;
            })
```

danegsta: You'll also need something like this to optionally enable HTTPS on the endpoint. It's a bit boilerplate currently as we're still working on the API for updating endpoints:

```C#
if (builder.ExecutionContext.IsRunMode)
{
    builder.Eventing.Subscribe<BeforeStartEvent>((@event, cancellationToken) =>
    {
        var developerCertificateService = @event.Services.GetRequiredService<IDeveloperCertificateService>();

        bool addHttps = false;
        if (!zitadelBuilder.Resource.TryGetLastAnnotation<HttpsCertificateAnnotation>(out var annotation))
        {
            if (developerCertificateService.UseForHttps)
            {
                // If no certificate is configured, and the developer certificate service supports container trust,
                // configure the resource to use the developer certificate for its key pair.
                addHttps = true;
            }
        }
        else if (annotation.UseDeveloperCertificate.GetValueOrDefault(developerCertificateService.UseForHttps) || annotation.Certificate is not null)
        {
            addHttps = true;
        }

        if (addHttps)
        {
            // If a TLS certificate is configured, override the endpoint to use HTTPS instead of HTTP
            // Zitadel only binds to a single port
            zitadelBuilder
                .WithEndpoint(ZitadelResource.HttpEndpointName, ep => ep.UriScheme = "https");
        }

        return Task.CompletedTask;
    });
}
```

danegsta:
Added a couple comments on how to get the endpoint reference to give you the correct `ZITADEL_EXTERNALDOMAIN` value and get HTTPS working (once 13.1 ships):

image

dealloc:
> Added a couple comments on how to get the endpoint reference to give you the correct `ZITADEL_EXTERNALDOMAIN` value and get HTTPS working (once 13.1 ships):
> image

The API's didn't seem available yet even after rebasing on upstream main; but I've added the code changes and I'll patiently await Aspire 13.1 and re-rebase my changes.

Appreciate the helpful input @danegsta!

danegsta:
> > Added a couple comments on how to get the endpoint reference to give you the correct `ZITADEL_EXTERNALDOMAIN` value and get HTTPS working (once 13.1 ships):
> > image
> 
> The API's didn't seem available yet even after rebasing on upstream main; but I've added the code changes and I'll patiently await Aspire 13.1 and re-rebase my changes.
> 
> Appreciate the helpful input @danegsta!

Some of the APIs are behind an experimental flag, so you may need to suppress the warning from experimental API usage to access them.

dealloc:
> > The API's didn't seem available yet even after rebasing on upstream main; but I've added the code changes and I'll patiently await Aspire 13.1 and re-rebase my changes.
> > Appreciate the helpful input @danegsta!
> 
> Some of the APIs are behind an experimental flag, so you may need to suppress the warning from experimental API usage to access them.

I must be doing something wrong then 😅
Screenshot 2025-12-10 at 22 22 10

danegsta:
> > > The API's didn't seem available yet even after rebasing on upstream main; but I've added the code changes and I'll patiently await Aspire 13.1 and re-rebase my changes.
> > > Appreciate the helpful input @danegsta!
> > 
> > 
> > Some of the APIs are behind an experimental flag, so you may need to suppress the warning from experimental API usage to access them.
> 
> I must be doing something wrong then 😅 Screenshot 2025-12-10 at 22 22 10

Those types are only available in Aspire main (13.1.0-preview) releases; they should ship as the new stable next week.

dealloc:
do I need to rebase onto `13.1.0-preview` and have this PR target that instead, or do we just wait for it to hit main and continue this then?

danegsta:
> do I need to rebase onto `13.1.0-preview` and have this PR target that instead, or do we just wait for it to hit main and continue this then?

I'd wait for 13.1 to be release and main to be updated for those HTTPS endpoint changes.

aaronpowell:
I'm starting the PR to update to 13.1 (see #1056) and you could retarget this branch to that one if you want to.

dealloc:
I feel like I'm still missing something, my branch is rebased on latest `main` but several methods in @danegsta's samples aren't available (like `WithHttpsCertificateConfiguration`)

I've tried rebasing on `aspire-13.1` but they don't exist there.

I'll wait for `13.1` to be on main and I'll give it another go then 👍

dealloc:
rebased and retarget to `aspire-13.1` as requested by @aaronpowell 
I am noticing that I can't seem to run the tests anymore (see below)

image

examples/zitadel/CommunityToolkit.Aspire.Hosting.Zitadel.AppHost/CommunityToolkit.Aspire.Hosting.Zitadel.AppHost.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Aspire.AppHost.Sdk/13.0.0">
```
ahmedhemdan21: I think this should be 13.1.0

dealloc:
With the holidays I haven't had much time to work on this, is it possible to re-open and extend the stale?

dealloc:
think this is ready for review :)

dealloc:
@aaronpowell can you clarify why this PR was closed?

aaronpowell:
oh drat - because I merged the 13.1 branch (which then automatically deletes the branch) it must cascade down to close PR's that target that branch.

aaronpowell:
I can't change the base the PR targets, if you can then it can be reopened, or create a new PR from the zitadel branch targeting `main` (since 13.1 is there now).

PR #1071, opened 2025-12-18
More update to xUnit v3 PR

Evangelink:
@aaronpowell Some updates for your PR


aaronpowell (approved):

aaronpowell:
It looks like the PowerShell tests are still not exiting, which is the centre of the problem we're facing (the other two tests seem to have failed due to Actions issues)

PR #1090, opened 2026-01-09
feat: create AzureStorageExplorer hosting integration

Odonno:
Given https://github.com/dotnet/aspire/issues/3778 and the need for an [Azure Storage Explorer](https://github.com/sebagomez/azurestorageexplorer), I create a new package called `CommunityToolkit.Aspire.Hosting.Azure.Extensions`.

This is a 3rd party project that is constantly maintained (last commit last month), has 141 stars on GitHub and 2M+ downloads on Docker. 

## PR Checklist

- [x] New integration
  - [x] Docs are written

### DIFF
--- a/tests/CommunityToolkit.Aspire.Hosting.Azure.Extensions.Tests/AppHostTests.cs
+++ b/tests/CommunityToolkit.Aspire.Hosting.Azure.Extensions.Tests/AppHostTests.cs
@@ -1,3 +1,4 @@
+using Aspire.Components.Common.Tests;
 using CommunityToolkit.Aspire.Testing;
 
 namespace CommunityToolkit.Aspire.Hosting.Azure.Extensions.Tests;


Odonno:
@aaronpowell I am curious what you think of the public API design before writing some tests.

aaronpowell (commented):
For consistency sake, I think it makes more sense of the public API to be:

```csharp
builder.AddAzureStorage("storage")
    .RunAsEumulator()
    .WithAzureStorageExplorer();
```

This is then consistent with the other web UI tools for inspecting storage resources (pgadmin, dbgate, etc.), it also removes the ambiguity that you might see if you were to do this on the current API:

```csharp
var azureStorageExplorer = builder
    .AddAzureStorageExplorer("explorer")
    .WithAzurite()
    .WithBlobs(blobs)
    .WithQueues(queue);
```

Odonno:
> For consistency sake, I think it makes more sense of the public API to be:
> 
> ```cs
> builder.AddAzureStorage("storage")
>     .RunAsEumulator()
>     .WithAzureStorageExplorer();
> ```
> 
> This is then consistent with the other web UI tools for inspecting storage resources (pgadmin, dbgate, etc.), it also removes the ambiguity that you might see if you were to do this on the current API:
> 
> ```cs
> var azureStorageExplorer = builder
>     .AddAzureStorageExplorer("explorer")
>     .WithAzurite()
>     .WithBlobs(blobs)
>     .WithQueues(queue);
> ```

That is reasonably good. However, the `AzureStorageResource` does not currently expose a connection string. I can only attach to a blob/table/queue resource. Do you want to go this way:

```cs
storage.AddBlobs("blobs").WithAzureStorageExplorer();
```

aaronpowell:
> That is reasonably good. However, the `AzureStorageResource` does not currently expose a connection string.

I didn't realise that.

Yes, then it makes sense to have it off the blob/queue/etc. resource instead. Looking at the storage explorer, it seems like it's not a singleton resource either, so if you had blobs + queues it would make sense to have multiple explorers right?

Odonno:
> ```cs
> WithAzureStorageExplorer
> ```

AFAIK I know, yes, an explorer can only a single Azure storage/emulator resource. I suppose this is a feature that can be worked on in the future for this project.

Best option would be to get the full connection string from the Azure storage resource but until then, we can only work with an instance per blobs/queues/tables resource unfortunateyl.

src/CommunityToolkit.Aspire.Hosting.Azure.Extensions/AzureStorageExplorerResource.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AzureStorageExplorerResource"/> class.
+    /// </summary>
+    /// <param name="name">The name of the resource.</param>
+    public AzureStorageExplorerResource(
+        [ResourceName] string name
+    ) : base(name)
+    {
+        PrimaryEndpoint = new(this, PrimaryEndpointName);
+    }
```
aaronpowell: Can you turn this into a primary constructor so it's consistent with other types

Odonno:
I am not sure what is the issue. The tests work on my machine.

tests/CommunityToolkit.Aspire.Hosting.Azure.Extensions.Tests/AppHostTests.cs
```diff
@@ -0,0 +1,19 @@
+using CommunityToolkit.Aspire.Testing;
+
+namespace CommunityToolkit.Aspire.Hosting.Azure.Extensions.Tests;
+
+public class AppHostTests(AspireIntegrationTestFixture<Projects.CommunityToolkit_Aspire_Azure_Extensions_AppHost> fixture) : IClassFixture<AspireIntegrationTestFixture<Projects.CommunityToolkit_Aspire_Azure_Extensions_AppHost>>
```
aaronpowell, line 7:
```suggestion
[RequiresDocker]
public class AppHostTests(AspireIntegrationTestFixture<Projects.CommunityToolkit_Aspire_Azure_Extensions_AppHost> fixture) : IClassFixture<AspireIntegrationTestFixture<Projects.CommunityToolkit_Aspire_Azure_Extensions_AppHost>>
```

aaronpowell (changes requested):
I think it fails on Windows because it needs docker

tests/CommunityToolkit.Aspire.Hosting.Azure.Extensions.Tests/AppHostTests.cs
```diff
@@ -0,0 +1,20 @@
+using CommunityToolkit.Aspire.Testing;
```
aaronpowell, line 2:
```suggestion
using Aspire.Components.Common.Tests;
using CommunityToolkit.Aspire.Testing;
```

aaronpowell (approved):

PR #1098, opened 2026-01-14
fix: disable parallelization of surrealdb conformance tests

### LINKED ISSUE #974: SurrealDB test stability
The tests for the SurrealDB client integration are not always successful on the first run, it looks like something is timing out when releasing a connection, see [here](https://github.com/CommunityToolkit/Aspire/actions/runs/19412409375/job/55535656047) for an example.

The test results asset contains a dump file that might help understand what isn't being released.

Odonno:
Fixes #974 

## PR Checklist


Odonno:
For context, `CollectionDefinition` only seem to work with child class and not the base class.

aaronpowell:
Is that why they haven't been shutting down properly?!

Odonno:
Yes, I reproduced the issue locally and this code fixes it.

aaronpowell (approved):

PR #1100, opened 2026-01-15
Hosting integration for Zitadel

### LINKED ISSUE #1015: Hosting integration for Zitadel

dealloc:
**Closes #1015 **

This PR is a continuation of the previous PR for Zitadel support: #1021

## PR Checklist

- [X] New integration
  - [X] Docs are written
  - [X] Added description of major feature to project description for NuGet package (4000 total character limit, so don't push entire description over that)

### DIFF
--- a/.devcontainer/post-create.sh
+++ b/.devcontainer/post-create.sh
@@ -5,7 +5,13 @@ sudo apt-get update && \
     sudo rm -rf /var/lib/apt/lists/*
 
 echo Install .NET dev certs
-dotnet dev-certs https --trust
+EXIT_CODE=0
+PARTIAL_TRUST_EXIT_CODE=4 # Exit code 4 indicates "partial trust" per aspnetcore#65391.
+dotnet dev-certs https --trust || EXIT_CODE=$?
+if [ "$EXIT_CODE" -ne 0 ] && [ "$EXIT_CODE" -ne "$PARTIAL_TRUST_EXIT_CODE" ]; then
+  echo "dotnet dev-certs https --trust failed with exit code $EXIT_CODE"
+  exit "$EXIT_CODE"
+fi
 
 echo Install Aspire
 curl -sSL https://aspire.dev/install.sh | bash
--- a/.github/actions/setup-runtimes-caching/action.yml
+++ b/.github/actions/setup-runtimes-caching/action.yml
@@ -85,7 +85,13 @@ runs:
       shell: bash
       if: ${{ matrix.os == 'ubuntu-latest' }}
       run: |
-        dotnet dev-certs https --trust
+        EXIT_CODE=0
+        dotnet dev-certs https --trust || EXIT_CODE=$?
+        PARTIAL_TRUST_EXIT_CODE=4 # See aspnetcore#65391: exit code 4 indicates partial trust
+        if [ "$EXIT_CODE" -ne 0 ] && [ "$EXIT_CODE" -ne "$PARTIAL_TRUST_EXIT_CODE" ]; then
+          echo "dotnet dev-certs https --trust failed with exit code $EXIT_CODE"
+          exit "$EXIT_CODE"
+        fi
 
     - name: Setup Node globals
       shell: bash


aaronpowell:
I'm seeing failures in the test log starting here: https://github.com/CommunityToolkit/Aspire/actions/runs/21052154522/job/60540747581?pr=1100#step:6:497

The tests themselves don't seem to report failing, so possibly the `Assert` isn't capturing the stack properly.

Do the test pass locally?

dealloc:
@aaronpowell the problem seems to happen when `GetEnvironmentVariablesAsync` is called and the tests just 'hang'

dealloc:
I don't mind continuing to work on this, I just need input on why `GetEnvironmentVariablesAsync` causes the tests to hang.
From what I can tell it's whenever a Postgres integration is also in play

dealloc:
@aaronpowell any input on the `GetEnvironmentVariablesAsync`?

aaronpowell:
@dealloc sorry, been under the pump with some other stuff recently. I've just pushed a commit which fixes the timeout issue. I think some of the asserts are now failing though

dealloc:
no worries, just wanted to make sure this PR didn't slip under the radar :)
One of the failing asserts is `Assert.Equal("false", env["ZITADEL_TLS_ENABLED"]);`

it seems that `WithHttpsCertificateConfiguration` didn't run yet as I set those vars in the callback.

```csharp
    [Fact]
    public async Task AddZitadel_Sets_Default_Environment_Variables()
    {
        var builder = DistributedApplication.CreateBuilder();

        var zitadel = builder.AddZitadel("zitadel");

        var env = await zitadel.Resource.GetEnvironmentVariablesAsync();

        Assert.Equal("false", env["ZITADEL_TLS_ENABLED"]);
        Assert.Equal("false", env["ZITADEL_EXTERNALSECURE"]);
        Assert.Equal("zitadel.dev.localhost", env["ZITADEL_EXTERNALDOMAIN"]);
        Assert.Equal("false", env["ZITADEL_DEFAULTINSTANCE_FEATURES_LOGINV2_REQUIRED"]);
        Assert.Equal("false", env["ZITADEL_FIRSTINSTANCE_ORG_HUMAN_PASSWORDCHANGEREQUIRED"]);
    }
```

I suspect I need to do something so Aspire actually calls the callback for it?

aaronpowell:
To test that, I _think_ you'll need to actually start the app host. Looking into the Aspire source I found this test https://github.com/dotnet/aspire/blob/a6b68c75e6a789e938c9e6630073e27231fb06f9/tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs#L768-L809

https://github.com/dotnet/aspire/blob/a6b68c75e6a789e938c9e6630073e27231fb06f9/tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs#L797 is a test against something setup in the callback to `WithHttpsCertificateConfiguration`, so you'll need to do something along these lines to create the app from the builder and start it, then resolve the resource: https://github.com/dotnet/aspire/blob/a6b68c75e6a789e938c9e6630073e27231fb06f9/tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs#L784-L788

We've got a bunch of tests that do that already in the toolkit codebase you can use as reference.

dealloc:
turns out the specific assertion I was making was incorrect, it's never set to `false` only to `true` if the callback would run (since `false` is a default value anyway)

PR #1124, opened 2026-02-04
Treat SQL projects as "regular" projects and enable publishing

jmezach:
**Closes #<ISSUE_NUMBER>**

## PR Checklist

- [ ] New integration
  - [ ] Docs are written

### DIFF
--- a/src/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects/SqlProjectBuilderExtensions.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects/SqlProjectBuilderExtensions.cs
@@ -94,8 +94,7 @@ public static class SqlProjectBuilderExtensions
                           Properties = properties,
                           ResourceType = "SqlProject",
                           State = KnownResourceStates.Waiting
-                      })
-                      .ExcludeFromManifest();
+                      });
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


jmezach:
@ErikEJ One aspect of this is of course that this requires the referenced MSBuild.Sdk.SqlProj to target at least version 4.0. If you don't you get a fairly generic message stating that the `dotnet publish` failed with a non-zero exit code

ErikEJ:
@jmezach Could we detect the SDK version in use?

jmezach:
@ErikEJ We are already reading MSBuild properties from the project file, so I guess we could use that to detect the SDK version. Although I'm not sure if we're currently setting an MSBuild property in `Sdk.props` that contains the version. Perhaps MSBuild already does so since it has to resolve the SDK anyway, but I'm not sure.

Also have some failing tests know that I have to figure out ;).

jmezach:
It appears that inheriting our `SqlProjectResource` from `ProjectResource` is also opting us into some other behaviours in Aspire that don't translate very well to SQL projects. For example I'm now seeing this:

```
 Unable to proceed with project '/Users/jmezach/Code/open-source/rr-wfm/CommunityToolkit.Aspire/examples/sql-database-projects/SdkProject/SdkProject.csproj'.
 Ensure you have a runnable project type.
 A runnable project should target a runnable TFM (for instance, net10.0) and have OutputType 'Exe'.
 The current OutputType is 'Library'.
```

Fun fact is that after that I'm seeing our regular output of actually deploying the project to the database so it seems that there are now two things happening when you start the resource.

@ErikEJ Maybe this isn't the right approach?

ErikEJ:
@jmezach yeah, I am not sure this is the right approach, I think an Aspire project resource is considered "executable" - maybe @DamianEdwards or @davidfowl are able to assist here? 🙏

jmezach:
@ErikEJ @DamianEdwards @davidfowl Would still appreciate some feedback on this approach 🙏

ErikEJ:
@jmezach I will try to make some in-person noise soon 😅

PR #1136, opened 2026-02-23
Add Neon integration

### LINKED ISSUE #398: Add Neon integration
### .NET Aspire issue link

https://github.com/dotnet/aspire/issues/7118

### Overview

The Aspire framework does not currently support [Neon](https://neon.tech/), a modern, open-source serverless PostgreSQL-compatible database. It has [managed integration for Azure](https://learn.microsoft.com/en-us/azure/partner-solutions/neon/overview) as well. While it is possible to use the existing client integrations for [PostgreSQL](https://github.com/dotnet/docs-aspire/blob/main/docs/database/postgresql-entity-framework-integration.md) or [Npgsql](https://github.com/dotnet/aspire/blob/main/src/Components/Aspire.Npgsql/README.md), Neon offers additional features like [database branching](https://neon.tech/docs/introduction/branching) that are not supported out of the box. We want to enable these unique features through dedicated .NET packages for Aspire.

### Usage example

# Here is an example of usage

## Hosting Integration for Neon

To integrate open source Neon as a hosted resource in a .NET Aspire application:

Install the `Aspire.Hosting.Neon` NuGet package:

`dotnet add package Aspire.Hosting.Neon`

Add the Neon server resource in the app host project:

```csharp

// Add Neon server resource with branching support
var neon = builder.AddNeonProject("neon");

// Add a Neon database
var neonDb = neon.AddDatabase("neondb");

// Reference the Neon database in a project
var exampleProject = builder.AddProject<Projects.ExampleProject>()
                            .WithReference(neonDb);

// Initialize and run the application
builder.Build().Run();
```

## Client Integration for Neon

To enable client projects to connect to Neon:

Install the Aspire.Neon NuGet package:

`dotnet add package Aspire.Neon`

Add the Neon client to the Program.cs file in the client project:

```csharp
builder.AddNeonDataSource("neondb").addBranch("dev");
```
Both integrations should align with Aspire's existing conventions and provide a similar experience to the PostgreSQL integration.

### Additional context

We are looking for .NET Aspire experts from the community to help us in developing this integration. Let me know your feedback!

### Help us help you

Yes, but only if others can assist

YandyZaldivar:
**Closes #398**

Adds Neon support to Aspire Community Toolkit with:
- A hosting integration for modeling and wiring Neon resources in AppHost.
- A client helper integration for consuming Neon connections in application projects.
- A one-shot provisioner workflow supporting attach/provision behavior and branch lifecycle scenarios.

### New packages
- `CommunityToolkit.Aspire.Hosting.Neon`
- `CommunityToolkit.Aspire.Neon`
- Provisioner project: `CommunityToolkit.Aspire.Neon.Provisioner`

### Hosting integration capabilities
- `AddNeon(...)` resource model with health integration.
- Provisioner modes:
  - `Attach` for existing resources.
  - `Provision` for create-or-attach workflows.
- Project and branch configuration:
  - project + branch targeting
  - ephemeral branches
  - branch restore
  - anonymized branch support / masking rules
- Database resource wiring via `AddDatabase(...)`.
- Dashboard compute commands (Suspend/Resume) via provisioner command flow.
- Startup ordering/health semantics for dependent resources (`WaitFor(neon)`).

### Client integration capabilities
- Convenience registration (`AddNeonClient(...)`) for consuming injected connection strings.

### Examples and tests
- Added Neon example app host.
- Added hosting and client test coverage for public API and functional behavior.

### Repo wiring updates
- Solution/package/workflow updates required to include the new Neon projects.

## Validation

- Focused Neon hosting tests:
  - `AddNeonTests`
  - `NeonFunctionalTests`

## PR Checklist

- [x] New integration
  - [x] Docs are written

### DIFF
--- a/src/CommunityToolkit.Aspire.Hosting.Neon/NeonProjectResource.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.Neon/NeonProjectResource.cs
@@ -87,6 +87,18 @@ public sealed class NeonProjectResource : Resource, IResourceWithConnectionStrin
     public IResourceWithWaitSupport? ProvisionerResource { get; internal set; }
 
     /// <summary>
+    /// Gets the host directory path where provisioner output files are written.
+    /// Used by consumer containers for bind mounts in run mode.
+    /// </summary>
+    public string? HostOutputDirectory { get; internal set; }
+
+    /// <summary>
+    /// Gets the Docker named volume used for sharing provisioner output in publish mode.
+    /// Consumers mount this volume to read per-database <c>.env</c> files at container startup.
+    /// </summary>
+    public string? OutputVolumeName { get; internal set; }
+
+    /// <summary>
     /// Gets the connection string expression for the Neon project.
     /// </summary>
     public ReferenceExpression ConnectionStringExpression =>
--- a/src/CommunityToolkit.Aspire.Neon.Provisioner/CommunityToolkit.Aspire.Neon.Provisioner.csproj
+++ b/src/CommunityToolkit.Aspire.Neon.Provisioner/CommunityToolkit.Aspire.Neon.Provisioner.csproj
@@ -7,6 +7,11 @@
     <Nullable>enable</Nullable>
     <DefineConstants>$(DefineConstants);NEON_PROVISIONER_ENTRYPOINT</DefineConstants>
     <IsPackable>false</IsPackable>
+    <!-- The provisioner writes to a Docker named volume whose root directory
+         is owned by root. .NET 8+ containers run as UID 1654 by default and
+         cannot write there, so we run the provisioner as root. This is safe
+         because the provisioner is a short-lived, one-shot process. -->
+    <ContainerUser>root</ContainerUser>
   </PropertyGroup>
 
 </Project>
--- a/src/CommunityToolkit.Aspire.Neon/AspireNeonExtensions.cs
+++ b/src/CommunityToolkit.Aspire.Neon/AspireNeonExtensions.cs
@@ -14,6 +14,95 @@ namespace Microsoft.Extensions.Hosting;
 public static class AspireNeonExtensions
 {
     private const string DefaultConfigSectionName = "Aspire:Neon:Client";
+    private const string NeonEnvFilePrefix = "NEON_ENV_FILE";
+    private const string NeonOutputDirKey = "NEON_OUTPUT_DIR";
+    private const string NeonConnectionUriKey = "NEON_CONNECTION_URI=";
+
+    /// <summary>
+    /// Reads Neon provisioner <c>.env</c> files from the shared output volume and
+    /// registers their connection URIs as standard <c>ConnectionStrings:{name}</c>
+    /// entries in the application configuration.
+    /// </summary>
+    /// <param name="builder">The <see cref="IHostApplicationBuilder"/> to configure.</param>
+    /// <returns>The same <paramref name="builder"/> for fluent chaining.</returns>
+    /// <remarks>
+    /// <para>
+    /// In publish mode (Docker Compose, ACA, etc.) the Neon provisioner container
+    /// writes per-database <c>.env</c> files to a shared Docker volume. Consumer
+    /// containers receive environment variables pointing to these files
+    /// (<c>NEON_ENV_FILE__{name}</c>) and the volume mount path
+    /// (<c>NEON_OUTPUT_DIR</c>).
+    /// </para>
+    /// <para>
+    /// Call this method early in <c>Program.cs</c> — before
+    /// <c>AddNpgsqlDataSource</c>, <c>AddDbContext</c>, or any other
+    /// client that reads <c>ConnectionStrings</c> — to make the
+    /// provisioned connection strings available through standard
+    /// configuration.
+    /// </para>
+    /// <para>
+    /// If you are using <see cref="AddNeonClient(IHostApplicationBuilder, string, Action{NeonClientSettings}?)"/> the env-file
+    /// resolution happens automatically; you do <b>not</b> need to call
+    /// this method separately.
+    /// </para>
+    /// <example>
+    /// <code lang="csharp">
+    /// var builder = WebApplication.CreateBuilder(args);
+    /// builder.AddNeonConnectionStrings();
+    /// builder.AddNpgsqlDataSource("appdb"); // connection string is now resolved
+    /// </code>
+    /// </example>
+    /// </remarks>
+    public static IHostApplicationBuilder AddNeonConnectionStrings(
+        this IHostApplicationBuilder builder)
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+
+        var envFileSection = builder.Configuration.GetSection(NeonEnvFilePrefix);
+        foreach (var child in envFileSection.GetChildren())
+        {
+            string connectionName = child.Key;
+            string? filePath = child.Value;
+
+            if (string.IsNullOrEmpty(filePath) || !File.Exists(filePath))
+            {
+                continue;
+            }
+
+            if (builder.Configuration.GetConnectionString(connectionName) is not null)
+            {
+                continue;
+            }
+
+            string? connectionUri = ParseNeonConnectionUri(filePath);
+            if (connectionUri is not null)
+            {
+                builder.Configuration[$"ConnectionStrings:{connectionName}"] = connectionUri;
+            }
+        }
+
+        string? outputDir = builder.Configuration[NeonOutputDirKey];
+        if (outputDir is not null && Directory.Exists(outputDir))
+        {
+            foreach (string envFile in Directory.GetFiles(outputDir, "*.env"))
+            {
+                string dbName = Path.GetFileNameWithoutExtension(envFile);
+
+                if (builder.Configuration.GetConnectionString(dbName) is not null)
+                {
+                    continue;
+                }
+
+                string? connectionUri = ParseNeonConnectionUri(envFile);
+                if (connectionUri is not null)
+                {
+                    builder.Configuration[$"ConnectionStrings:{dbName}"] = connectionUri;
+                }
+            }
+        }
+
+        return builder;
+    }
 
     /// <summary>
     /// Registers <see cref="NpgsqlDataSource"/> for Neon in the services provided by the <paramref name="builder"/>.
@@ -66,6 +155,12 @@ public static class AspireNeonExtensions
             settings.ConnectionString = connectionString;
         }
 
+        if (string.IsNullOrWhiteSpace(settings.ConnectionString))
+        {
+            settings.ConnectionString = TryReadProvisionerConnectionString(
+                builder.Configuration, connectionName);
+        }
+
         configureSettings?.Invoke(settings);
 
         ConnectionStringValidation.ValidateConnectionString(settings.ConnectionString, connectionName, DefaultConfigSectionName, configurationSectionName);
@@ -90,5 +185,49 @@ public static class AspireNeonExtensions
                 tags: default,
                 timeout: settings.HealthCheckTimeout > 0 ? TimeSpan.FromMilliseconds(settings.HealthCheckTimeout.Value) : null));
         }
+    }
+
+    /// <summary>
+    /// Attempts to read a Neon connection string from the provisioner env file.
+    /// Checks <c>NEON_ENV_FILE__{connectionName}</c> first, then falls back to
+    /// <c>NEON_OUTPUT_DIR/{connectionName}.env</c>.
+    /// </summary>
+    private static string? TryReadProvisionerConnectionString(
+        IConfiguration configuration, string connectionName)
+    {
+        string? envFilePath = configuration[$"{NeonEnvFilePrefix}:{connectionName}"];
+
+        if (envFilePath is null)
+        {
+            string? outputDir = configuration[NeonOutputDirKey];
+            if (outputDir is not null)
+            {
+                envFilePath = Path.Combine(outputDir, $"{connectionName}.env");
+            }
+        }
+
+        if (envFilePath is null || !File.Exists(envFilePath))
+        {
+            return null;
+        }
+
+        return ParseNeonConnectionUri(envFilePath);
+    }
+
+    /// <summary>
+    /// Parses a provisioner <c>.env</c> file and returns the
+    /// <c>NEON_CONNECTION_URI</c> value, or <see langword="null"/>.
+    /// </summary>
+    private static string? ParseNeonConnectionUri(string envFilePath)
+    {
+        foreach (string line in File.ReadAllLines(envFilePath))
+        {
+            if (line.StartsWith(NeonConnectionUriKey, StringComparison.Ordinal))
+            {
+                return line[NeonConnectionUriKey.Length..].Trim();
+            }
+        }
+
+        return null;
     }
 }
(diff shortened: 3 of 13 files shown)


YandyZaldivar:
@dotnet-policy-service agree

src/CommunityToolkit.Aspire.Hosting.Neon/NeonProvisionerMode.cs
tommasodotNET, line 1: aspire has a convention about resources that could be existing on azure or needs to be provisioned on first run/deploy.
`AsExisting(string existingResourceName, string existingResourceGroupName)` should be preferred, see https://aspire.dev/integrations/cloud/azure/overview/#use-existing-azure-resources

src/CommunityToolkit.Aspire.Hosting.Neon/NeonProjectResource.cs
```diff
@@ -0,0 +1,155 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="NeonProjectResource"/> class.
+    /// </summary>
+    /// <param name="name">The name of the resource.</param>
+    /// <param name="apiKey">The Neon API key parameter.</param>
+    /// <param name="options">The Neon project options.</param>
+    public NeonProjectResource(
+        [ResourceName] string name,
+        ParameterResource apiKey,
+        NeonProjectOptions options
```
tommasodotNET: we should handle the configuration using standard aspire convention. example here:
https://github.com/dotnet/aspire/pull/14149/changes#diff-44409193ed6bbf4f8b764b817264eb7646a3e92cbb8af1bbc05f5bc54c22e990

YandyZaldivar:
> overall looks good, but i have doubts about the way provisioning is handled. it looks like it doesn’t follow aspire conventions. i’m not sure how much we want to be tied to core azure integrations, @aaronpowell

@tommasodotNET thanks for the review. I’ve pushed a second commit addressing the provisioning concerns — would appreciate your thoughts on whether this aligns better with the direction you had in mind.

In this update I removed `.AddProvisioner(mode)` and introduced `.AsExisting()` to align more closely with the standard Aspire pattern for existing resources. I also removed configuration from the project resource constructor to better follow Aspire conventions.

Previously, the default behavior was always attach (`AsExisting`) semantics (since Neon resources are intended to live in the cloud, not locally). Now, provisioning intent is inferred from usage (e.g, `WithProject` vs `AddProject`), and calling `AsExisting()` explicitly forces attach semantics by disabling creation flags.

That said, Neon is not modeled or provisioned as an Azure resource in this integration. The goal is not to tie it to Azure semantics, but to follow Aspire’s public conventions where they make sense.

One nuance is that Neon (as modeled here) is always an external cloud resource — it never lives locally or necessarily inside the deployment environment. Provisioning (attach or create) always runs because the integration validates the project/branch/database and surfaces connection information in both run and publish modes.

I specifically tested this by deploying to a VPS via docker-compose (https://github.com/davidfowl/aspire-ssh-deploy), and Neon is still provisioned even though it does not live on that host. The deployment target (Azure, Kubernetes, Docker, etc.) is independent from where Neon physically resides. For example, a user can deploy their services to Azure while selecting `NeonProjectOptions.RegionId = "aws-us-east-1"`.

So while the public API now approximates Aspire’s Azure-style `AsExisting()` convention, Neon itself is not an Azure-scoped resource — it is an external database service that can be used regardless of the infrastructure provider.

Also for clarity: the Neon Azure Native integration has been deprecated and reached end-of-life on January 31, 2026 (https://neon.com/docs/introduction/billing-azure-marketplace). This integration is not related to that model; it targets Neon directly via its API.

tommasodotNET:
@YandyZaldivar thanks for clarifying.

i get what you'r saying, which is why i didn't request changes but just posted my comments as such. not sure what the correction direction for this integration should be.

the fact that neon should always be provisioned and never run locally is not unheard of. if you microsoft foundry, that can only be deployed on azure by aspire or attached to a pre-deployed resource.

that being said, it looks like we are closer to what the average aspire experience is.
the one thing that (if i'm not mistaken) is still different is setting deployment options via `ConfigureInfrastructure`.

aaronpowell:
Looking at the provisioner part of it, it does make me rather apprehensive on the design, the fact that it does a `dotnet run` on the fly of stuff that's pulled from an embedded resource really doesn't seem right. What if `dotnet` isn't in `$PATH`? what if they don't have the right .NET SDK installed? Ultimately, why isn't it just part of the hosting integration?

YandyZaldivar:
Neon as an integration can be modeled in different ways. In fact, it could even make sense to support more than one integration approach for the same resource. That’s why I like to frame this PR as: "within the scope of this integration", and avoid generalizing Neon. I also considered alternatives such as using tooling like neon_local: https://github.com/neondatabase/neon_local, not having a publish mode provisioner, and a few others.

@tommasodotNET
I've updated the API to use ConfigureInfrastructure and removed constructs like .WithProjectOptions and .WithBranchOptions, along with a few other adjustments in the latest commit. Every design choice comes with trade-offs, but my goal is to simplify the initial experience and better align with existing patterns. I'm happy to keep iterating if you still see room for improvement.

@aaronpowell
The provisioner is intentionally modeled as a standard AddProject(...) resource so it participates fully in the Aspire lifecycle: dashboard visibility, logs, health status, start/stop controls, and clear signaling when provisioning has completed. Under the hood this means dotnet run, but that is the same mechanism Aspire uses for any project added to a distributed application.

Regarding the .NET SDK and dotnet in $PATH, since Aspire itself runs via the .NET SDK, this integration doesn't introduce any additional requirement beyond what is already needed to run the AppHost locally.

The provisioner project is embedded and materialized on disk because it's an internal implementation detail of the integration. The intent is to avoid requiring consumers to manage it as a ProjectReference, while still allowing Aspire to treat it as a first-class project resource.

In publish mode, the provisioner is emitted as other projects, so in production it behaves like any other containerized resource. The dotnet run path is strictly for local development and dashboard-driven suspend/resume operations.

Why not part of the hosting integration?

- The provisioner needs to appear as a visible resource in the Aspire dashboard (logs, health status, lifecycle controls) and clearly indicate when provisioning has completed.
- Neon API operations (project creation, branch provisioning, endpoint polling) can take an uncertain amount of time. Running them in a separate process avoids blocking and triggering errors in the AppHost startup pipeline.
- Modeling it as a project resource ensures it works consistently across current and future deployment targets.
- In publish mode, it is automatically containerized, requiring no additional configuration from the user.
- Suspend/resume operations triggered from the dashboard invoke the provisioner in a targeted mode, and this is limited to local development scenarios.

So the intended flow is:
- Development (dotnet run): the provisioner runs as a standard Aspire project resource.
- Production (dotnet publish): the provisioner is emitted likely as a container and runs as a normal containerized resource.

Happy to discuss further a better design for the provisioner.

aaronpowell:
Shipping as a series of embedded resources that then have to be hydrated back to disk and then pushed into Aspire is a very inefficient model when you just want something to be modelled as a resource within Aspire.

> The provisioner needs to appear as a visible resource in the Aspire dashboard (logs, health status, lifecycle controls) and clearly indicate when provisioning has completed.

Yes, so `AddProvisioner` would create a `NeonProvisionerResource` which inherits from something like `Resource` since you're manually managing the life cycle

> Neon API operations (project creation, branch provisioning, endpoint polling) can take an uncertain amount of time. Running them in a separate process avoids blocking and triggering errors in the AppHost startup pipeline.

Resources don't block the main thread of Aspire. Take a look at the Ollama implementation, we have sub-resources that pull images and those are async operations that could take a long time, and it won't block Aspire, only resources that depend on it.

> Modeling it as a project resource ensures it works consistently across current and future deployment targets.

Aspire has teased shipping non-.NET app hosts, so that's probably not a safe assumption to make. Modelling using resources ensures consistency, not dynamically extracting files from an assuembly.

YandyZaldivar:
I think there are two valid goals we are trying to satisfy here:

1. The hosting integration should own the provisioning logic and model neon avoiding embedded project extraction.

2. There should still be a way to deploy a physical provisioner process/container to target environments when restart-driven reconciliation or self-healing behavior is required.

To reconcile these we can:

- Move the provisioning logic into a shared library.
- Refactor the hosting integration to execute provisioning logic directly (no embedded project hydration by default).
- Local development and dashboard provisioning would continue to work through the resource model.
- On deploy, provisioning would execute as part of aspire deploy or CI/CD workflows. 

For advanced scenarios, we could still allow an opt-in mode that emits a physical provisioner container. This would enable restart-driven reconciliation (e.g branch restore, use ephemeral branch, self-healing workflows) without making that the default behavior. For example:

```csharp
var neon = builder.AddNeon(apiKey)...
neon.WithRuntimeProvisioner(); // Explicit call to trigger .AddProject(provisioner)
```

This would:
- Align better with resource modeling.
- Remove the embedded project extraction pattern by deafult.
- Strengthen the hosting integration responsibility.
- Preserve the ability to run provisioning logic as a standalone container when operational requirements justify it.

Let me know if this direction makes more sense or if you have other advice.

src/CommunityToolkit.Aspire.Neon/AspireNeonExtensions.cs
```diff
@@ -0,0 +1,233 @@
...
+    /// </para>
+    /// <example>
+    /// <code lang="csharp">
+    /// var builder = WebApplication.CreateBuilder(args);
+    /// builder.AddNeonConnectionStrings();
+    /// builder.AddNpgsqlDataSource("appdb"); // connection string is now resolved
+    /// </code>
+    /// </example>
+    /// </remarks>
+    public static IHostApplicationBuilder AddNeonConnectionStrings(
```
eerhardt, line 56: This might be a silly question, but why wouldn't the client applications simply use the Npgsql client integration libraries? Why do you need a Neon specific aspire client integration?

For example, Valkey, Garnet, and Redis all use the same client integration for StackExchange.Redis.

src/CommunityToolkit.Aspire.Hosting.Neon/NeonBuilderExtensions.cs
```diff
@@ -0,0 +1,607 @@
...
+                $"Unable to run provisioner operation because project path '{annotation.ProjectPath}' was not found.");
+        }
+
+        var apiKey = await resource.ApiKeyParameter.GetValueAsync(cancellationToken).ConfigureAwait(false);
+        if (string.IsNullOrWhiteSpace(apiKey))
+        {
+            throw new DistributedApplicationException("Neon API key is required but was not provided.");
+        }
+
+        var processStartInfo = new ProcessStartInfo("dotnet", $"run --project \"{annotation.ProjectPath}\"")
```
eerhardt, line 579: Why does this need to be in a separate process? Why can't you just run the command in the current process?

src/CommunityToolkit.Aspire.Neon/AspireNeonExtensions.cs
YandyZaldivar: If the runtime provisioner mode proposed in this PR gets accepted, then the connection string is unknown at deployment time. What the provisioner has is a neon api-key, as part of the provisioner execution, the connection string, including the actual database password, is resolved. Then `AddNeonConnectionStrings` will inject the resolved connection strings.

Documented here [src/CommunityToolkit.Aspire.Neon/README.md](https://github.com/CommunityToolkit/Aspire/pull/1136/changes#diff-b5d0813fac05f83224cd9c25725aae98144e41775315b322255ec26456faf67d) 

Other than that, as you mention, and when using aspire run, the connection string can be read by any other conventional method, or the existing postgres client integration. However, it may still make sense to offer this package for convenience.

src/CommunityToolkit.Aspire.Hosting.Neon/NeonBuilderExtensions.cs
YandyZaldivar: In the current design, suspend/resume is intentionally routed through the external provisioner project because the provisioner is the single component that talks to the Neon API. So when a user runs Suspend/Resume from the Aspire dashboard, we invoke that provisioner entrypoint (mode=suspend|resume) rather than duplicating Neon API logic in hosting.

This was chosen to keep one Neon API execution surface and one environment contract across attach/provision/suspend/resume flows.

We are evaluating a follow up refactor where provisioning logic moves into a shared core and runtime provisioner becomes optional. In that model, hosting could execute suspend/resume in-process by default, with runtime provisioner remaining opt-in.

src/CommunityToolkit.Aspire.Neon.Provisioner/CommunityToolkit.Aspire.Neon.Provisioner.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
eerhardt, line 1: I think I'd do this a different way if it were me.

Instead of generating a (basically static) .NET project. If you want this to be a resource in the Aspire app, why don't you create a docker image with this functionality? Then the Neon Hosting library would just `.AddContainer("provisioner", "neon-provisioner:latest")`, or similar. Then you wouldn't need to generate and build this project everywhere.

Then to reuse the logic in the Hosting package, have a library that is shared between the Hosting integration and the container.

YandyZaldivar: Yes, but then who hosts the image, who maintains and keeps it sync? 

Ideally, aspire supports .AddProject, .AddContainer or something similar accepting a nuget package instead of having to copy sources.

eerhardt: > but then who hosts the image

https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

> who maintains and keeps it sync?

I assume you do? Are you going to be the maintainer of the Neon integration?

eerhardt: Note that in a [polyglot apphost](https://github.com/dotnet/aspire/blob/main/docs/specs/polyglot-apphost.md), if someone wants to use the Neon integration, now they need to install .NET on their machine, when they don't for other integrations because the code is already compiled.

YandyZaldivar: The idea of contributing to this repository is that the community is the maintainer, of course that I will happily assist.

About your docker image proposal I think it is too risky, that is creating a hard dependency, and we can't compare this to images like keycloak or ollama that have a whole community behind instead of a random guy.

What do you think of the idea of creating the concept of a nuget package executable resource in aspire? 

This way we can distribute the neon provisioner as a package as part of this repository. This will play nice with the polyglot apphost and will avoid the image dependency. For example, in aspire we can add the following to the KnownResourceTypes:
`public const string PackageExecutable = "PackageExecutable";` 

```csharp
internal static class KnownResourceTypes
{
    public const string Executable = "Executable";
    public const string ContainerExec = "ContainerExec";
    public const string Project = "Project";
    public const string Tool = "Tool";
    public const string Container = "Container";
    public const string Parameter = "Parameter";
    public const string ConnectionString = "ConnectionString";
    public const string ExternalService = "ExternalService";
}

```

I can draft a PR to discuss details.

PR #1241, opened 2026-03-27
Logto

axies20:
**Closes #<ISSUE_NUMBER>**

## PR Checklist

- [X] New integration
  - [X] Docs are written
  - [X] Added description of major feature to project description for NuGet package (4000 total character limit, so don't push entire description over that)


examples/logto/CommunityToolkit.Aspire.Hosting.Logto.AppHost/CommunityToolkit.Aspire.Hosting.Logto.AppHost.csproj
```diff
@@ -0,0 +1,22 @@
+<Project Sdk="Aspire.AppHost.Sdk/13.0.0">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net10.0</TargetFramework>
```
aaronpowell:
```suggestion
```

TFM managed above

examples/logto/CommunityToolkit.Aspire.Hosting.Logto.ClientJWT/Program.cs
```diff
@@ -0,0 +1,34 @@
+using CommunityToolkit.Aspire.Hosting.Logto.Client;
+using Microsoft.AspNetCore.Authentication.JwtBearer;
+using Microsoft.AspNetCore.Authorization;
+
+var builder = WebApplication.CreateBuilder(args);
+builder.AddServiceDefaults();
+
+builder.Services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
+    .AddLogtoJwtBearer("logto", "http://localhost:5072/",
```
aaronpowell: Shouldn't the endpoint come from Aspire?

examples/logto/CommunityToolkit.Aspire.Logto.ClientOIDC/Program.cs
```diff
@@ -0,0 +1,76 @@
...
+using Microsoft.AspNetCore.Authorization;
+using System.Security.Claims;
+
+var builder = WebApplication.CreateBuilder(args);
+builder.AddServiceDefaults();
+
+builder.AddLogtoOIDC("logto", logtoOptions: config =>
+{
+    config.AppId = "s6zda5bqn1qlsjzaiklqn";
+    config.AppSecret = "Df77aDt13MG3nSTgo8eKZP2HdeSfbed0";
```
aaronpowell: I assume these are things that could come from Aspire

src/CommunityToolkit.Aspire.Hosting.Logto.Client/CommunityToolkit.Aspire.Hosting.Logto.Client.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net10.0</TargetFramework>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
```
aaronpowell:
```suggestion
```

Comes from parent props file

src/CommunityToolkit.Aspire.Hosting.Logto/CommunityToolkit.Aspire.Hosting.Logto.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  
+  <PropertyGroup>
+    <Description>.NET Aspire hosting extensions for Logto (includes PostgreSQL and Redis integration).</Description>
+    <AdditionalPackageTags>logto redis postgres hosting extensions</AdditionalPackageTags>
+    <IsPreview>true</IsPreview>
```
aaronpowell: Why are we marking this as preview? Should we be putting `Experimental` attributes in?

src/CommunityToolkit.Aspire.Hosting.Logto/LogtoBuilderExtensions.cs
```diff
@@ -0,0 +1,222 @@
...
+            .WithImageRegistry(LogtoContainerTags.Registry);
+
+        builderWithResource.WithResourcePort();
+        builderWithResource.WithDatabase(postgres);
+        SetHealthCheck(builder, builderWithResource, name);
+
+
+        builderWithResource
+            .WithEntrypoint("sh")
+            .WithArgs("-c", "npm run cli db seed -- --swe && npm start");
```
aaronpowell: it could be very opaque to the user that the entrypoint is being overridden, what's the reason we do it?

Directory.Packages.props
```diff
@@ -20,7 +20,16 @@
...
+<<<<<<< Logto
+    <PackageVersion Include="Logto.AspNetCore.Authentication" Version="0.2.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Authentication" Version="2.3.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="10.0.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Authentication.OpenIdConnect" Version="10.0.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Http.Abstractions" Version="2.3.0" />
+    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="10.0.0" />
+=======
     <PackageVersion Include="AspNetCore.HealthChecks.Network" Version="9.0.0" />
+>>>>>>> main
```
aaronpowell: Think this is a merge gone bad

aaronpowell (changes requested):
Some questions and comments but the main thing is that the client project shouldn't have `Hosting` in the name (or namespace).

examples/logto/CommunityToolkit.Aspire.Logto.ClientOIDC/Program.cs
axies20: Not directly from Aspire today. These values are Logto client credentials, so they need to come from a Logto application registration first. Aspire can pass them through configuration/secrets, but it doesn't create the Logto application automatically.

It would be possible to automate that through the Logto Management API, but that would add a separate provisioning workflow, which felt out of scope for this sample.

src/CommunityToolkit.Aspire.Hosting.Logto/LogtoBuilderExtensions.cs
axies20: We override the entrypoint because Logto requires the DB seed step before the app starts in our Aspire scenario, otherwise the container doesn't come up in a usable state. I agree this is not obvious from AddLogtoContainer(), so I can make it explicit either by documenting it clearly or by moving it behind an opt-in method like `WithDatabaseSeeding()/WithSeededStartup()`.

examples/logto/CommunityToolkit.Aspire.Hosting.Logto.ClientJWT/Program.cs
axies20: It does. `logto` is the Aspire resource name, so the endpoint is resolved from Aspire via `.WithReference(logto)`. The second parameter is the JWT `appIdentification`/audience, not the Logto endpoint, so I updated the sample to make that clearer.

aaronpowell (approved):

aaronpowell (changes requested):
Need to update the test list to include the new test projects

axies20:
Something wrong? Can it be reopened?

axies20:
Can you please reopen and check?

src/CommunityToolkit.Aspire.Hosting.Logto/LogtoBuilderExtensions.cs
aaronpowell: I think we should move it to a method like `WithDatabaseSeeding` (or `WithSeededStartup` - naming things is hard and I'm not fussed either way) just so that we make it obvious.

That is, I assume that Logto will run without having to run this seed right?

src/CommunityToolkit.Aspire.Hosting.Logto/LogtoTags.cs
```diff
@@ -0,0 +1,16 @@
...
+public class LogtoTags
+{
+    /// <summary>docker.io</summary>
+    public const string Registry = "docker.io";
+
+    /// <summary>svhd/logto</summary>
+    public const string Image = "svhd/logto";
+
+    /// <summary>1.38</summary>
+    public const string Tag = "1.38";
```
aaronpowell:
```suggestion
    public const string Tag = "1.39";
```

That looks like latest since the PR has been delayed.

aaronpowell:
Need to bump the SDK version on the example to match the rest of repo - that'd be why the tests are failing

examples/logto/CommunityToolkit.Aspire.Hosting.Logto.AppHost/CommunityToolkit.Aspire.Hosting.Logto.AppHost.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Aspire.AppHost.Sdk/13.2.0">
```
aaronpowell:
```suggestion
<Project Sdk="Aspire.AppHost.Sdk/13.3.0">
```

aaronpowell:
Updated the SDK version but now we're seeing legit failures on the test run - it looks like the backend service might not be starting on the expected port (tests report `Connection refused (localhost:3001)`).

Also, Copilot has left a stack of comments, can you do an audit on them to see what is accurate and needs addressing.

src/CommunityToolkit.Aspire.Hosting.Logto/CommunityToolkit.Aspire.Hosting.Logto.csproj
```diff
@@ -0,0 +1,15 @@
...
+  <PropertyGroup>
+    <Description>.NET Aspire hosting extensions for Logto (includes PostgreSQL and Redis integration).</Description>
+    <AdditionalPackageTags>logto redis postgres hosting extensions</AdditionalPackageTags>
+  </PropertyGroup>
+  
+    <ItemGroup>
+      <PackageReference Include="Aspire.Hosting" />
+      <PackageReference Include="Aspire.Hosting.PostgreSQL" />
+      <PackageReference Include="Aspire.Hosting.Redis" />
+    </ItemGroup>
```
aaronpowell, line 14:
```suggestion
      <PackageReference Include="MessagePack" />
    </ItemGroup>
```

aaronpowell (approved):

Issue #1232, opened 2026-03-24
Add Aspire.Hosting.Kind resource for local Kubernetes cluster provisioning


tamirdresher:
## Summary

I'd like to propose adding an `Aspire.Hosting.Kind` integration resource that enables provisioning and managing local Kubernetes clusters using [Kind (Kubernetes IN Docker)](https://kind.sigs.k8s.io/) from within .NET Aspire application host code.

## Motivation

Kind is the standard tool for running local Kubernetes clusters for development and CI/CD pipelines. Aspire already has excellent integrations for containers and cloud services, but there's currently no first-class way to spin up a local Kubernetes cluster as part of an Aspire app host. This gap means developers who want to test their apps on Kubernetes locally have to set up and manage Kind clusters manually, outside the Aspire lifecycle.

An `Aspire.Hosting.Kind` resource would:
- Let developers declare a local k8s cluster as part of their Aspire app host
- Automatically create/delete the Kind cluster with the app lifecycle
- Support deploying Kubernetes manifests (Deployments, Services, ConfigMaps, etc.) onto the cluster
- Integrate with Aspire's health checks, dashboard, and resource model
- Simplify local k8s development for teams already using Aspire

## Proposed API

```csharp
var builder = DistributedApplication.CreateBuilder(args);

var cluster = builder.AddKindCluster("my-cluster")
    .WithKubernetesVersion("v1.30.0");

cluster.AddDeployment("api-deployment", "manifests/api-deployment.yaml");
cluster.AddService("api-service", "manifests/api-service.yaml");

builder.Build().Run();
```

## Implementation

A working implementation already exists in the [tamirdresher/aspire-kind](https://github.com/tamirdresher/aspire-kind) repository. It includes:

- `KindClusterResource` — core Aspire resource implementing `IResourceWithConnectionString`
- `KindClusterLifecycleHook` — `IDistributedApplicationLifecycleHook` that creates/destroys the cluster
- `KindClusterExtensions` — fluent API extension methods (`AddKindCluster`, `AddDeployment`, `AddService`, etc.)
- `KubernetesManifestResource` — child resource representing k8s manifests deployed to the cluster
- Full integration tests using `DistributedApplicationTestingBuilder`
- XML documentation and usage samples

The implementation uses the `kind` CLI (must be installed) and wraps it via `CliWrap`. It was tested on Linux and Windows.

## Prior Art / Related

[...]

PR #1270, opened 2026-04-09
Add CommunityToolkit.Aspire.Hosting.Kind integration

andrey-noskov:
## What this PR is about
This PR adds `CommunityToolkit.Aspire.Hosting.Kind`, a hosting integration for running real local Kubernetes clusters with Kind (Kubernetes in Docker) as part of an Aspire app host.

The goal is to make local development and validation much closer to a real Kubernetes environment. This is useful both for Kubernetes-specific components such as operators, plugins, controllers, and cluster-level services, and for regular application workloads that need to be exercised in a Kubernetes environment before they reach shared or staging clusters.

## What is included
- Kind cluster resource and lifecycle support via `AddKindCluster(...)`
- Helm chart deployment support for Kind-managed clusters
- Aspire publish/deploy support targeting Kind
- example AppHost projects
- unit and integration test coverage
- required repo wiring for package metadata, solution files, root README, and test matrix

## Why this matters
Today, teams typically either manage Kind clusters manually outside Aspire or wait to validate Kubernetes behavior in a shared or remote environment. This integration keeps that workflow inside Aspire's resource model and shortens the feedback loop for local Kubernetes-oriented development.

## Context
- relates to #1232


andrey-noskov:
Contributors on this PR: @andrey-noskov @MattKotsenas

andrey-noskov:
@dotnet-policy-service agree company="Microsoft"

andrey-noskov:
@aaronpowell @davidfowl I'd appreciate your review and your perspective on whether this is the right shape for a CommunityToolkit Kind integration, especially around the local Kubernetes and publish/deploy workflow story.

tamirdresher:
@aaronpowell would appreciate your help promoting this

src/CommunityToolkit.Aspire.Hosting.Kind/DefaultProcessRunner.cs
```diff
@@ -0,0 +1,91 @@
...
+            catch (InvalidOperationException)
+            {
+                // Process already exited.
+            }
+
+            throw;
+        }
+
+        return new ProcessResult(process.ExitCode, stdout.ToString().TrimEnd(), stderr.ToString().TrimEnd());
+    }
```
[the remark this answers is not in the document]
MattKotsenas: I don't think that's true anymore? I'm the one that accidentally caused that bug, let me check...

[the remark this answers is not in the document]
MattKotsenas: Yeah, fixed back in 2020: https://github.com/dotnet/runtime/pull/42585

tests/CommunityToolkit.Aspire.Hosting.Kind.Tests/RequiresKindAttribute.cs
```diff
@@ -0,0 +1,59 @@
...
+
+            using var process = Process.Start(psi);
+            if (process is null)
+            {
+                return false;
+            }
+
+            process.WaitForExit(TimeSpan.FromSeconds(5));
+            return process.ExitCode == 0;
+        }
```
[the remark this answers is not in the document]
MattKotsenas: We actually shouldn't be using Process.Start directly at all here, we should be using our IProcessRunner.

src/CommunityToolkit.Aspire.Hosting.Kind/KindClusterManager.cs
```diff
@@ -0,0 +1,164 @@
...
+            _logger.LogInformation("Creating Kind cluster '{ClusterName}'...", _resource.Name);
+            var result = await _processRunner.RunAsync(
+                _logger,
+                "kind",
+                [
+                    "create",
+                    "cluster",
+                    $"--name={_resource.Name}",
+                    $"--config={configPath}",
+                    $"--kubeconfig={_resource.KubeconfigPath}",
```
MattKotsenas, line 67: Do we need quoting here (and similar places) for paths with spaces?

src/CommunityToolkit.Aspire.Hosting.Kind/KindClusterResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,261 @@
...
+            if (result.ExitCode != 0)
+            {
+                throw new InvalidOperationException(
+                    "Kind CLI not found. Install it from https://kind.sigs.k8s.io/docs/user/quick-start/#installation");
+            }
+        }
+        catch (Win32Exception)
+        {
+            throw new InvalidOperationException(
+                "Kind CLI not found. Install it from https://kind.sigs.k8s.io/docs/user/quick-start/#installation");
```
MattKotsenas: Can / should we provide the `Win32Exception` as the inner exception in case it has more details?

src/CommunityToolkit.Aspire.Hosting.Kind/KindConfigModels.cs
```diff
@@ -0,0 +1,216 @@
...
+
+    /// <summary>
+    /// Gets or sets the resource kind.
+    /// </summary>
+    public string Kind { get; set; } = "Cluster";
+
+    /// <summary>
+    /// Gets the list of nodes in the cluster.
+    /// </summary>
+    public List<KindNodeModel> Nodes { get; } = [];
```
MattKotsenas: While old, it's probably still a good idea to not use concrete collection types in public APIs -- https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections

Also applies to Dictionary down below.

src/CommunityToolkit.Aspire.Hosting.Kind/KindContainerExtensions.cs
```diff
@@ -0,0 +1,209 @@
...
+            });
+
+        return builder;
+    }
+
+    /// <summary>
+    /// Connects the container to the "kind" Docker network if not already connected.
+    /// </summary>
+    /// <returns><see langword="true"/> if the connection was made (or was already present); <see langword="false"/> on failure.</returns>
+    private static async Task<bool> ConnectToKindNetworkAsync(
```
MattKotsenas: nit: Might make more sense to name this something like `EnsureConnectedToKindNetworkAsync`

src/CommunityToolkit.Aspire.Hosting.Kind/KindContainerExtensions.cs
```diff
@@ -0,0 +1,209 @@
...
+            .Single()
+            .Name;
+
+        var loggerService = services.GetRequiredService<ResourceLoggerService>();
+        var logger = loggerService.GetLogger(containerName);
+        var processRunner = services.GetRequiredService<IProcessRunner>();
+
+        var connectResult = await processRunner.RunAsync(
+            logger,
+            "docker",
```
MattKotsenas: Flagging for discussion what we should do about podman support? Generally, Aspire supports both. I think if someone is running podman in root mode, everything here should work OK? Rootless would be a different story though.

src/CommunityToolkit.Aspire.Hosting.Kind/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+| From \\ To | Host Process | Aspire-managed container | Kind control plane | Kind-managed workload |
+|---|---|---|---|---|
+| **Host Process** | Yes | Yes | Yes — Kind publishes the API to host `127.0.0.1:<port>`; `WithReference(kind)` injects kubeconfig | Yes, via `kubectl port-forward` or NodePort on localhost |
+| **Aspire-managed container** | Yes | Yes | Yes — `WithReference(kind)` rewrites the kubeconfig and `WithKindNetwork()` joins the Docker `kind` network | Yes — requires `WithKindNetwork()` plus a Kubernetes exposure mechanism (NodePort, HostPort) |
+| **Kind control plane** | No by default | No by default — the control plane has no route to Aspire's Docker network | Yes | Yes |
+| **Kind-managed workload** | No by default | No by default — pods run in an isolated network namespace inside the Kind node container | Yes | Yes |
+
+> `WithKindNetwork()` connects the Aspire container to the Docker `kind` network, giving it L3 connectivity to the Kind node containers. It does **not** grant access to the Kubernetes pod or service networks — workloads inside the cluster must be exposed via NodePort, HostPort, or a similar mechanism.
```
MattKotsenas: This I think has a missing `[!NOTE]` or `[!TIP]` at the top

src/CommunityToolkit.Aspire.Hosting.Kind/README.md
```diff
@@ -0,0 +1,243 @@
...
+    .WithReference(cluster);
+```
+
+The container-specific `WithReference` overload automatically:
+- Bind-mounts the Kind kubeconfig into the container at `/etc/kubeconfig/config`
+- Sets `KUBECONFIG` to the in-container mount path
+- Sets `K8S_CLUSTER_NAME` to the Kind cluster name
+- Connects the container to the Kind Docker network
+
+> **Note:** The kubeconfig mounted into containers uses the Kind control-plane container name (e.g., `mycluster-control-plane:6443`) instead of `127.0.0.1`, enabling container-to-container communication over the Kind Docker network.
```
MattKotsenas: same here about note

src/CommunityToolkit.Aspire.Hosting.Kind/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+```csharp
+var cluster = builder.AddKindCluster("mycluster");
+
+var worker = builder.AddContainer("my-worker", "myregistry/my-worker")
+    .WithReference(cluster)
+    .WithKindNetwork();
+```
+
+> **Note:** `WithKindNetwork` is available on `IResourceBuilder<ContainerResource>`. It connects the container to the `kind` Docker network automatically when the container starts.
```
MattKotsenas: same here

tests/CommunityToolkit.Aspire.Hosting.Kind.Tests/AddKindClusterTests.cs
```diff
@@ -0,0 +1,574 @@
...
+            var yaml = await File.ReadAllTextAsync(configPath);
+            Assert.Contains($"image: {"kindest/node"}:v1.31.0", yaml);
+        }
+        finally
+        {
+            File.Delete(configPath);
+        }
+    }
+
+    // ── DefaultProcessRunner tests ──────────────────────────────────────
```
MattKotsenas: looks left over?

tests/CommunityToolkit.Aspire.Hosting.Kind.Tests/CommunityToolkit.Aspire.Hosting.Kind.Tests.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <IsPackable>false</IsPackable>
+    <IsTestProject>true</IsTestProject>
+    <!-- Verify.XunitV3 brings in newer xunit.v3.common/extensibility.core transitively.
```
MattKotsenas, line 6: Flagging this for discussion. I think the "right" thing to do would be to bump the Verify version, but that would require bumping the xunit version for the repo

CommunityToolkit.Aspire.Hosting.Kind.slnx
```diff
@@ -0,0 +1,12 @@
+<Solution>
```
MattKotsenas: I think this is in the wrong directory

examples/kind/CommunityToolkit.Aspire.Hosting.Kind.AppHost/CommunityToolkit.Aspire.Hosting.Kind.AppHost.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Aspire.AppHost.Sdk/13.2.0">
```
aaronpowell: SDK needs bumping

examples/kind/CommunityToolkit.Aspire.Hosting.Kind.Publish.AppHost/CommunityToolkit.Aspire.Hosting.Kind.Publish.AppHost.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Aspire.AppHost.Sdk/13.2.0">
```
aaronpowell: SDK needs updating

src/CommunityToolkit.Aspire.Hosting.Kind/KindContainerExtensions.cs
```diff
@@ -0,0 +1,209 @@
...
+        var containerName = resource.Annotations
+            .OfType<ContainerNameAnnotation>()
+            .Single()
+            .Name;
+
+        var loggerService = services.GetRequiredService<ResourceLoggerService>();
+        var logger = loggerService.GetLogger(containerName);
+        var processRunner = services.GetRequiredService<IProcessRunner>();
+
+        var connectResult = await processRunner.RunAsync(
```
davidfowl, line 163: Broken with podman?

andrey-noskov: I am open to suggestions

CommunityToolkit.Aspire.slnx
```diff
@@ -70,6 +70,9 @@
     <Project Path="examples\keycloak-postgres\CommunityToolkit.Aspire.Hosting.Keycloak.Extensions.Prod\CommunityToolkit.Aspire.Hosting.Keycloak.Extensions.Prod.csproj" />
     <Project Path="examples\keycloak-postgres\CommunityToolkit.Aspire.Hosting.Keycloak.Extensions.ServiceDefaults\CommunityToolkit.Aspire.Hosting.Keycloak.Extensions.ServiceDefaults.csproj" />
   </Folder>
+  <Folder Name="/examples/kind/">
+    <Project Path="examples/kind/CommunityToolkit.Aspire.Hosting.Kind.AppHost/CommunityToolkit.Aspire.Hosting.Kind.AppHost.csproj" />
```
mitchdenny, line 85: **Blocker — solution files inconsistent.** The main solution only adds `examples/kind/CommunityToolkit.Aspire.Hosting.Kind.AppHost`, but the focused `CommunityToolkit.Aspire.Hosting.Kind.slnx` adds **both** `...AppHost` and `...Publish.AppHost`. Please also add `examples/kind/CommunityToolkit.Aspire.Hosting.Kind.Publish.AppHost/CommunityToolkit.Aspire.Hosting.Kind.Publish.AppHost.csproj` here so the publish/deploy example builds as part of the main solution and gets covered by repo-wide build/format checks.

mitchdenny (commented):
Flagging the three blockers I'd like resolved before this merges. Full review notes shared separately — these inline comments are the must-fix-before-merge subset.

mitchdenny:
**Correction to my earlier review.** I just re-verified against the PR's actual diff (via `gh api ... /files`) and two of the three "blockers" I called out were mistaken — they came from me diffing the branch against current `upstream/main` rather than against the PR's merge-base. Specifically:

- The `Directory.Packages.props` "downgrades" of `Aspire.Hosting.Elasticsearch`, `Azure.Provisioning.AppContainers`, `coverlet.collector`, `MartinCostello.Logging.XUnit.v3`, `GitHubActionsTestLogger`, `OpenFeature.Providers.Flagd`, and the `OpenTelemetry.*` switch to `$(OpenTelemetryVersion)` are **not in this PR**. They're upstream changes that landed on `main` after this branch was created. A rebase will pick them up automatically — no action needed.
- The `continue-on-error: true` removal from the **Upload bin log artifact** step in `.github/workflows/tests.yaml` is **also not in this PR**. The PR's only change to that workflow is adding `Hosting.Kind.Tests,` to the matrix, which is correct.

Apologies for the noise. The remaining blocker — the inline `CommunityToolkit.Aspire.slnx` comment about adding the Publish.AppHost example — stands.

The rest of the review (architectural/alignment items, especially `WithKind()` shape, helm-runner sharing with [microsoft/aspire#17425](https://github.com/microsoft/aspire/issues/17425), the `WithKindNetwork` restart loop, cluster health check, and the temp-path collision) is unchanged.

aaronpowell (commented):
Need to bump the SDK on projects.

src/CommunityToolkit.Aspire.Hosting.Kind/KindContainerExtensions.cs
tamirdresher: @andrey-noskov  @davidfowl 
I think the Kind integration should avoid assuming Docker specifically and instead align with Aspire’s container runtime behavior.

Aspire already supports both Docker and Podman, with `ASPIRE_CONTAINER_RUNTIME` / `DOTNET_ASPIRE_CONTAINER_RUNTIME` as advanced overrides and runtime probing when no override is set. For this package, that means the Kind hosting layer should:

1. Resolve the selected container runtime the same way Aspire does: Docker when available, Podman when Docker is unavailable, with explicit env-var overrides respected.
2. Use the selected runtime for direct container operations like network connection instead of hardcoding `docker`.
3. When Podman is selected, set `KIND_EXPERIMENTAL_PROVIDER=podman` internally for `kind` CLI calls, rather than requiring users to know or configure that manually.
4. Keep `kind load docker-image` as-is, since that is Kind’s CLI terminology even when the provider is Podman.

Kind documents Podman support through its provider env var here:
https://kind.sigs.k8s.io/docs/user/rootless/

I have a patch for this locally that adds the runtime resolver, wires the Podman provider env into Kind cluster/image-load operations, and adds tests for Docker/Podman selection behavior. If this sounds good, I can get the patch into the PR.

davidfowl: There’s a built in IContainerRuntimeResolver for this now.

tamirdresher: Even better!

examples/kind/CommunityToolkit.Aspire.Hosting.Kind.Publish.AppHost/CommunityToolkit.Aspire.Hosting.Kind.Publish.AppHost.csproj
andrey-noskov: done

examples/kind/CommunityToolkit.Aspire.Hosting.Kind.AppHost/CommunityToolkit.Aspire.Hosting.Kind.AppHost.csproj
andrey-noskov: done

examples/kind/CommunityToolkit.Aspire.Hosting.Kind.AppHost/CommunityToolkit.Aspire.Hosting.Kind.AppHost.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Aspire.AppHost.Sdk/13.3.0">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <IsAspireHost>true</IsAspireHost>
+  </PropertyGroup>
```
aaronpowell:
```suggestion
  </PropertyGroup>
```

examples/kind/CommunityToolkit.Aspire.Hosting.Kind.Publish.AppHost/CommunityToolkit.Aspire.Hosting.Kind.Publish.AppHost.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Aspire.AppHost.Sdk/13.3.0">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <IsAspireHost>true</IsAspireHost>
+  </PropertyGroup>
```
aaronpowell:
```suggestion
  </PropertyGroup>
```

tests/CommunityToolkit.Aspire.Hosting.Kind.Tests/AppHostTests.cs
```diff
@@ -0,0 +1,19 @@
...
+public class AppHostTests(AspireIntegrationTestFixture<Projects.CommunityToolkit_Aspire_Hosting_Kind_AppHost> fixture)
+    : IClassFixture<AspireIntegrationTestFixture<Projects.CommunityToolkit_Aspire_Hosting_Kind_AppHost>>
+{
+    [Fact]
+    public async Task KindClusterStartsAndBecomesHealthy()
+    {
+        await fixture.ResourceNotificationService
+            .WaitForResourceHealthyAsync("kind-cluster")
+            .WaitAsync(TimeSpan.FromMinutes(5));
+    }
```
aaronpowell, line 23: Is there something more explicit we can assert on in here rather than just assuming that there was no timeout exception means tests pass

CommunityToolkit.Aspire.Hosting.Kind.slnx
aaronpowell, line 1: Let's not add a new slnx file, it'll just cause pain

aaronpowell (changes requested):
Done a few inline nits on the PR but here's some more overarching comments:
* We try to use primary constructors on types as much as possible, and there's a number of classes that could be converted to them
* I'm not entirely sure we want to introduce Verify, esp if it's going to result in dependency bumping our testing framework. The testing infrastructure is already pretty thinly stretched and I'm hesitant to add more to it

examples/kind/CommunityToolkit.Aspire.Hosting.Kind.AppHost/CommunityToolkit.Aspire.Hosting.Kind.AppHost.csproj
```diff
@@ -0,0 +1,14 @@
...
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\src\CommunityToolkit.Aspire.Hosting.Kind\CommunityToolkit.Aspire.Hosting.Kind.csproj" IsAspireProjectResource="false" />
+  </ItemGroup>
+
```
aaronpowell, line 10:
```suggestion

  <ItemGroup>
    <PackageReference Include="MessagePack" />
  </ItemGroup>
```

examples/kind/CommunityToolkit.Aspire.Hosting.Kind.Publish.AppHost/CommunityToolkit.Aspire.Hosting.Kind.Publish.AppHost.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Aspire.AppHost.Sdk/13.3.0">
```
aaronpowell:
```suggestion
<Project Sdk="Aspire.AppHost.Sdk/13.4.3">
```

src/CommunityToolkit.Aspire.Hosting.Kind/CommunityToolkit.Aspire.Hosting.Kind.csproj
```diff
@@ -0,0 +1,19 @@
...
+    <PackageReference Include="Aspire.Hosting" />
+    <PackageReference Include="Aspire.Hosting.Kubernetes" />
+    <PackageReference Include="KubernetesClient" />
+    <PackageReference Include="YamlDotNet" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <InternalsVisibleTo Include="CommunityToolkit.Aspire.Hosting.Kind.Tests" />
+  </ItemGroup>
+
```
aaronpowell, line 18:
```suggestion

  <ItemGroup>
    <PackageReference Include="MessagePack" />
  </ItemGroup>
```

tests/CommunityToolkit.Aspire.Hosting.Kind.Tests/CommunityToolkit.Aspire.Hosting.Kind.Tests.csproj
```diff
@@ -0,0 +1,21 @@
...
+  <ItemGroup>
+    <PackageReference Include="Verify.XunitV3" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\examples\kind\CommunityToolkit.Aspire.Hosting.Kind.AppHost\CommunityToolkit.Aspire.Hosting.Kind.AppHost.csproj" />
+    <ProjectReference Include="..\..\src\CommunityToolkit.Aspire.Hosting.Kind\CommunityToolkit.Aspire.Hosting.Kind.csproj" />
+    <ProjectReference Include="..\CommunityToolkit.Aspire.Testing\CommunityToolkit.Aspire.Testing.csproj" />
+  </ItemGroup>
+
```
aaronpowell, line 21:
```suggestion

  <ItemGroup>
    <PackageReference Include="MessagePack" />
  </ItemGroup>
```

aaronpowell (commented):
Fixing some stuff to sync to main

davidfowl:
Why?

tamirdresher:
@andrey-noskov on purpose?

aaronpowell (approved):

PR #1294, opened 2026-04-27
feat: upgrade to support SurrealDB v3

Odonno:
Update SurrealDB integration to support v3:

* hosting container tag updated
* .NET SDK version updated

## PR Checklist

## Other information

* Projects failed to compile due to vulnerabilities on some OpenTelemetry package versions. Also weird OpenTelemetry versioning, needed to pin latest version available on NuGet.

### DIFF
--- a/.github/actions/setup-runtimes-caching/action.yml
+++ b/.github/actions/setup-runtimes-caching/action.yml
@@ -35,7 +35,7 @@ runs:
     - name: Install Aspire CLI
       uses: timheuer/setup-aspire@v0.1.0
       with:
-        quality: staging
+        version: ${{ steps.aspire-version.outputs.version }}
 
     - name: Verify Aspire CLI version
       shell: bash
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -97,6 +97,7 @@ jobs:
           docker-registry: ${{ secrets.CUSTOM_CONTAINER_REGISTRY }}
 
       - name: Verify Docker is running
+        if: runner.os == 'Linux'
         run: docker info
 
       - name: Build test project
@@ -126,7 +127,7 @@ jobs:
           CUSTOM_CONTAINER_REGISTRY: ${{ secrets.CUSTOM_CONTAINER_REGISTRY }}
 
       - name: Dump docker info
-        if: always()
+        if: always() && runner.os == 'Linux'
         run: |
           docker container ls --all
           docker container ls --all --format json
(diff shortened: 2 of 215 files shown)


aaronpowell (approved):

aaronpowell:
We've got some merge conflicts @Odonno which is because of the OTEL packages - if you can fix those, when CI goes green it should auto-merge

aaronpowell:
@Odonno it seems that the TypeScript AppHost is failing here. The logs are saying that the resource isn't starting, so you might need to run it locally and verify/resolve

Odonno:
For info, Aspire 13.3.0 is unusable on Arch Linux, I have to upgrade to 13.3.5.

Odonno:
Sometimes the tests works, sometimes they don't. I don't know why at this point.. Could that be related to an issue with Aspire core?

aaronpowell:
Test instability _I think_ is caused by just the sheer number of runners we start when we do a full test run. While this PR is focusing on the update of SurrealDB, there's also a change the ConformanceTests class in the shared testing library, which then triggers a full test rerun because that is used across most of our tests.

Ideally, that change should be spun off to a separate PR, but I've kicked the runners to go on a second pass.

Odonno:
Ah, that could explain the strange behavior where the test got triggered a second time before it even finished a first time? I had that behavior sometimes in Debug mode.

davidfowl:
I think this is mostly a law of large numbers problem. If any one test has even a 0.1% chance of flaking, that sounds fine in isolation, but once you run thousands of tests across lots of runners, those tiny probabilities compound quickly.

And it’s not just the test code. At that scale, we’re also multiplying exposure to network hiccups, disk contention, package feeds, NuGet, npm, Docker pulls, GitHub APIs, and any external service the tests depend on.

Rare failures stop being rare when the system is big enough.

Odonno:
I agree with that perspective but here the problem is that test failure feels like ~50% for a single test project, for a not so large number of tests (less than 100). Even in the local machine, this is too much IMO. But I know saying this is not helpful at all. I will try to dig more into it when I can.

aaronpowell:
Looks like some merge conflicts with the 13.4 update @Odonno

aaronpowell:
@Odonno looks like we have a legit test failure on the Surreal test suite

Odonno:
Not sure why, all tests are green locally.

aaronpowell:
Weird, re-ran the job and it passed. Will need to keep an eye on it.

PR #1322, opened 2026-05-14
feat: CommunityToolkit.Aspire.Hosting.K3s — k3s Kubernetes cluster hosting integration

### LINKED ISSUE #1321: feat: K3s Kubernetes cluster hosting integration
#### Is your feature request related to a problem? Please describe the problem.
​
Developing and testing Kubernetes operators, controllers, and workloads locally requires a real Kubernetes API server. Today, developers must manually provision a cluster (k3d, Kind, minikube, Docker Desktop Kubernetes) *before* running an Aspire app that includes such components. This breaks the "works on my machine in one `F5`" promise that Aspire delivers for databases, caches, and message brokers.
​
There is no first-class Aspire community resource that:
​
- Starts a local Kubernetes cluster automatically as part of the dev inner loop
- Lets other resources declare `WaitFor` dependencies on the cluster and receive kubeconfig automatically via `WithReference`
- Supports deploying Helm charts into the cluster before dependent services start
- Applies Kustomize overlays and plain YAML manifests with CRD readiness gating
- Exposes Kubernetes services as Aspire endpoint resources reachable from both host processes and DCP-network containers
- Requires no host-side tooling beyond a compatible container runtime (Docker or Podman)
​
#### Describe the solution you'd like
​
A `CommunityToolkit.Aspire.Hosting.K3s` package that models a k3s Kubernetes cluster as a first-class Aspire resource tree, with the following resource types:
​
| Resource | Type | Description |
|---|---|---|
| `K3sClusterResource` | `ContainerResource` | The k3s server (privileged `rancher/k3s` container) |
| `HelmReleaseResource` | `ContainerResource` | Installs a Helm chart via `alpine/helm`; exits 0 when done |
| `K8sManifestResource` | `ContainerResource` | Applies YAML manifests or a Kustomize overlay via `alpine/kubectl`; exits 0 when done |
| `K3sServiceEndpointResource` | `Resource` | In-process KubernetesClient WebSocket port-forward exposing a k8s service to the Aspire network |
​
#### Proposed API
​
```csharp
var cluster = builder.AddK3sCluster("k8s")
    .WithDataVolume()
    .WithK3sVersion("v1.36.0-k3s1");
​
// Helm chart — installs via alpine/helm container, no host helm required
var argocd = cluster.AddHelmRelease("argocd", "argo-cd",
    repo: "https://argoproj.github.io/argo-helm",
    version: "7.8.0",
    @namespace: "argocd")
    .WithHelmValuesFile("./deploy/argocd-values.yaml")
    .WithHelmValue("server.insecure", "true");
​
// YAML manifest / Kustomize overlay — applies via alpine/k8s container
var crd = cluster.AddK8sManifest("widget-crd", "./k8s/crds/");
​
// Service endpoint — in-process WebSocket port-forward, no NodePort required
var ui = cluster.AddServiceEndpoint("argocd-ui", "argocd-server", 443, "argocd")
    .WaitForCompletion(argocd);
​
// Host process: receives KUBECONFIG=/path/.k3s/k8s/local/kubeconfig.yaml
builder.AddProject<Projects.WidgetOperator>("operator")
    .WaitForCompletion(crd)
    .WithReference(cluster);
​
// Container: receives a bind-mount of the container/ kubeconfig directory 
// (server: https://{name}:6443) at /var/k3s/, and KUBECONFIG=/var/k3s/kubeconfig.yaml. 
// All standard Kubernetes tooling (kubectl, helm, KubernetesClient SDK) works without 
// any custom bootstrap code.
builder.AddContainer("operator", "myorg/operator")
    .WaitForCompletion(crd)
    .WithReference(cluster);
​
// Any consumer: receives services__argocd-ui__url=https://localhost:{port}
    .WaitFor(ui)
    .WithReference(ui);
```
​
#### Key design decisions
​
- **k3s as `ContainerResource`** — the cluster runs as a standard privileged privileged Linux container. No external CLI (k3d, Kind) required.
- **Helm and kubectl as containers** — `HelmReleaseResource` runs `docker.io/alpine/helm`, `K8sManifestResource` runs `docker.io/alpine/k8s`. No host-side `helm` or `kubectl` required.
- **No docker exec** — the health check reads the kubeconfig from a bind-mounted host directory written by k3s via `K3S_KUBECONFIG_OUTPUT`. Node readiness is probed via KubernetesClient `ListNodeAsync`.
- **Dual kubeconfig** — `local/kubeconfig.yaml` (`https://localhost:{port}`) for host processes; `container/kubeconfig.yaml` (`https://{name}:6443`) for DCP-network containers.
- **Service exposure without NodePort** — `K3sServiceEndpointResource` uses an in-process KubernetesClient WebSocket port-forward bound to `0.0.0.0:{port}`.
- **Kustomize auto-detection** — `AddK8sManifest` detects `kustomization.yaml` and uses `kubectl apply -k` automatically; otherwise uses `kubectl apply -f --server-side`.

edmondshtogu:
**Closes #1321**

## Overview of changes

Adds CommunityToolkit.Aspire.Hosting.K3s, a hosting integration that runs a lightweight Kubernetes cluster as an Aspire resource tree. Developers can declare a local Kubernetes cluster in Program.cs — with Helm charts, manifests, and exposed service endpoints — the same way they add Redis or PostgreSQL. No external tooling beyond a compatible container runtime (Docker or Podman) is required.

### What's included

**New package: src/CommunityToolkit.Aspire.Hosting.K3s/**
| File | Responsibility |
|---|---|
| K3sClusterResource.cs | ContainerResource — k3s server; holds kubeconfig directory path and image settings |
| K3sClusterOptions.cs | Configuration (pod/service CIDR, disabled components, k3s image tag, helm/kubectl image overrides) |
| K3sBuilderExtensions.cs | AddK3sCluster, WithDataVolume, WithLifetime, WithReference(cluster), WithK3sVersion, … |
| K3sReadinessHealthCheck.cs | File-based health check — polls cluster/kubeconfig.yaml, writes local/ + container/ variants, probes nodes via KubernetesClient |
| HelmReleaseResource.cs | ContainerResource — runs alpine/helm; child of cluster; exits 0 on success |
| K3sBuilderExtensions.Helm.cs | AddHelmRelease, WithHelmValue, WithHelmValuesFile |
| K8sManifestResource.cs | ContainerResource — runs alpine/k8s; child of cluster; exits 0 on success |
| K3sBuilderExtensions.Manifest.cs | AddK8sManifest with auto-detected Kustomize support |
| K3sServiceEndpointResource.cs | Resource — in-process port-forward; M1 passive health via IsReady flag |
| K3sBuilderExtensions.ServiceEndpoint.cs | AddServiceEndpoint, WithReference(endpoint) |
| K3sInProcessPortForwarder.cs | KubernetesClient WebSocket TCP forwarder; binds 0.0.0.0:{port} |
| K3sAgentResource.cs | Worker node support (K3sClusterOptions.AgentCount) |
| HelmContainerImageTags.cs / KubectlContainerImageTags.cs | Pinned image defaults; overridable via K3sClusterOptions |

**New Tests & Examples:**
 * **Unit Tests:** tests/CommunityToolkit.Aspire.Hosting.K3s.Tests/ — 87 unit tests covering resource registration, script generation, kubeconfig variants, Kustomize detection, values file injection, and public API null guards.
 * **Examples:** examples/k3s/CommunityToolkit.Aspire.Hosting.K3s.AppHost/
 * **TypeScript playground:** playground/polyglot/TypeScript/CommunityToolkit.Aspire.Hosting.K3s/ValidationAppHost/

### Key design decisions

 * **Health check via bind-mount, not docker exec.** k3s writes its kubeconfig to `K3S_KUBECONFIG_OUTPUT=/tmp/k3s-kubeconfig/kubeconfig.yaml`, bind-mounted to `AppHostDirectory/.k3s/{name}/cluster/` on the host. The health check polls `File.Exists`, rewrites server URLs into `local/` and `container/` variants, then confirms node readiness via `IKubernetes.CoreV1.ListNodeAsync`. No shell access, no `docker exec`, works with any container runtime.

 * **Helm and kubectl run as containers.** `HelmReleaseResource` and `K8sManifestResource` extend `ContainerResource` and are shown as children of the cluster in the Aspire dashboard. The install/apply script is injected via `WithContainerFiles`. They cannot use `WaitFor(cluster)` (Aspire forbids a child waiting for its parent), so their scripts poll for `/root/.kube/kubeconfig.yaml` — which only appears after the cluster health check passes — before proceeding. Consumers use `WaitForCompletion(helmRelease)` since these are run-to-completion containers.

 * **Kubeconfig delivered via bind-mount to all containers.** `WithReference(cluster)` on containers, the helm installer, and the kubectl applier all bind-mount `AppHostDirectory/.k3s/{name}/container/` (server: `https://{name}:6443`) at a known in-container path and set `KUBECONFIG`. Bind-mount is used uniformly so the kubeconfig updates automatically if the cluster is recreated without restarting dependent containers. No `KUBECONFIG_DATA` base64 encoding — all standard Kubernetes tooling (`kubectl`, `helm`, KubernetesClient SDK) works without custom bootstrap code.

 * **Kustomize auto-detected.** `AddK8sManifest` checks for `kustomization.yaml` at configuration time: if present, it bind-mounts the directory (preserving relative base references) and uses `kubectl apply -k`; otherwise it uses an async `WithContainerFiles` callback to copy only the YAML files at container-start time and applies with `kubectl apply -f --server-side`. The callback approach avoids Aspire's build-time path validation on the string overload. The script auto-detects the mode at runtime.

 * **Service exposure without NodePort.** `K3sServiceEndpointResource` starts an in-process KubernetesClient WebSocket port-forward bound to `0.0.0.0:{hostPort}`. Host resources receive `services__{name}__url=http(s)://localhost:{port}`; container resources receive `services__{name}__url=http(s)://host.docker.internal:{port}` with `--add-host=host.docker.internal:host-gateway` injected automatically via `ContainerRuntimeArgsCallbackAnnotation` (DCP does not inject this on Linux). The forwarder resolves `targetPort` from the service spec (not the service port) before opening the pod WebSocket, and only signals ready after a running pod is confirmed — not when the TCP listener starts.

 * **Image overrides via K3sClusterOptions.** The helm and kubectl container images are configurable via `HelmImage`/`HelmTag`/`HelmRegistry` and `KubectlImage`/`KubectlTag`/`KubectlRegistry` on `K3sClusterOptions`. Defaults: `docker.io/alpine/helm:3.17.3` and `docker.io/alpine/kubectl:1.36.0`.

 * **Robustness fixes from review.** `WithK3sVersion` propagates the image tag to all agent nodes (prevents server/agent version skew). `AddServiceEndpoint` validates the port is in the range 1–65535. Helm values files are indexed as `{i}-{filename}` so declaration order is preserved and basename collisions are safe. All helm `--set` values and `--values` paths are POSIX single-quote escaped via `ShellEscape()`.

### Usage

```csharp
var cluster = builder.AddK3sCluster("k8s")
    .WithDataVolume()
    .WithK3sVersion("v1.36.0-k3s1");

var widgetCrd = cluster.AddK8sManifest("widget-crd", "./k8s/crds/");

var argocd = cluster.AddHelmRelease("argocd", "argo-cd",
    repo: "https://argoproj.github.io/argo-helm",
    version: "7.8.0",
    @namespace: "argocd")
    .WithHelmValuesFile("./deploy/argocd-values.yaml");

var ui = cluster.AddServiceEndpoint("argocd-ui", "argocd-server", 443, "argocd")
    .WaitForCompletion(argocd);

builder.AddProject<Projects.WidgetOperator>("operator")
    .WaitForCompletion(widgetCrd)
    .WithReference(cluster);

    .WaitFor(ui)
    .WithReference(ui);
```

### Example
image

## PR Checklist
 * [x] Created a feature/dev branch in your fork (vs. submitting directly from a commit on main)
 * [x] Based off latest main branch of toolkit
 * [x] PR doesn't include merge commits (always rebase on top of our main, if needed)
 * [x] New integration
   * [x] Docs are written
   * [x] Added description of major feature to project description for NuGet package (4000 total character limit, so don't push entire description over that)
 * [x] Tests for the changes have been added (for bug fixes / features) (if applicable)
 * [x] Contains **NO** breaking changes
 * [x] Every new API (including internal ones) has full XML docs
 * [x] Code follows all style conventions

## Other information

**Security Assumptions**
This change assumes local development only. k3s runs in privileged mode (required by k3s/containerd). The bind-mounted kubeconfig directory is readable only by the AppHost user. `KubernetesClientConfiguration` uses the embedded CA cert from the kubeconfig; no `DangerousAcceptAnyServerCertificateValidator` is used.

**Remaining Follow-up Work**
 * Publish-time diagnostic when a `K3sClusterResource` has no production counterpart configured.


edmondshtogu:
@dotnet-policy-service agree

src/CommunityToolkit.Aspire.Hosting.K3s/K3sBuilderExtensions.Helm.cs
```diff
@@ -0,0 +1,226 @@
...
+
+        // --set flags override everything above (highest Helm precedence).
+        // Both key and value are single-quoted so quotes, spaces, or $ in values are safe.
+        foreach (var (key, value) in release.HelmValues)
+            sb.Append($" --set {ShellEscape($"{key}={value}")}");
+
+        return sb.ToString();
+    }
+
+    /// <summary>
```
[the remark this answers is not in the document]
edmondshtogu: Added HelmEscape() that escapes ,, {, }, \ before shell-quoting — Helm receives \, and treats it as a literal comma

src/CommunityToolkit.Aspire.Hosting.K3s/K3sInProcessPortForwarder.cs
```diff
@@ -0,0 +1,245 @@
...
+    public async Task RunAsync(ILogger logger, CancellationToken ct)
+    {
+        var backoff = TimeSpan.FromSeconds(2);
+
+        while (!ct.IsCancellationRequested)
+        {
+            var listener = new TcpListener(IPAddress.Any, localPort);
+            try
+            {
+                listener.Start();
```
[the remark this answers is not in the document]
edmondshtogu: Comment explains why 0.0.0.0 is required: DCP containers reach host.docker.internal:{port} which is not 127.0.0.1; warns users on shared networks

src/CommunityToolkit.Aspire.Hosting.K3s/api/CommunityToolkit.Aspire.Hosting.K3s.cs
```diff
@@ -0,0 +1,161 @@
...
+    public static partial class K3sManifestBuilderExtensions
+    {
+        [AspireExport("addK8sManifest", Description = "Applies Kubernetes YAML manifests to the k3s cluster")]
+        public static ApplicationModel.IResourceBuilder<ApplicationModel.K8sManifestResource> AddK8sManifest(this ApplicationModel.IResourceBuilder<ApplicationModel.K3sClusterResource> builder, string name, string path) { throw null; }
+    }
+
+    public static partial class K3sServiceEndpointExtensions
+    {
+        [AspireExport("addServiceEndpoint", Description = "Exposes a Kubernetes service as an Aspire endpoint resource")]
+        public static ApplicationModel.IResourceBuilder<ApplicationModel.K3sServiceEndpointResource> AddServiceEndpoint(this ApplicationModel.IResourceBuilder<ApplicationModel.K3sClusterResource> builder, string name, string serviceName, int servicePort, string @namespace = "default") { throw null; }
```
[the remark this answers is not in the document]
edmondshtogu: scheme parameter now appears in the API snapshot

src/CommunityToolkit.Aspire.Hosting.K3s/K3sBuilderExtensions.Manifest.cs
```diff
@@ -0,0 +1,204 @@
...
+        sb.AppendLine("if [ -f /k8s-manifests/kustomization.yaml ] || [ -f /k8s-manifests/kustomization.yml ]; then");
+        sb.AppendLine("  echo 'Detected kustomization — using kubectl apply -k'");
+        sb.AppendLine("  kubectl apply -k /k8s-manifests --server-side --field-manager=aspire-k3s --force-conflicts");
+        sb.AppendLine("else");
+        sb.AppendLine("  kubectl apply -f /k8s-manifests --server-side --field-manager=aspire-k3s --force-conflicts");
+        sb.AppendLine("fi");
+
+        // Wait for CRD Established condition if any CRDs are present.
+        sb.AppendLine("if kubectl get crd --no-headers 2>/dev/null | grep -q .; then");
+        sb.AppendLine("  kubectl wait --for=condition=Established crd --all --timeout=300s");
```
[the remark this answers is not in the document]
edmondshtogu: Now filters CRDs by aspire-k3s field-manager before waiting — a stuck pre-existing CRD no longer blocks an unrelated manifest

edmondshtogu:
@aaronpowell, @IEvangelist, @marshalhayes or @tommasodotNET can one of you please review the PR (sorry for pulling you in the comment, but you are the most active contributors here)

edmondshtogu:
@davidfowl do you know who can help with the review here?

IEvangelist (changes requested):
## Thorough code review

13 findings total: **1 blocking**, **7 major**, **5 minor/medium**. Overall verdict: **request changes**. The design is sound — the execution needs a careful pass on lifecycle, idempotency, and atomicity.

---

### 🚨 Blocking

#### 1. Unresolved Git merge conflict markers in `.gitignore` (lines 26–38)
Literal `<<<<<<< main`, `=======`, `>>>>>>> main` markers were committed. `git grep "<<<<<<<"` confirms `.gitignore:26`. The conflict was resolved locally but the markers were never deleted. This alone should keep the PR out of `main`.

---

### 🔴 Major

#### 2. `K3sReadinessHealthCheck._cachedClient` is dead state — `Kubernetes` client leaks every tick
`K3sBuilderExtensions.cs:211-215` registers the check with `sp => new K3sReadinessHealthCheck(resource, resource.ApiEndpoint)`. `DefaultHealthCheckService.RunCheckAsync` invokes the factory **on every check**, so `_cachedClient` always starts `null` and the cache is dead code. Every tick allocates a fresh `Kubernetes`/`HttpClient` (line 116) which is never disposed (the class is not `IDisposable`). The "stale credentials" recovery in the catch block (lines 70–84) can only ever dispose the just-created instance, never a stale one across calls — it's a no-op pattern.

**Fix:** register as a singleton (`var hc = new K3sReadinessHealthCheck(...); sp => hc`) and make the class `IDisposable`, **or** drop the cache and `using` the client locally within each `CheckHealthAsync` call.

#### 3. Non-atomic kubeconfig overwrite on every health check
`K3sReadinessHealthCheck.cs:99-113`. Because of #2, `EnsureClientAsync` runs every tick and `File.WriteAllTextAsync` truncates both `local/kubeconfig.yaml` and `container/kubeconfig.yaml` in place. Readers hitting the race window can observe a partial/empty file — host projects with `KUBECONFIG=…/local/kubeconfig.yaml`, the helm/kubectl scripts polling for the file, and any guest `BuildConfigFromConfigFile(...)` at startup that does not retry.

**Fix:** write to a temp file in the same directory then `File.Move(tempPath, finalPath, overwrite: true)`; only rewrite when the source raw YAML's hash has changed since last write.

#### 4. `AllocatePort()` TOCTOU race + interface mismatch
`K3sBuilderExtensions.ServiceEndpoint.cs:200-207`. Two distinct defects:

1. **Window race:** the probe listener is stopped *before* `K3sInProcessPortForwarder` rebinds on the same port. Any other process can claim the port in the gap, causing the subsequent `new TcpListener(IPAddress.Any, localPort).Start()` to throw `SocketException(AddressAlreadyInUse)`. The exception is caught by the outer `catch (Exception ex)` (lines 79–85), logged, and the loop backs off and **retries with the same already-stolen port** — it never picks a fresh one. The endpoint appears in the dashboard as perpetually unhealthy with no obvious cause.
2. **Interface mismatch:** the probe binds on `IPAddress.Loopback`, but the forwarder binds on `IPAddress.Any` (line 43 of `K3sInProcessPortForwarder.cs`). A port free on 127.0.0.1 can still conflict on 0.0.0.0 (another service bound on a non-loopback interface) — same failure mode.

**Fix:** probe on `IPAddress.Any` (match the actual bind). Better: pass the live listener instance into `K3sInProcessPortForwarder` and keep it bound through the handoff so there is no release-rebind window. On retry inside the forwarder loop, allocate a fresh port rather than reusing the same one.

#### 5. `WithDataVolume` is not idempotent — duplicate volume mounts
`K3sBuilderExtensions.cs:327-335`. Each call appends a fresh volume annotation targeting `/var/lib/rancher/k3s`. Two calls (a common error when composing builder extensions) produce two `ContainerMountAnnotation`s at the same target → Docker rejects with `Duplicate mount point: /var/lib/rancher/k3s`. The unit test `WithDataVolumeAddsSingleVolumeMount` only exercises a single call; there is no test asserting idempotency.

**Fix:** remove the previous data-volume annotation before adding, or use `ResourceAnnotationMutationBehavior.Replace` semantics consistent with `WithLifetime` (line 410-412).

#### 6. `WithReference(cluster)` adds a duplicate bind-mount on repeat invocation
`K3sBuilderExtensions.cs:368-388`. Same defect as #5: the container branch unconditionally appends a new `ContainerMountAnnotation` for `/var/k3s`. Two calls — easy when composing helper extension methods — create two bind-mounts to the same target, which Docker rejects at start with the duplicate-mount error buried in DCP logs.

**Fix:** check `destination.Resource.Annotations.OfType<ContainerMountAnnotation>().Any(m => m.Target == "/var/k3s")` and skip (or replace).

#### 7. Helm `--set` key injection — only the value is escaped against Helm's parser
`K3sBuilderExtensions.Helm.cs:214-228`:
```csharp
foreach (var (key, value) in release.HelmValues)
    sb.Append($" --set {ShellEscape($"{key}={HelmEscape(value)}")}");
```
The shell layer is correct — POSIX `'\''` single-quoting is robust. But `HelmEscape` is only applied to the **value**, never the **key**. Helm's own `--set` parser splits on `,`, treats `{`/`}` as list literals, and uses `\` as the escape character. A key containing a comma — e.g. `WithHelmValue("a.b,c.d=evil", "true")` — produces a single shell-quoted argument that Helm then parses as **two** assignments: `a.b=true` and `c.d=evil`. For local dev this is a foot-gun rather than a vulnerability (the AppHost is the trust boundary), but a real bug. None of the tests exercise a key with commas/braces/backslashes.

**Fix:** apply `HelmEscape` to both key and value, or validate keys against a conservative pattern (e.g. `^[A-Za-z_][A-Za-z0-9_.\[\]]*$`) and throw `ArgumentException` otherwise.

#### 8. Port-forwarder has no shutdown/dispose path — fire-and-forget leak risk
`K3sBuilderExtensions.ServiceEndpoint.cs:179` + the whole of `K3sInProcessPortForwarder.cs`. `_ = Task.Run(() => forwarder.RunAsync(logger, ct), ct);` — fire-and-forget with no handle retained. `K3sInProcessPortForwarder` exposes neither `IDisposable`/`IAsyncDisposable` nor a `Stop` method. The only way to terminate the listener on `0.0.0.0:{port}` is the `ct` passed in from the `ResourceReadyEvent` subscription.

Two concerns:

1. **Cancellation token lifetime:** `ct` is the event-handler's token. It is application-scoped in most Aspire versions, but the contract is not "AppHost stopping" — if the platform ever changes the semantics, the 0.0.0.0 socket leaks for the rest of the AppHost session with no `IAsyncDisposable` to fall back to.
2. **Per-connection task orphan:** inside `RunAsync`, each accepted connection spawns `_ = Task.Run(() => ForwardConnectionAsync(tcp, logger, ct), CancellationToken.None)` — uses `CancellationToken.None`, not `ct`. Unhandled exceptions in those tasks are caught and `LogDebug`'d only, so a thrown WebSocket exception during shutdown is invisible to operators.

Also note: `ForwardConnectionAsync` creates a fresh `Kubernetes` client (and thus a fresh TLS handshake) per accepted TCP connection (line 168-169) — wasteful but each is `using`-disposed.

**Fix:** make `K3sInProcessPortForwarder` implement `IAsyncDisposable`, store the listener and a linked `CancellationTokenSource` as fields, retain the forwarder reference on the endpoint resource (or in a registry keyed by resource name), and dispose on AppHost stop via `builder.Eventing.Subscribe<BeforeStopEvent>` or similar.

---

### 🟡 Medium

#### 9. Kustomize directory and helm/kubectl kubeconfig bind-mounts are read-write
`K3sBuilderExtensions.Manifest.cs:81, 88` and `K3sBuilderExtensions.Helm.cs:99`. None of these bind-mounts pass `isReadOnly: true`:

- Kustomize overlay dir → `/k8s-manifests` (Manifest.cs:88) — the user's source tree under their checkout.
- `container/` (kubeconfig) → `/root/.kube` (Manifest.cs:81 and Helm.cs:99) — host-side kubeconfig directory.

The kubectl/helm scripts don't write to these paths today, so this is latent. But any plugin (`krew`, `helm-secrets`, etc.) the user adds via a custom image override (`opts.HelmImage`, `opts.KubectlImage`) gains write access to the host filesystem at those paths. `WithReference(cluster)` for user containers correctly uses `isReadOnly: true` (line 385) — the helm/kubectl/manifest paths should match.

**Fix:** add an overload that passes `isReadOnly: true`, or build `ContainerMountAnnotation` directly as `WithReference` does.

#### 10. `K3sServiceEndpointResource.IsReady` can be stuck `false` forever with no recovery surface
`K3sBuilderExtensions.ServiceEndpoint.cs:134-188` + `K3sServiceEndpointResource.cs:56`. `RunEndpointAsync` is fire-and-forget. The outer `try/catch` covers the synchronous setup, but the spawned `forwarder.RunAsync(...)` Task is not awaited — any unhandled exception inside `RunAsync` that escapes its inner `catch (Exception ex)` (e.g. an exception thrown by `notifications.PublishUpdateAsync` during the `onReadyChanged` callback) terminates the task silently. `IsReady` stays `false`, the dependent `K3sServiceEndpointHealthCheck.CheckHealthAsync` keeps returning `Unhealthy`, and any consumer using `WaitFor(serviceEndpoint)` blocks forever with no diagnostic.

The `_ = notifications.PublishUpdateAsync(...)` call on line 172 is also fire-and-forget; if it throws, the exception escapes the synchronous part of the callback (since `Action<bool>` cannot await), reaches `TaskScheduler.UnobservedTaskException`, and disappears.

**Fix:** make `onReadyChanged` an `async`-aware delegate, await `PublishUpdateAsync` inside it, and surface failures via the resource notification with `State = FailedToStart`. Add a watchdog: if the forwarder Task faults, publish `FailedToStart` on the endpoint resource so the dashboard surfaces the problem.

---

### 🟢 Minor

#### 11. PR description references an integration test project that doesn't exist
The PR overview claims `tests/CommunityToolkit.Aspire.Hosting.K3s.IntegrationTests/ — ubuntu-latest only` exists. Only the unit-test project `tests/CommunityToolkit.Aspire.Hosting.K3s.Tests/` (87 facts, no `[RequiresDocker]`) is present. Given the scope of dangerous code (port forwarder lifecycle, kubeconfig race, privileged containers, shell-script injection), the absence of any integration test that exercises an actual k3s container is a real gap — but the discrepancy with the description is what concerns me here.

**Fix:** add the integration test project (it would catch #2, #3, #5, #6, #8 trivially) or correct the PR description.

#### 12. `WithHttpsDeveloperCertificate()` on the k3s container is misuse
`K3sBuilderExtensions.cs:141`. `WithHttpsDeveloperCertificate()` configures the ASP.NET Core dev cert for JavaScript/Python apps. k3s manages its own server certificate (the very cert whose SANs are extended via `--tls-san` on lines 90-93). Injecting the ASP.NET dev cert into the k3s container is at best a no-op, at worst pollutes the container with unused files. The build still succeeds.

**Fix:** remove the `.WithHttpsDeveloperCertificate()` call; the `WithHttpsEndpoint` declaration alone is sufficient.

#### 13. `AddServiceEndpoint` missing whitespace validation on `serviceName`/`@namespace`
`K3sBuilderExtensions.ServiceEndpoint.cs:32-47`. The port range validation is correctly added (lines 44-46). But `serviceName` and `@namespace` only get `ArgumentNullException.ThrowIfNull` — empty or whitespace strings pass validation and produce an unhelpful runtime failure inside the Kubernetes API call. The convention elsewhere in this same file (e.g. `WithHelmValuesFile` line 133, `WithPodSubnet` line 281, `WithServiceSubnet` line 293) is `ArgumentException.ThrowIfNullOrWhiteSpace`.

**Fix:** use `ArgumentException.ThrowIfNullOrWhiteSpace(serviceName)` and `(@namespace)`.

---

### ✅ Categories with no findings

- **Aspire resource semantics for parent-child** — correctly modeled via `IResourceWithParent<K3sClusterResource>`.
- **Helm shell injection via `ShellEscape`** — the POSIX `'\''` technique is correctly implemented and properly applied to release name, chart ref, namespace, version, values-file paths, and `--set` values. The keys are not Helm-escaped (#7) but the shell layer itself is safe.
- **Repo conventions on namespaces** — extension method classes correctly use `namespace Aspire.Hosting;` file-scoped; resources use `Aspire.Hosting.ApplicationModel;`; internal helpers use `CommunityToolkit.Aspire.Hosting;`. `is null`/`is not null` is used consistently. `[ResourceName]` is applied to `name` parameters in public extension methods. The CI workflow (`tests.yaml`) was updated.
- **Container runtime annotation usage** — `ContainerRuntimeArgsCallbackAnnotation(Action<IList<object>>)` matches the documented Aspire API. Multiple invocations append redundant `--add-host=...` flags but Docker handles duplicates idempotently.
- **`api/CommunityToolkit.Aspire.Hosting.K3s.cs`** — this is a new file, not an edit to an existing generated file, so the convention is not violated.

---

### Summary

**Top 3 most important findings:**

1. **Blocking:** `.gitignore` contains literal Git merge conflict markers.
2. **Major:** `K3sReadinessHealthCheck._cachedClient` is dead state — the registration-factory pattern creates a fresh instance per check, so the cache never carries over. Every health-check tick allocates a new `Kubernetes`/`HttpClient` that is never disposed, **and** rewrites both kubeconfig variants non-atomically (#2 + #3 combined).
3. **Major:** `AllocatePort()` + `K3sInProcessPortForwarder` lifecycle. The port-allocation TOCTOU + interface-mismatch race (#4), the fire-and-forget forwarder with no dispose path (#8), and the `IsReady`-stuck-false failure mode (#10) together mean the service-endpoint feature has no reliable failure path: a transient startup race can leave the resource permanently unhealthy with no operator-visible diagnostic.

**Design assessment:**

The overall design — modelling the cluster as a privileged container with Helm/manifest/service-endpoint children — is sound for the stated local-dev inner-loop use case. The decisions to (a) run helm/kubectl as containers rather than requiring host binaries, (b) rewrite kubeconfig into `local/` and `container/` variants and bind-mount the appropriate one per consumer type, (c) make child containers poll for the kubeconfig file rather than `WaitFor`-ing the parent (avoiding the cluster-vs-node Ready deadlock), and (d) use an in-process `KubernetesClient` WebSocket port-forward bound on 0.0.0.0 for both host and container reach — are all well-considered and individually correct. The k3d-inspired init entrypoint and the deliberate avoidance of `--cluster-init` (with the explanatory comment about etcd peer IP changes) show evidence of real operational testing.

The execution falls short on **resource lifecycle management** (no `IDisposable` surfaces, fire-and-forget tasks with no observability, broken cache, leaked `HttpClient`s), **idempotency of builder operations** (`WithDataVolume`, `WithReference` are not safe to call twice), and **atomicity of shared state** (kubeconfig file rewrites). These are all fixable without touching the design — the core architecture is good; the implementation needs a careful pass on "what happens on the second call / on shutdown / on transient failure".

IEvangelist (commented):
A few other things to consider.

src/CommunityToolkit.Aspire.Hosting.K3s/K3sBuilderExtensions.cs
```diff
@@ -0,0 +1,467 @@
...
+            // k3s starts — exactly what k3d does via /bin/k3d-entrypoint*.sh.
+            // WithContainerFiles injects the script via `docker cp` — no bind mounts,
+            // no host-side temp files, works on all platforms.
+            .WithContainerFiles("/", [new ContainerFile
+            {
+                Name = "aspire-k3s-entrypoint.sh",
+                Contents = K3sInitEntrypointScript,
+                Mode = UnixFileMode.UserRead | UnixFileMode.UserWrite | UnixFileMode.UserExecute
+                     | UnixFileMode.GroupRead | UnixFileMode.GroupExecute
+                     | UnixFileMode.OtherRead | UnixFileMode.OtherExecute,
```
IEvangelist: This exact `UnixFileMode` combination (effectively `0755` / `rwxr-xr-x`) is repeated four times across the integration:

- `K3sBuilderExtensions.cs:71-73` (server entrypoint, here)
- `K3sBuilderExtensions.cs:182-184` (agent entrypoint)
- `K3sBuilderExtensions.Helm.cs:79-81` (`helm-install.sh`)
- `K3sBuilderExtensions.Manifest.cs:74-76` (`kubectl-apply.sh`)

Let's encapsulate it into a single internal constant with a doc comment explaining **why** these specific bits were chosen as the default for injected shell scripts (UID assumptions inside the various base images, why `OtherExecute` is required, etc.) — that context is non-obvious from the call sites and would help future maintainers reason about whether `0750` or `0700` would be safer for a given script.

Something like:

```csharp
/// <summary>
/// Default Unix file mode (0755 / rwxr-xr-x) for shell scripts injected
/// into k3s/helm/kubectl containers via <c>WithContainerFiles</c>.
/// </summary>
/// <remarks>
/// // …explain the rationale here: why these exact bits, what fails without
/// // them, and the trust assumptions about the container's runtime UID.
/// </remarks>
internal const UnixFileMode ExecutableScriptMode =
    UnixFileMode.UserRead  | UnixFileMode.UserWrite  | UnixFileMode.UserExecute |
    UnixFileMode.GroupRead | UnixFileMode.GroupExecute |
    UnixFileMode.OtherRead | UnixFileMode.OtherExecute;
```

Each call site then becomes `Mode = K3sFileModes.ExecutableScriptMode,`.

edmondshtogu: Thanks for the detailed review @IEvangelist, in the last two commits I addressed what you found.

|# | Finding | Status|
|-- | -- | --|
|1 🚨 | `.gitignore` conflict markers | ✅ Resolved|
|2 🔴 | Health check factory leaks client every tick | ✅ Singleton + cache dropped|
|3 🔴 | Non-atomic `kubeconfig` overwrite | ✅ `WriteAtomicAsync`|
|4 🔴 | `AllocatePort` TOCTOU + interface mismatch | ✅ `IPAddress.Any`, fresh port on retry|
|5 🔴 | `WithDataVolume` not idempotent | ✅ Removes existing annotation first|
|6 🔴 | `WithReference` duplicate mount | ✅ Idempotency guard on target path|
|7 🔴 | Helm `--set` key not Helm-escaped | ✅ `HelmEscape` on both key and value|
|8 🔴 | Port-forwarder fire-and-forget, no dispose | ✅ `IAsyncDisposable`, linked CTS, `ResourceStoppedEvent`|
|9 🟡 | Bind-mounts read-write | ✅ `isReadOnly: true` on all `kubeconfig` mounts|
|10 🟡 | `IsReady` stuck false, `PublishUpdateAsync` fire-and-forget | ✅ `await using` ensures cleanup; `PublishUpdateAsync` at line 185 is fire-and-forget inside the synchronous callback — acceptable for local dev since the notification system is internal and well-tested — if you want the full async-delegate refactor, let me know and I will change it|
|11 🟢 | PR description references missing integration test project | ✅ Removed (I created the project initially under `tests/CommunityToolkit.Aspire.Hosting.K3s.IntegrationTests/`, but the tests locally took to long to run, therefore I removed)|
|12 🟢 | `WithHttpsDeveloperCertificate()` misuse | ✅ Removed|
|13 🟢 | Missing whitespace validation on `serviceName`/`@namespace` | ✅ `ThrowIfNullOrWhiteSpace`|

examples/k3s/CommunityToolkit.Aspire.Hosting.K3s.AppHost/k8s/monitoring/kustomization.yaml
IEvangelist, line 1: Is the name of this file common, or is this a play on words leaning on the `k` from Kubernetes?

playground/polyglot/TypeScript/CommunityToolkit.Aspire.Hosting.K3s/ValidationAppHost/aspire.config.json
```diff
@@ -0,0 +1,26 @@
...
+      "applicationUrl": "http://localhost:15243",
+      "environmentVariables": {
+        "ASPIRE_DASHBOARD_OTLP_ENDPOINT_URL": "http://localhost:19027",
+        "ASPIRE_RESOURCE_SERVICE_ENDPOINT_URL": "http://localhost:20141",
+        "ASPIRE_ALLOW_UNSECURED_TRANSPORT": "true"
+      }
+    }
+  },
+  "packages": {
+    "CommunityToolkit.Aspire.Hosting.K3s": ""
```
IEvangelist: Hey @sebastienros, is this correct? I always forget...`""` is valid, right?

examples/k3s/CommunityToolkit.Aspire.Hosting.K3s.AppHost/k8s/monitoring/kustomization.yaml
edmondshtogu: yes, check here https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/

IEvangelist (approved):
This LGTM, I'm not going to merge it though as this really belongs to @aaronpowell — I'll leave that to him. Thank you for the PR though, @edmondshtogu.

edmondshtogu:
Thanks @IEvangelist, also test I added are passing, the failing ones are not related with my changes.

playground/polyglot/TypeScript/CommunityToolkit.Aspire.Hosting.K3s/ValidationAppHost/aspire.config.json
sebastienros: Put the location of the .csproj there, there are other examples in the repo

examples/k3s/CommunityToolkit.Aspire.Hosting.K3s.AppHost/CommunityToolkit.Aspire.Hosting.K3s.AppHost.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Aspire.AppHost.Sdk/13.3.0">
+<Project Sdk="Aspire.AppHost.Sdk/13.4.0-preview.1.26275.15">
```
davidfowl: This should be 13.4.0

davidfowl:
Did we verify that the TypeScript path works end to end?

examples/k3s/CommunityToolkit.Aspire.Hosting.K3s.AppHost/CommunityToolkit.Aspire.Hosting.K3s.AppHost.csproj
edmondshtogu: I merged the branch with main and updated the version 🙌🏻

edmondshtogu:
> Did we verify that the TypeScript path works end to end?

@davidfowl yes, I introduced additional changes since the callback (Action<K3sClusterOptions>) was unusable in TypeScript polyglot apphosts — the generated Promise-based wrapper never resolved, making agentCount and the Helm/kubectl image overrides silently unreachable. Here is the TypeScript polyglot in action:
image

edmondshtogu:
Can someone approve the tests so I can see if there is anything else to be improved?

edmondshtogu:
Can someone approve the workflows again so I can see if there is anything else to be improved?
> [!IMPORTANT]  
> The `timheuer/setup-aspire` action installs the latest available CLI build in staging channel, which may be a newer patch than the version pinned in `Directory.Build.props`. The previous check required an exact match (or the exact version with a `+sha` suffix), so it would fail whenever the action installed `13.4.3+abc` while the repo expected `13.4.0` — even though a newer patch is always compatible. I introduced a fix to strips build metadata before comparing and allow any installed version ≥ the expected version.  The [`.Net main`](https://github.com/edmondshtogu/CommunityToolkit-Aspire/actions/runs/27134466334) Workflow now is passing!

edmondshtogu:
@aaronpowell could you please take a look at the PR and share if there is something you would like to change, or if you are happy to approve the integration? I'd love to start using this in my projects. Currently, because https://github.com/microsoft/aspire/issues/16878 is still open, I can't add third-party integrations via the Aspire CLI, so having this package available is the only way around.

aaronpowell (changes requested):
Tests that are going to require Docker to run will need to be marked with the attribute to indicate that, otherwise they'll fail on Windows during CI.

edmondshtogu:
@aaronpowell The Windows CI failure was a file-locking race in `EnsureKubeconfigPlaceholder`, not a Docker dependency. `AddK3sCluster` writes placeholder files to `{AppHostDirectory}/.k3s/{name}/` at configuration time. With xunit's default parallel execution, multiple tests calling `AddK3sCluster("k8s")` simultaneously hit a `TOCTOU` on Windows — both check `!File.Exists`, both pass, then the second `File.WriteAllText` fails with `IOException`: file is being used by another process.

Fixed by replacing the check-then-create pattern with `FileMode.CreateNew` and swallowing the `IOException` when the file already exists — atomic create-if-not-exists, safe under concurrent calls.

None of the tests actually require Docker: they all use either mocked `IKubernetes` via `Moq` or `DistributedApplication.CreateBuilder()` + resource inspection without ever calling `RunAsync`. Marking them `[RequiresDocker]` would skip valid platform coverage unnecessarily.

.github/actions/setup-runtimes-caching/action.yml
```diff
@@ -43,13 +43,14 @@ runs:
...
-            ;;
-        esac
+        # Strip build metadata (e.g. 13.4.3+becb48e → 13.4.3) before comparing so that
+        # a newer patch release installed by the action does not fail this check.
+        INSTALLED_BASE="${INSTALLED_ASPIRE_CLI_VERSION%%+*}"
+        # Pass when installed base version >= expected version (sort -V = version ordering).
+        if ! printf '%s\n' "${EXPECTED_ASPIRE_CLI_VERSION}" "${INSTALLED_BASE}" | sort -V -C; then
+          echo "Expected Aspire CLI version >= ${EXPECTED_ASPIRE_CLI_VERSION}, but found ${INSTALLED_ASPIRE_CLI_VERSION}."
+          exit 1
+        fi
```
aaronpowell: I don't think this is needed, we already handle that in CI

examples/k3s/CommunityToolkit.Aspire.Hosting.K3s.AppHost.TypeScript/apphost.mts
aaronpowell, line 1: Can we move this file, and the others related to it, into the examples folder as the `AppHost.TypeScript` for the integration, to follow the pattern we have with other integrations.

playground/polyglot/TypeScript/CommunityToolkit.Aspire.Hosting.K3s/ValidationAppHost/aspire.config.json
aaronpowell, line 1: Need to update this to match the latest design of the config - SDK version should be there.

src/CommunityToolkit.Aspire.Hosting.K3s/api/CommunityToolkit.Aspire.Hosting.K3s.cs
aaronpowell, line 1: Can you delete this file, we auto-generate it.

src/CommunityToolkit.Aspire.Hosting.K3s/K3sClusterResource.cs
```diff
@@ -0,0 +1,101 @@
...
+{
+    internal const string ApiServerEndpointName = "api";
+
+    /// <summary>Container image settings for the Helm installer, resolved from cluster options.</summary>
+    internal (string Registry, string Image, string Tag) HelmImageInfo { get; set; }
+        = ("docker.io", "alpine/helm", "3.17.3");
+
+    /// <summary>Container image settings for the kubectl manifest applier, resolved from cluster options.</summary>
+    internal (string Registry, string Image, string Tag) KubectlImageInfo { get; set; }
+        = ("docker.io", "alpine/kubectl", "1.36.0");
```
aaronpowell: The default values of the tuple should be driven from the types we have to ensure there's only a single place to update.

src/CommunityToolkit.Aspire.Hosting.K3s/K3sBuilderExtensions.ServiceEndpoint.cs
```diff
@@ -0,0 +1,249 @@
...
+        string? scheme = null)
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentNullException.ThrowIfNull(name);
+        ArgumentException.ThrowIfNullOrWhiteSpace(serviceName);
+        ArgumentException.ThrowIfNullOrWhiteSpace(@namespace);
+
+        if (servicePort is < 1 or > 65535)
+            throw new ArgumentOutOfRangeException(nameof(servicePort),
+                servicePort, "Service port must be in the range 1–65535.");
```
aaronpowell: We should leave this to Aspire to handle. While it's a valid check, no other integration does that so I'd prefer to be consistent.

src/CommunityToolkit.Aspire.Hosting.K3s/K3sBuilderExtensions.cs
```diff
@@ -0,0 +1,732 @@
...
+        int? agentCount = null)
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentNullException.ThrowIfNull(name);
+
+        var resource = new K3sClusterResource(name)
+        {
+            HelmImageInfo = (HelmContainerImageTags.Registry, HelmContainerImageTags.Image, HelmContainerImageTags.Tag),
+            KubectlImageInfo = (KubectlContainerImageTags.Registry, KubectlContainerImageTags.Image, KubectlContainerImageTags.Tag),
+        };
```
aaronpowell: Given that `K3sClusterResource` is a container resource, why does it then also have container info for other containers? Should they not be modelled as their own resources?

src/CommunityToolkit.Aspire.Hosting.K3s/K3sBuilderExtensions.cs
```diff
@@ -0,0 +1,732 @@
...
+        var tag = K3sContainerImageTags.Tag;
+
+        // ── Kubeconfig directory on the host ──────────────────────────────────
+        // AppHostDirectory/.k3s/{name}/ holds three sub-directories:
+        //   cluster/  — bind-mounted into the k3s container; k3s writes kubeconfig.yaml here
+        //   local/    — rewritten by the health check with server: https://localhost:{port}
+        //   container/ — rewritten by the health check with server: https://{name}:6443
+        var kubeconfigDir = Path.Combine(builder.AppHostDirectory, ".k3s", name);
+        var clusterDir = Path.Combine(kubeconfigDir, "cluster");
+        Directory.CreateDirectory(clusterDir);
```
aaronpowell, line 80: Should we error-handle the folder already existing?

src/CommunityToolkit.Aspire.Hosting.K3s/K3sReferenceAnnotations.cs
aaronpowell, line 1: Should this be deleted?

aaronpowell (changes requested):
Pretty solid integration, almost ready to merge in.

Couple of changes needed on this and we'll need to have a test that bootstraps the TypeScript app host (you'll find examples in the repo on how to do that).

Some of the questions I have might be due to naivety on k3 (I've not used it myself).

.github/actions/setup-runtimes-caching/action.yml
edmondshtogu: Done, I added since the CI was failing in my fork side due to version mismatch.

examples/k3s/CommunityToolkit.Aspire.Hosting.K3s.AppHost.TypeScript/apphost.mts
edmondshtogu: Sure

playground/polyglot/TypeScript/CommunityToolkit.Aspire.Hosting.K3s/ValidationAppHost/aspire.config.json
edmondshtogu: Done, the config now includes `"sdk": { "version": "13.4.3" }` matching the current repository SDK version.

src/CommunityToolkit.Aspire.Hosting.K3s/api/CommunityToolkit.Aspire.Hosting.K3s.cs
edmondshtogu: Removed it!

src/CommunityToolkit.Aspire.Hosting.K3s/K3sClusterResource.cs
edmondshtogu: Done — defaults now reference the constants: `HelmContainerImageTags.Registry.Image.Tag` and `KubectlContainerImageTags.Registry.Image.Tag`. No more hardcoded strings in the class.

src/CommunityToolkit.Aspire.Hosting.K3s/K3sBuilderExtensions.ServiceEndpoint.cs
edmondshtogu: Done, removed the port range validation to stay consistent with other integrations.

src/CommunityToolkit.Aspire.Hosting.K3s/K3sBuilderExtensions.cs
edmondshtogu: `HelmReleaseResource` and `K8sManifestResource` are modeled as their own separate resources — they show up independently in the Aspire dashboard and DCP manages them as one-shot containers. `HelmImageInfo`/`KubectlImageInfo` on `K3sClusterResource` are internal configuration bags whose only purpose is to carry image overrides from the user-facing `WithHelmImage`/`WithKubectlImage` calls on the cluster builder to the ephemeral runner containers created later inside `AddHelmRelease`/`AddK8sManifest`. The cluster container itself never reads these fields. There is no cleaner alternative: the user calls `cluster.WithHelmImage(tag: "3.18.0")` before any releases are added, so the override has to live somewhere the child factory can find it at `AddHelmRelease` time.

src/CommunityToolkit.Aspire.Hosting.K3s/K3sReferenceAnnotations.cs
edmondshtogu: Done, deleted!

examples/k3s/CommunityToolkit.Aspire.Hosting.K3s.AppHost.TypeScript/apphost.mts
edmondshtogu: Added another example for `AppHost.Go` as well.

src/CommunityToolkit.Aspire.Hosting.K3s/K3sBuilderExtensions.cs
edmondshtogu: No change needed for the already-exists case. `Directory.CreateDirectory` is idempotent by design: it creates all missing directories in the path and returns normally if they already exist.

Other exceptions it can throw (`UnauthorizedAccessException`, `PathTooLongException`, `IOException` when the path is a file) represent genuine issues that should surface to the developer at startup rather than be swallowed — a missing permission or a stale file blocking the `kubeconfig` directory is something the user needs to act on.

tests/CommunityToolkit.Aspire.Hosting.K3s.Tests/TypeScriptAppHostTests.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    [Fact]
+    public async Task TypeScriptAppHostCompilesAndStarts()
+    {
+        await TypeScriptAppHostTest.Run(
+            appHostProject: "CommunityToolkit.Aspire.Hosting.K3s.AppHost.TypeScript",
+            packageName: "CommunityToolkit.Aspire.Hosting.K3s",
+            exampleName: "k3s",
+            waitForResources: ["k8s", "podinfo-web"],
+            waitTimeoutSeconds: 300,
```
edmondshtogu, line 17: @aaronpowell I enabled the TS tests and also extended the `TypeScriptAppHostTest.Run` to expose the timeout as in my device it takes `~2m50s` for the tests to complete and in the workflow job might need more time to complete.

aaronpowell: Is the time required just due to the complexity of setting up all the resources for this integration? Seems like a very long startup time.

edmondshtogu: K3s need some time to setup all the moving parts.

edmondshtogu: I can see that the k8s took 15.7s to become healthy, but the scheduling didn't happen and I suspect that is due to available resources in the runner. I changed the test to start the cluster with a single node and will monitor the new time, if the port-forward is going to start faster will remove the extended timeout

edmondshtogu: @aaronpowell can you approve the tests again please?

aaronpowell:
I've just tried running the TypeScript app host on my devcontainer and the `podinfo-web` resource doesn't seem to be starting.

Here's the logs for it:

```
Port-forward: 0.0.0.0:40629 → svc/podinfo.podinfo:9898
Service podinfo/podinfo not yet ready; retrying…
k8s.Autorest.HttpOperationException: Operation returned an invalid status code 'NotFound', response body {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"services \"podinfo\" not found","reason":"NotFound","details":{"name":"podinfo","kind":"services"},"code":404}

   at k8s.Kubernetes.SendRequestRaw(String requestContent, HttpRequestMessage httpRequest, CancellationToken cancellationToken) in /_/src/KubernetesClient/Kubernetes.cs:line 180
   at k8s.AbstractKubernetes.ICoreV1Operations_ReadNamespacedServiceWithHttpMessagesAsync[T](String name, String namespaceParameter, Nullable`1 pretty, IReadOnlyDictionary`2 customHeaders, CancellationToken cancellationToken) in /_/src/KubernetesClient/obj/Release/net10.0/generated/LibKubernetesGenerator/LibKubernetesGenerator.KubernetesClientSourceGenerator/CoreV1Operations.g.cs:line 4018
   at k8s.AbstractKubernetes.k8s.ICoreV1Operations.ReadNamespacedServiceWithHttpMessagesAsync(String name, String namespaceParameter, Nullable`1 pretty, IReadOnlyDictionary`2 customHeaders, CancellationToken cancellationToken) in /_/src/KubernetesClient/obj/Release/net10.0/generated/LibKubernetesGenerator/LibKubernetesGenerator.KubernetesClientSourceGenerator/CoreV1Operations.g.cs:line 4028
   at k8s.CoreV1OperationsExtensions.ReadNamespacedServiceAsync(ICoreV1Operations operations, String name, String namespaceParameter, Nullable`1 pretty, CancellationToken cancellationToken) in /_/src/KubernetesClient/obj/Release/net10.0/generated/LibKubernetesGenerator/LibKubernetesGenerator.KubernetesClientSourceGenerator/CoreV1OperationsExtensions.g.cs:line 24966
   at CommunityToolkit.Aspire.Hosting.K3sInProcessPortForwarder.WaitForServiceReadyAsync(ILogger logger, CancellationToken ct) in /workspaces/aspire-communitytoolkit/src/CommunityToolkit.Aspire.Hosting.K3s/K3sInProcessPortForwarder.cs:line 147
Service podinfo/podinfo not yet ready; retrying…
k8s.Autorest.HttpOperationException: Operation returned an invalid status code 'NotFound', response body {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"services \"podinfo\" not found","reason":"NotFound","details":{"name":"podinfo","kind":"services"},"code":404}

   at k8s.Kubernetes.SendRequestRaw(String requestContent, HttpRequestMessage httpRequest, CancellationToken cancellationToken) in /_/src/KubernetesClient/Kubernetes.cs:line 180
   at k8s.AbstractKubernetes.ICoreV1Operations_ReadNamespacedServiceWithHttpMessagesAsync[T](String name, String namespaceParameter, Nullable`1 pretty, IReadOnlyDictionary`2 customHeaders, CancellationToken cancellationToken) in /_/src/KubernetesClient/obj/Release/net10.0/generated/LibKubernetesGenerator/LibKubernetesGenerator.KubernetesClientSourceGenerator/CoreV1Operations.g.cs:line 4018
   at k8s.AbstractKubernetes.k8s.ICoreV1Operations.ReadNamespacedServiceWithHttpMessagesAsync(String name, String namespaceParameter, Nullable`1 pretty, IReadOnlyDictionary`2 customHeaders, CancellationToken cancellationToken) in /_/src/KubernetesClient/obj/Release/net10.0/generated/LibKubernetesGenerator/LibKubernetesGenerator.KubernetesClientSourceGenerator/CoreV1Operations.g.cs:line 4028
   at k8s.CoreV1OperationsExtensions.ReadNamespacedServiceAsync(ICoreV1Operations operations, String name, String namespaceParameter, Nullable`1 pretty, CancellationToken cancellationToken) in /_/src/KubernetesClient/obj/Release/net10.0/generated/LibKubernetesGenerator/LibKubernetesGenerator.KubernetesClientSourceGenerator/CoreV1OperationsExtensions.g.cs:line 24966
   at CommunityToolkit.Aspire.Hosting.K3sInProcessPortForwarder.WaitForServiceReadyAsync(ILogger logger, CancellationToken ct) in /workspaces/aspire-communitytoolkit/src/CommunityToolkit.Aspire.Hosting.K3s/K3sInProcessPortForwarder.cs:line 147
```

It just keeps repeating that. Is there any additional things that need to be installed to run it?

Edit: I left it running for ~30 minutes. The resource started but never became healthy - same error as above in the logs.

aaronpowell:
I also tried running the .NET app host and had the same outcome (although I only let that one run for 10 minutes 🤣)

edmondshtogu:
@aaronpowell I found the issue! 

On native Linux, Docker binds files using their `inode` number on start time, not their path name. `EnsureKubeconfigPlaceholder` created an empty placeholder file (`inode X`), which the `Helm`/`kubectl` containers bind-mount as `/tmp/k3s-kubeconfig.yaml`. When `K3sReadinessHealthCheck` later ran `WriteAtomicAsync` (write to temp → `File.Move` → rename), that `rename()` created a new `inode Y` at the same path. The containers kept tracking the old empty `inode X`, so `[ -f /tmp/k3s-kubeconfig.yaml ]` was true but the file was always empty, therefor helm install never run, leaving the service endpoint in waiting state. Additionally I added a timeout as the safety net for helm/manifest health check to prevent hanging indefinitely.

In my macOS it worked as I have configuring Colima with `sshfs` which resolves bind-mounted files by path (not inode), so the rename was visible inside containers.

Tests in `codespaces` are now passing!

edmondshtogu:
K3s tests passed 🎉

aaronpowell (approved):

PR #1370, opened 2026-05-27
Add Aspire integration analyzer

sebastienros:
This enables the Aspire hosting integration analyzer so CommunityToolkit integrations are continuously checked for polyglot-compatible exported APIs.

The change wires `Aspire.Hosting.Integration.Analyzers` from the dotnet9 feed into integration projects as a private analyzer dependency, then updates exported hosting APIs to satisfy the analyzer. The cleanup removes redundant export IDs, keeps `ats-summary` documentation only where summaries contain C# XML tags, marks synchronous callback exports for background-thread execution, and explicitly exports or ignores members that affect the generated polyglot surface.

A few integrations needed non-obvious handling: C# extension-block syntax is preserved for Flyway and PostgreSQL Flyway helpers, with a narrow suppression for the analyzer diagnostic that currently treats exported extension-block members as non-static.


src/CommunityToolkit.Aspire.Hosting.Ollama/OllamaUtilities.cs
```diff
@@ -7,8 +8,11 @@
 
 namespace CommunityToolkit.Aspire.Hosting.Ollama;
 
+#pragma warning disable ASPIREATS001 // AspireExport is experimental
+
 internal static class OllamaUtilities
 {
+    [AspireExportIgnore(Reason = "Internal helper returns a tuple and accepts CancellationToken; it is not part of the polyglot surface.")]
```
aaronpowell, line 15: Won't the fact this is on an `internal` class mean that it won't be exported?

src/CommunityToolkit.Aspire.Hosting.OpenTelemetryCollector/OpenTelemetryCollectorRoutingExtensions.cs
```diff
@@ -17,7 +17,7 @@ public static class OpenTelemetryCollectorRoutingExtensions
     /// <param name="builder">The resource builder.</param>
     /// <param name="collectorBuilder">The OpenTelemetry Collector resource builder.</param>
     /// <returns>A reference to the resource builder.</returns>
-    [AspireExport("withOpenTelemetryCollectorRouting", Description = "Routes telemetry for a resource through the specified OpenTelemetry Collector")]
+    [AspireExport("withOpenTelemetryCollectorOpenTelemetryCollectorRouting")]
```
aaronpowell: This exported method name looks pretty clunky, does it need `OpenTelemetryCollector` in there twice?

src/CommunityToolkit.Aspire.Hosting.Zitadel/ZitadelHostingExtensions.cs
```diff
@@ -14,13 +14,14 @@ public static class ZitadelHostingExtensions
     /// <summary>
     /// Adds a Zitadel container resource to the <see cref="IDistributedApplicationBuilder"/>.
     /// </summary>
+    /// <ats-summary>Adds a Zitadel container resource</ats-summary>
```
aaronpowell, line 17: I think this is the only usage of `ats-summary`. I assume that's to make something appear for the polyglot apphosts, but why is it only on this method?

src/CommunityToolkit.Aspire.Hosting.Zitadel/ZitadelHostingExtensions.cs
```diff
@@ -183,11 +184,12 @@ public static IResourceBuilder<ZitadelResource> WithDatabase(this IResourceBuild
     /// <summary>
     /// Configures the external domain for the Zitadel resource. This overrides the default domain set in <see cref="AddZitadel"/>.
     /// </summary>
+    /// <ats-summary>Configures the external domain for the Zitadel resource</ats-summary>
```
aaronpowell, line 187: Ok, there was a second usage 😅

aaronpowell (commented):
Some comments, and the branch should probably target the 13.4 upgrade branch so we keep it all in sync.

src/CommunityToolkit.Aspire.Hosting.Ollama/OllamaUtilities.cs
sebastienros: The analyzer detects it as an exported extension, it's likely a bug in the analyzer but until it's fixed we need to ignore the method.

src/CommunityToolkit.Aspire.Hosting.Zitadel/ZitadelHostingExtensions.cs
sebastienros: The idea is that if the standard `<summary>` contains special C# tags (`<see cref`) or type names (generics) then we need to provide one for polyglot. There is the equivalent for params and return values.

aaronpowell (approved):

nuget.config
```diff
@@ -12,6 +12,9 @@
     <packageSource key="dotnet-eng">
       <package pattern="Microsoft.DotNet.XUnitV3Extensions" />
     </packageSource>
+    <packageSource key="dotnet9">
+      <package pattern="Aspire.Hosting.Integration.Analyzers" />
+    </packageSource>
```
aaronpowell:
```suggestion
```

PR #1382, opened 2026-06-01
Simplified Aspire name resolution

sliekens:
Since there is only one Aspire used throughout this codebase, it is unnecessary to reference it using its fully qualified name. The compiler was already able to resolve Aspire unambiguously, making the additional qualification redundant. The longer form was technically correct but unnecessarily verbose. Additional qualification was found to be informationally redundant. The surrounding context already establishes the intended Aspire.


Omnideth (commented):
Ship it.  Looks good.

I was nervous to leave it as an approve review instead of a comment review. 

An approve review from me would still require a core maintainer to actually release, correct?

afscrome (approved):

PR #1431, opened 2026-06-25
test(SurrealDb): bump VerifySurrealDbResource and WithDataShouldPersistStateBetweenUsages timeouts to 10 minutes

tamirdresher:
## Why

The post-merge `.NET main` run for #1394 hit a SurrealDB test failure (run [28141532703](https://github.com/CommunityToolkit/Aspire/actions/runs/28141532703), attempt 2): `SurrealDbFunctionalTests.WithDataShouldPersistStateBetweenUsages(useVolume: False)` cancelled at **5m 03s** with the resource already reporting:

```
- Current State: Running
- Current Health: Healthy
- - - - System.OperationCanceledException : The operation was canceled.
```

i.e. the healthy event arrived right as the 5-minute CTS tripped.

## Root cause

`WithDataShouldPersistStateBetweenUsages` runs **two full app lifecycles back to back** (start surreal → wait healthy → write data → stop, then build a new app → start → wait healthy → read data). On a cold runner the first SurrealDB image pull + container start + healthy check eats most of the 5-minute budget, leaving no headroom for the second app. The other slow tests in this file (`VerifyWaitForOnSurrealDbBlocksDependentResources`, `VerifyWaitForOnSurrealDbBlocksDependentResourcesUntilCancellation`) already use `FromMinutes(10)` -- aligning these two with that.

## What changes

Two-line change:

| Test                                              | Before        | After          |
|---------------------------------------------------|---------------|----------------|
| `VerifySurrealDbResource`                         | `FromMinutes(5)`  | `FromMinutes(10)`  |
| `WithDataShouldPersistStateBetweenUsages`         | `FromMinutes(5)`  | `FromMinutes(10)`  |

Leaves the `[Fact(Skip = ...)]` test (`AddDatabaseCreatesDatabaseWithCustomScript`) at 5 minutes since it doesn't run.

Aaron mentioned image caching is supposed to be in place; if it later proves to be working, the test will simply finish well under the new 10-minute ceiling -- this purely raises the upper bound, no behavior change.

cc @aaronpowell

### DIFF
--- a/tests/CommunityToolkit.Aspire.Hosting.SurrealDb.Tests/SurrealDbFunctionalTests.cs
+++ b/tests/CommunityToolkit.Aspire.Hosting.SurrealDb.Tests/SurrealDbFunctionalTests.cs
@@ -36,7 +36,7 @@ public class SurrealDbFunctionalTests(ITestOutputHelper testOutputHelper)
     [Fact]
     public async Task VerifySurrealDbResource()
     {
-        using var cts = new CancellationTokenSource(TimeSpan.FromMinutes(5));
+        using var cts = new CancellationTokenSource(TimeSpan.FromMinutes(10));
         var ct = cts.Token;
         await using var builder = TestDistributedApplicationBuilder.Create(testOutputHelper);
 
@@ -74,7 +74,7 @@ public class SurrealDbFunctionalTests(ITestOutputHelper testOutputHelper)
     [InlineData(false)]
     public async Task WithDataShouldPersistStateBetweenUsages(bool useVolume)
     {
-        using var cts = new CancellationTokenSource(TimeSpan.FromMinutes(5));
+        using var cts = new CancellationTokenSource(TimeSpan.FromMinutes(10));
         var ct = cts.Token;
 
         string? volumeName = null;


aaronpowell:
I've done some investigating into this more deeply and I think that there's a mistaken assumption on the test failure here being due to the timeout and not that there's an underlying problem.

Looking at sibling tests in the class that `WithDataShouldPersistStateBetweenUsages` is in, we can see they run in ~10s (`VerifyWaitForOnSurrealDbBlocksDependentResources` in [this run](https://github.com/CommunityToolkit/Aspire/actions/runs/28350517832/job/83982335574?pr=1439) has a duration of `00:00:09.8701816` per the TRX file. Given that, extending the timeline isn't going to resolve the problem, as we see in the test run of this PR.

With the above mentioned test run I have expanded the amount of logging that we do in these kinds of tests, and it looks like the root cause is that it fails to create the namespace because it's unable to connect to the remote server - the websocket connection fails. If the namespace isn't created because it can't connect to the server we have a cascade of problems which results in the resource never becoming healthy and eventually the test hits the timeout and is terminated.

Reviewing the tests in this test class, I actually don't believe we should have them in at all, the tests are really testing Aspire core behaviour, not really the behaviour of the integration - at least not in a way that isn't already covered by our integration test stack using the example app host. Let's take the method `WithDataShouldPersistStateBetweenUsages` for example, this test is verifying that the mounted container volume will be persisted or not by Aspire between runs, depending on whether we call `WithDataVolume`. This test is not something that we should have in our repo because the behaviour of `WithDataVolume` is just a wrapper around `WithVolume` from Aspire, and if `WithVolume` fails to persist a volume with the underlying container runtime, there's nothing that we can do in the Toolkit to resolve it. We should have unit tests that verify that we have the correct annotations set (which it doesn't look like we do have), but this test as it stands, and the others in the test class, shouldn't be in our repo.

aaronpowell:
Closing in favour of #1441

PR #1467, opened 2026-07-12
Ensure SQLite Web always pulls latest image

oondriss:
## Summary

- always pull the SQLite Web container image before starting the resource
- verify the pull policy in the SQLite resource model tests
- verify the SQLite Web resource becomes healthy and serves its UI

## Root cause

The SQLite integration has to use `ghcr.io/coleifer/sqlite-web:latest` because the image does not publish versioned tags. The resource currently uses the container runtime's default pull policy, which can indefinitely reuse a cached `latest` image.

Older cached images can have a different startup contract from the current image. In particular, an image without an entrypoint receives the database filename from Aspire as its command and fails with `exec: "db.sqlite": executable file not found in $PATH`.

Using `ImagePullPolicy.Always` keeps the mutable image tag and its startup contract in sync for every `WithSqliteWeb()` consumer.

## Validation

```text
dotnet test tests/CommunityToolkit.Aspire.Hosting.Sqlite.Tests/CommunityToolkit.Aspire.Hosting.Sqlite.Tests.csproj --no-restore -- --filter-not-class CommunityToolkit.Aspire.Hosting.Sqlite.Tests.TypeScriptAppHostTests --output Normal

Passed: 17
Failed: 0
```

The TypeScript AppHost test was excluded locally because the installed Aspire CLI does not provide the repository's expected `aspire restore` command. It remains covered by CI.

### DIFF
--- a/src/CommunityToolkit.Aspire.Hosting.Sqlite/SqliteResourceBuilderExtensions.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.Sqlite/SqliteResourceBuilderExtensions.cs
@@ -80,6 +80,7 @@ public static class SqliteResourceBuilderExtensions
         var resourceBuilder = builder.ApplicationBuilder.AddResource(resource)
                                 .WithImage(SqliteContainerImageTags.SqliteWebImage, SqliteContainerImageTags.SqliteWebTag)
                                 .WithImageRegistry(SqliteContainerImageTags.SqliteWebRegistry)
+                                .WithImagePullPolicy(ImagePullPolicy.Always)
                                 .WithHttpEndpoint(targetPort: 8080, name: "http")
                                 .WithArgs(builder.Resource.DatabaseFileName)
                                 .WithBindMount(builder.Resource.DatabasePath, "/data")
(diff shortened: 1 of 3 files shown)


oondriss:
@dotnet-policy-service agree

Odonno:
Why not use the last released tag that is `0.7.2`?

oondriss:
> Why not use the last released tag that is `0.7.2`?

i was following comment on the code here:

https://github.com/CommunityToolkit/Aspire/blob/d7b140891317fccdf1188dbbeccd688206c94cec/src/CommunityToolkit.Aspire.Hosting.Sqlite/SqliteContainerImageTags.cs#L6

and the integration was already using latest, so I don't want to lock it in current release

Odonno:
It looks like a mistake, we should always pin to a release tag. cc. @aaronpowell

aaronpowell:
Context: When I created the integration (and added the web container support) the project wasn't providing tagged released, it was only providing the `latest` tag - hence the comment in the code.

Looking at history, that was done on the 15th Jan 2025 and then looking at the sqlite-web project, it would appear that, as of the 16th Jan 2025 they started publishing tagged releases 🤣.

Given that, we should be updating so that the container image tags for SqliteWeb is pinned to the latest, which is 0.7.2, and not have the "always pull latest" logic in there.

Odonno:
Didn't we go for a specific pinned tag?

PR #1468, opened 2026-07-14
Fix build warnings treated as errors across the solution

DavidGarton8:
When the project is first opened, several warnings appear across the solution, which is noisy and can be distracting with `TreatWarningsAsErrors` enabled. I wanted to clean up all of these warnings in one pass rather than open a series of small PRs, but I’m happy to split or adjust this if that would be preferred.

## Changes
- Added missing XML documentation comments
- Removed a stray semicolon
- Fixed a nullability warning in tests
- Suppressed an intentional obsolete warning

## PR Checklist

- [ ] New integration
  - [ ] Docs are written


jmezach:
@aaronpowell It looks like the recently added integration codeowner workflow is failing here, but I'm not sure why. It's complaining about `core` being already defined, but I don't see it. Any ideas?

DavidGarton8:
Core is already defined by GitHub script itself. I don't think you need to declare it again.

https://github.com/actions/github-script/blob/main/src/main.ts

Odonno (approved):

aaronpowell:
Sorry, stuffed that script when I wrote it and fixed ages ago, just need the branch to sync

PR #1479, opened 2026-07-23
feat: `CommunityToolkit.Aspire.Hosting.Floci` — Floci cloud emulator hosing integration

### LINKED ISSUE #1242: [FEAT] Add Aspire hosting support (Aspire.Hosting.Floci)
## Service

Ecosystem Integration / Tooling

## API Action / Feature

Aspire Hosting Package `Aspire.Hosting.Floci` (for Distributed App Hosts)

## AWS Documentation

N/A for AWS — related to [Aspire Documentation](https://aspire.dev/)

## Why is this needed?

[Aspire](https://aspire.dev/) has evolved into a code-first, polyglot toolchain for building, running, and deploying distributed applications. It uses an `AppHost` project (in C# or TypeScript) to orchestrate local development containers, cloud resources, and polyglot services (Node.js, Python, Go, Java, Rust, etc.) with automatic service discovery.

Currently, Floci has excellent module support across Java, Python, and Go via Testcontainers. Adding dedicated hosting integrations for Aspire would allow developers to natively spin up Floci resources directly from an Aspire AppHost and inject the emulated AWS endpoints seamlessly into their polyglot microservices.

**Expected Developer Experience:**
Without this, developers have to manually configure raw Docker containers or write custom integration wrappers inside Aspire to use Floci. A native extension method (like `builder.AddFloci()` in C# or `builder.addFloci()` in TypeScript) would:

1. Automatically spin up the `floci/floci:latest` emulator via the Aspire control plane.
2. Configure the connection strings and endpoints for emulated AWS resources (S3, DynamoDB, SQS).
3. Automatically inject those environment variables into consuming projects, regardless of whether the target microservice is written in C#, Go, Python, or TypeScript.

This would create a flawless local development loop for AWS emulation across the entire Aspire polyglot ecosystem.

## Are you willing to contribute a PR?

* [x] Yes
* [ ] No

edmondshtogu:
**Closes floci-io/floci#1242**

Adds `CommunityToolkit.Aspire.Hosting.Floci` — a hosting integration for [Floci](https://floci.io/), a family of high-performance local cloud emulators (`floci/floci` for AWS, `floci/floci-az` for Azure, `floci/floci-gcp` for GCP) that run as containers inside the Aspire AppHost.

#### Core integration (`src/CommunityToolkit.Aspire.Hosting.Floci/`)

- **`AddFlociAws(name, port?, defaultRegion?, defaultAccountId?)`** — adds the Floci AWS container with an HTTP health check at `/_floci/info`
- **`AddFlociAzure(name, port?)`** — adds the Floci Azure container (`floci/floci-az`), health check at `/_floci/health`
- **`AddFlociGcp(name, port?, defaultProjectId?)`** — adds the Floci GCP container (`floci/floci-gcp`), health check at `/_floci-gcp/health`
- All three cloud resources share a common abstract base (`FlociContainerResource`) that holds endpoint/connection-string plumbing, so `Host`, `Port`, `PrimaryEndpoint`, and `ConnectionStringExpression` are implemented once and inherited — not duplicated per cloud
- **`WithReference`** automatically injects the standard `ConnectionStrings__<name>` plus provider-specific env vars via a `BeforeStartEvent` subscriber; host processes receive `localhost:{port}`, containers receive `host.docker.internal:{port}`:
  - AWS: `AWS_ENDPOINT_URL`, `AWS_DEFAULT_REGION`, `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`
  - Azure: `AZURE_STORAGE_CONNECTION_STRING` (well-known `devstoreaccount1` dev credentials)
  - GCP: `PUBSUB_EMULATOR_HOST`, `FIRESTORE_EMULATOR_HOST`, `DATASTORE_EMULATOR_HOST`, `STORAGE_EMULATOR_HOST`, `SECRET_MANAGER_EMULATOR_HOST`, `GOOGLE_CLOUD_PROJECT`, `CLOUDSDK_CORE_PROJECT`
- **`WithDataVolume` / `WithDataBindMount`** — switches Floci from in-memory to persistent storage mode; implemented once and shared across all three clouds
- **`WithDockerSocket`** — mounts the Docker socket so Lambda / Azure Functions / Cloud Run and other container-backed services can launch sibling containers; adds `--add-host=host.docker.internal:host-gateway` to container dependents; shared across all three clouds
- **`WithConfigFile(hostPath)`** — bind-mounts a Quarkus `application.yml` override at `/deployments/config/application.yml` for advanced configuration (AWS only — Azure/GCP images don't expose this)

#### Floci UI web console (`WithFlociUI` / `WithPluggedCloud`)

- **`WithFlociUI(configureContainer?, containerName?)`** — available on all three cloud resource types, adds a single shared `FlociUIContainerResource` running `docker.io/floci/floci-ui:0.2.0`:
  - Wired to the creating cloud's endpoint over the container network (`FLOCI_ENDPOINT` for AWS, `FLOCI_AZURE_ENDPOINT` for Azure, `FLOCI_GCP_ENDPOINT` for GCP)
  - HTTP health check on `/`; excluded from manifest (`ExcludeFromManifest()`)
  - Idempotent per creating cloud resource: calling `WithFlociUI` twice re-configures the existing UI container instead of adding a duplicate
- **`WithReference(cloud)`** — attaches an *additional* cloud resource to an already-created UI console, so one console can browse AWS + Azure + GCP at once instead of spinning up three separate UI containers. In C# this is a single overloaded method name; the generated TypeScript bindings expose it as `withAwsReference` / `withAzureReference` / `withGcpReference` since the polyglot export generator needs distinct names when all three overloads share the same target type
- **`WithHostPort(int?)`** on the UI resource builder to pin the host port
- image

#### Examples and tests

- C# and TypeScript AppHosts each provision all three clouds plus a single shared UI console (`aws.WithFlociUI(configureContainer: ui => { ui.WithPluggedCloud(azure); ui.WithPluggedCloud(gcp); })`)
- The example `ApiService` exercises all three clouds end-to-end: S3 (AWS SDK), Blob Storage (Azure SDK), and Cloud Storage (GCP SDK), each with its own health check (`floci-s3`, `floci-azure-blob`, `floci-gcp-storage`) and a hosted service that creates a demo bucket/container on startup
- 41 tests across 6 test classes: resource creation/env-var coverage for each cloud (AWS/Azure/GCP), the shared multi-cloud UI (`WithPluggedCloud` attaches correctly, single UI container, correct env vars per attached cloud), the TypeScript AppHost, and Docker-gated AppHost integration tests asserting all three cloud resources *and* the UI resource become healthy
- README with C#/TS examples for every extension method across all three clouds, including the standard `ConnectionStrings__<name>` behavior alongside each cloud's provider-specific env vars

## PR Checklist

- [x] New integration
  - [x] Docs are written

## Other information

* This integration does not support TLS — none of the three Floci images (`floci/floci`, `floci/floci-az`, `floci/floci-gcp`) expose certificate-based configuration the way the earlier AWS-only prototype assumed, so that support was removed rather than shipped half-working
* `AWS_ENDPOINT_URL` is the standard SDK v4 env var; no `AmazonS3Config.RegionEndpoint` should be set alongside `ServiceURL` (triggers a `NullReferenceException` in the v4 endpoint rule engine — documented in the example)
* GCP's `StorageClient` is built with `EmulatorDetection.EmulatorOnly` so it reads `STORAGE_EMULATOR_HOST` and skips real GCP credential resolution entirely
* The `ASPIREATS001` experimental diagnostic (polyglot export attributes) is suppressed at file level in the hosting extension files


edmondshtogu:
@aaronpowell in addition to this feature I had to fix the restore issue for the security vulnerabilities caused by `System.Security.Cryptography.Xml` version `8.0.3` (by overriding the version with `8.0.4`).

The PR is ready for your review, so please take a look and let me know if you have any concern.

edmondshtogu:
@aaronpowell can you take a look please?

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Aws.cs
```diff
@@ -0,0 +1,180 @@
...
+                    dependent.Annotations.Add(
+                        new ContainerRuntimeArgsCallbackAnnotation(
+                            args => args.Add("--add-host=host.docker.internal:host-gateway")));
+                    dependent.Annotations.Add(new EnvironmentCallbackAnnotation(ctx =>
+                    {
+                        ctx.EnvironmentVariables["AWS_ENDPOINT_URL"] =
+                            ReferenceExpression.Create($"http://host.docker.internal:{flociPort}");
+                        ctx.EnvironmentVariables["AWS_DEFAULT_REGION"] = resource.DefaultRegion;
+                        ctx.EnvironmentVariables["AWS_ACCESS_KEY_ID"] = "test";
+                        ctx.EnvironmentVariables["AWS_SECRET_ACCESS_KEY"] = "test";
```
aaronpowell: Are these environment variable names a required design for the AWS SDKs, or is it just the approach taken here?

If they are just this way for the integration, we should use the .NET config hierarchy style, so `Floci__AWS__EndpointUrl`, so we can resolve them using the .NET config system of `builder.Configuration.GetValue<string>("Floci:AWS:EndpointUrl")`, being consistent with other integrations.

src/CommunityToolkit.Aspire.Hosting.Floci/CommunityToolkit.Aspire.Hosting.Floci.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>An Aspire component leveraging the Floci AWS emulator container. Includes a WithFlociUI() extension for running the Floci UI web console alongside the emulator.</Description>
+    <AdditionalPackageTags>floci aws localstack emulator hosting ui</AdditionalPackageTags>
```
aaronpowell: Given there's provisions here for Azure and GCP, is the description/tags accurate?

src/CommunityToolkit.Aspire.Hosting.Floci/FlociAwsContainerResource.cs
```diff
@@ -0,0 +1,50 @@
+#pragma warning disable ASPIREATS001 // AspireExport is experimental
```
aaronpowell: Not required anymore

src/CommunityToolkit.Aspire.Hosting.Floci/FlociAzureContainerResource.cs
```diff
@@ -0,0 +1,30 @@
+#pragma warning disable ASPIREATS001 // AspireExport is experimental
```
aaronpowell: Not required anymore

src/CommunityToolkit.Aspire.Hosting.Floci/FlociContainerImageTags.cs
```diff
@@ -0,0 +1,20 @@
+namespace CommunityToolkit.Aspire.Hosting.Floci;
+
+internal static class FlociContainerImageTags
+{
+    public const string AwsRegistry = "docker.io";
+    public const string AwsImage = "floci/floci";
+    public const string AwsTag = "latest"; // Consider pinning to a specific version tag for reproducible builds/runs (or document why only `latest` is available).
```
aaronpowell: Specify an image version not `latest` to avoid images drifting from tested versions

src/CommunityToolkit.Aspire.Hosting.Floci/FlociContainerImageTags.cs
```diff
@@ -0,0 +1,20 @@
...
+
+    public const string AzureRegistry = "docker.io";
+    public const string AzureImage = "floci/floci-az";
+    public const string AzureTag = "latest";
```
aaronpowell: Specify an image version not `latest` to avoid images drifting from tested versions

src/CommunityToolkit.Aspire.Hosting.Floci/FlociContainerImageTags.cs
```diff
@@ -0,0 +1,20 @@
...
+
+    public const string GcpRegistry = "docker.io";
+    public const string GcpImage = "floci/floci-gcp";
+    public const string GcpTag = "latest";
```
aaronpowell: Specify an image version not `latest` to avoid images drifting from tested versions

src/CommunityToolkit.Aspire.Hosting.Floci/FlociContainerResource.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// <summary>
+    /// Gets the port endpoint reference for the primary endpoint.
+    /// </summary>
+    public EndpointReferenceExpression Port => PrimaryEndpoint.Property(EndpointProperty.Port);
+
+    /// <summary>
+    /// Gets the emulator endpoint URL.
+    /// </summary>
+    public ReferenceExpression ConnectionStringExpression =>
+        ReferenceExpression.Create($"http://{Host}:{Port}");
```
aaronpowell, line 56: `http` only?

src/CommunityToolkit.Aspire.Hosting.Floci/FlociGcpContainerResource.cs
```diff
@@ -0,0 +1,33 @@
+#pragma warning disable ASPIREATS001 // AspireExport is experimental
```
aaronpowell: Not required anymore

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Aws.cs
```diff
@@ -0,0 +1,180 @@
+using Aspire.Hosting.ApplicationModel;
+using CommunityToolkit.Aspire.Hosting.Floci;
+using Microsoft.Extensions.DependencyInjection;
+
+#pragma warning disable ASPIREATS001 // AspireExport is experimental
```
aaronpowell: Not required anymore

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Aws.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        FlociAwsContainerResource resource = new(name) { DefaultRegion = defaultRegion, DefaultAccountId = defaultAccountId };
+
+        // BeforeStartEvent: inject standard AWS env vars into every resource that called
+        // WithReference(floci). Standard WithReference already injects ConnectionStrings__<name>;
+        // this subscriber additionally sets AWS_ENDPOINT_URL (and companions) so the AWS SDK
+        // needs no extra configuration in dependent services.
+        //
+        // Processing is deferred to BeforeStartEvent so resources can be wired up in any order
+        // in Program.cs without worrying about whether Floci is fully configured yet.
```
aaronpowell: Why wouldn't you create a method `WithReference(this IResourceBuilder<IResourceWithEnvironment> builder, IResourceBuilder<FlociAwsContainerResource> floci)` that then adds them using the `WithEnvironment` callback?

This would optimise the execution pipeline because you no longer have to iterate over ever resource in the app model and find the ones that have a relationship to this resource.

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Azure.cs
```diff
@@ -0,0 +1,162 @@
+using Aspire.Hosting.ApplicationModel;
+using CommunityToolkit.Aspire.Hosting.Floci;
+using Microsoft.Extensions.DependencyInjection;
+
+#pragma warning disable ASPIREATS001 // AspireExport is experimental
```
aaronpowell: Not required anymore

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Azure.cs
```diff
@@ -0,0 +1,162 @@
...
+        int? port = null)
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentNullException.ThrowIfNull(name);
+
+        FlociAzureContainerResource resource = new(name);
+
+        // BeforeStartEvent: inject an Azure Storage connection string into every resource that
+        // called WithReference(floci). Deferred so resources can be wired up in any order in
+        // Program.cs without worrying about whether Floci is fully configured yet.
```
aaronpowell: Same observation as with AWS, this could be done with a custom `WithReference` method

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Azure.cs
```diff
@@ -0,0 +1,162 @@
...
+                {
+                    // Containers cannot reach the host via localhost — use host.docker.internal
+                    // so they can reach the host-exposed Floci port (4577).
+                    var flociPort = resource.Port;
+                    dependent.Annotations.Add(
+                        new ContainerRuntimeArgsCallbackAnnotation(
+                            args => args.Add("--add-host=host.docker.internal:host-gateway")));
+                    dependent.Annotations.Add(new EnvironmentCallbackAnnotation(ctx =>
+                    {
+                        var blobEndpoint = ReferenceExpression.Create($"http://host.docker.internal:{flociPort}/{FlociAzureContainerResource.DefaultAccountName}");
```
aaronpowell: Will this work with non-Docker container hosts?

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Common.cs
```diff
@@ -0,0 +1,66 @@
...
+    internal static IResourceBuilder<TFloci> WithDockerSocketCore<TFloci>(
+        IResourceBuilder<TFloci> builder,
+        string socketPath)
+        where TFloci : FlociContainerResource
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentNullException.ThrowIfNull(socketPath);
+
+        return builder
+            .WithEnvironment(builder.Resource.DockerHostEnvVar, $"unix://{ContainerSocketPath}")
```
aaronpowell, line 75: Is that going to be valid on Windows?

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.FlociUI.cs
```diff
@@ -0,0 +1,160 @@
+using Aspire.Hosting.ApplicationModel;
+using CommunityToolkit.Aspire.Hosting.Floci;
+
+#pragma warning disable ASPIREATS001 // AspireExport is experimental
```
aaronpowell: Not required

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.FlociUI.cs
```diff
@@ -0,0 +1,160 @@
...
+
+    /// <summary>
+    /// Attaches an additional Floci AWS emulator resource to an existing Floci UI console, so a
+    /// single console can browse multiple clouds at once instead of creating a UI container per cloud.
+    /// </summary>
+    /// <param name="builder">The Floci UI resource builder (from the <c>configureContainer</c> callback of <c>WithFlociUI</c>).</param>
+    /// <param name="cloud">The additional Floci AWS resource to attach.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{FlociUIContainerResource}"/> for further configuration.</returns>
+    [AspireExport("withPluggedCloudAws")]
+    public static IResourceBuilder<FlociUIContainerResource> WithPluggedCloud(
```
aaronpowell: Is "Plugged" a bit of terminology that's common in the Floci lexicon? Otherwise, wouldn't `WithFerence` be doing the job here and be consistent with Aspire terminology?

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Gcp.cs
```diff
@@ -0,0 +1,175 @@
+using Aspire.Hosting.ApplicationModel;
+using CommunityToolkit.Aspire.Hosting.Floci;
+using Microsoft.Extensions.DependencyInjection;
+
+#pragma warning disable ASPIREATS001 // AspireExport is experimental
```
aaronpowell: Not required

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Gcp.cs
```diff
@@ -0,0 +1,175 @@
...
+        string defaultProjectId = "floci-local")
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentNullException.ThrowIfNull(name);
+
+        FlociGcpContainerResource resource = new(name) { DefaultProjectId = defaultProjectId };
+
+        // BeforeStartEvent: inject the *_EMULATOR_HOST vars the GCP SDKs already honor into every
+        // resource that called WithReference(floci). Deferred so resources can be wired up in any
+        // order in Program.cs without worrying about whether Floci is fully configured yet.
```
aaronpowell: Like with the AWS comment, could a custom `WithReference` simplify this

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Gcp.cs
```diff
@@ -0,0 +1,175 @@
...
+                    // Containers cannot reach the host via localhost — use host.docker.internal
+                    // so they can reach the host-exposed Floci port (4588).
+                    var flociPort = resource.Port;
+                    dependent.Annotations.Add(
+                        new ContainerRuntimeArgsCallbackAnnotation(
+                            args => args.Add("--add-host=host.docker.internal:host-gateway")));
+                    dependent.Annotations.Add(new EnvironmentCallbackAnnotation(ctx =>
+                    {
+                        var hostAndPort = ReferenceExpression.Create($"host.docker.internal:{flociPort}");
+                        ctx.EnvironmentVariables["PUBSUB_EMULATOR_HOST"] = hostAndPort;
```
aaronpowell: Environment variables comment here along the lines of the AWS one - it'd be best to use the .NET config key hierarchy structure if we can.

src/CommunityToolkit.Aspire.Hosting.Floci/FlociUIContainerResource.cs
```diff
@@ -0,0 +1,49 @@
+#pragma warning disable ASPIREATS001 // AspireExport is still experimental
```
aaronpowell: Not required

tests/CommunityToolkit.Aspire.Hosting.Floci.Tests/AppHostTests.cs
```diff
@@ -0,0 +1,46 @@
+using Aspire.Components.Common.Tests;
+using CommunityToolkit.Aspire.Testing;
+
+namespace CommunityToolkit.Aspire.Hosting.Floci.Tests;
+
+[RequiresDocker]
+public class AppHostTests(AspireIntegrationTestFixture<Projects.CommunityToolkit_Aspire_Hosting_Floci_AppHost> fixture)
```
aaronpowell, line 10: Tests without asserts always bother me because there is a risk that you're not observing any action within the test.

Yes, if the resource fails to become healthy within that time it will result in a failure to start and the test will timeout and exit, resulting in a failure, but is there anything we can observe on the resource just in case? Even a 200 OK response from `/`

src/CommunityToolkit.Aspire.Hosting.Floci/FlociAwsContainerResource.cs
edmondshtogu: Removed from all files!

src/CommunityToolkit.Aspire.Hosting.Floci/FlociGcpContainerResource.cs
edmondshtogu: Removed from all files!

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Aws.cs
edmondshtogu: Removed from all files!

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.FlociUI.cs
edmondshtogu: Removed from all files!

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Gcp.cs
edmondshtogu: Removed from all files!

src/CommunityToolkit.Aspire.Hosting.Floci/FlociUIContainerResource.cs
edmondshtogu: Removed from all files!

src/CommunityToolkit.Aspire.Hosting.Floci/FlociContainerImageTags.cs
edmondshtogu: Image tags pined!

src/CommunityToolkit.Aspire.Hosting.Floci/FlociContainerImageTags.cs
edmondshtogu: Image tags pined!

src/CommunityToolkit.Aspire.Hosting.Floci/FlociContainerImageTags.cs
edmondshtogu: Image tags pined!

src/CommunityToolkit.Aspire.Hosting.Floci/CommunityToolkit.Aspire.Hosting.Floci.csproj
edmondshtogu: Fixed, thanks for pointing out as I missed once added support for all three providers.

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Aws.cs
edmondshtogu: The AWS SDK expects these variables. See: [docs](https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-envvars.html#envvars-set)!

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.FlociUI.cs
edmondshtogu: Renamed!

src/CommunityToolkit.Aspire.Hosting.Floci/FlociContainerResource.cs
edmondshtogu: Added back. `floci/floci` and `floci/floci-az` both support TLS (`FLOCI_TLS_*` / `FLOCI_AZ_TLS_*`), serving HTTP and HTTPS simultaneously on the same port. The integration now hooks Aspire's certificate plumbing via `WithHttpsCertificateConfiguration`, so `WithHttpsDeveloperCertificate()` / `WithHttpsCertificate(...)` just work. `ConnectionStringExpression` now follows `EndpointProperty.Scheme`. Plain HTTP stays the default (opt-in, so an existing AppHost doesn't silently flip). `floci/floci-gcp` genuinely has no HTTPS listener — no `FLOCI_GCP_TLS*` anywhere, and Quarkus SSL config makes it `exit(1)`, so nothing is registered for it.

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Aws.cs
edmondshtogu: Done. Each cloud now has a `WithReference<TDestination>(this IResourceBuilder<TDestination>, IResourceBuilder<Floci…ContainerResource>)` overload that chains Aspire's built-in `WithReference` (so `ConnectionStrings__{name}` and its connection properties still land) and adds the provider env vars via `WithEnvironment`. All three `BeforeStartEvent` subscribers and the per-resource scans are gone. Polyglot exports are `withFlociAwsReference` / `withFlociAzureReference` / `withFlociGcpReference`.

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Azure.cs
edmondshtogu: Done, same shape as AWS. Shared plumbing lives in `WithFlociReferenceCore`, so each cloud only names its own variables.

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Azure.cs
edmondshtogu: It wouldn't have, and it was worse than that: hard-coding `host.docker.internal:{hostPort}` also routed sibling-container traffic out to the host and back. Per Aspire's documented resolution rules, container->container resolves to `{resourceName}:{targetPort}` on the container network and only container->host-process uses `host.docker.internal`. The env vars now carry unresolved endpoint expressions and Aspire resolves them per dependent, so the `--add-host=host.docker.internal:host-gateway` runtime arg is gone entirely and this should work on Docker Desktop, plain Linux Docker, Podman and Rancher Desktop alike.

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Common.cs
edmondshtogu: I do not have a windows machine to test out, but based on some research I did, since Docker on Windows uses WSL2 that should work.

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Gcp.cs
edmondshtogu: Done, same shape as AWS. Shared plumbing lives in WithFlociReferenceCore, so each cloud only names its own variables.

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Gcp.cs
edmondshtogu: This is due to the SDKs' own contract, read by the SDK with no consumer code. In the AWS SDK for .NET, `ClientConfig.ServiceURL`'s getter itself resolves `AWS_ENDPOINT_URL_<SERVICE_ID>` then falls back to `AWS_ENDPOINT_URL` (see [here](https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Core/Amazon.Util/EnvironmentVariables.cs#L24) and [here](https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Core/Amazon.Runtime/ClientConfig.cs#L349)); same story for the credential/region chain, the GCP `*_EMULATOR_HOST` set (`Google.Cloud.*` reads these via EmulatorDetection), and `AZURE_STORAGE_CONNECTION_STRING`. A `Floci__AWS__EndpointUrl` key would force every consumer to read config and hand-wire `AmazonS3Config.ServiceURL`, which defeats the point of an emulator integration. The .NET-idiomatic path is there in parallel: the standard `ConnectionStrings__{name}` is still injected, so `GetConnectionString("floci-aws")` works.

tests/CommunityToolkit.Aspire.Hosting.Floci.Tests/AppHostTests.cs
edmondshtogu: Rewritten as a `[Theory]` asserting `HTTP 200` from each resource's real health path.

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Azure.cs
```diff
@@ -0,0 +1,162 @@
...
+                        ctx.EnvironmentVariables["AZURE_STORAGE_CONNECTION_STRING"] = ReferenceExpression.Create(
+                            $"DefaultEndpointsProtocol=http;AccountName={FlociAzureContainerResource.DefaultAccountName};AccountKey={FlociAzureContainerResource.DefaultAccountKey};BlobEndpoint={blobEndpoint};");
+                    }));
```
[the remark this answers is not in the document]
edmondshtogu: Fixed. `floci-az` serves Blob, Queue and Table from the same port, so all three endpoints now share one URL and `Queue/Table` clients no longer fall through to `core.windows.net`. `DefaultEndpointsProtocol` follows the endpoint scheme too.

src/CommunityToolkit.Aspire.Hosting.Floci/CommunityToolkit.Aspire.Hosting.Floci.csproj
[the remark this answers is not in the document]
edmondshtogu: Updated to cover all three clouds (AddFlociAws/AddFlociAzure/AddFlociGcp) plus WithFlociUI, with azure gcp azurite tls added to the tags.

README.md
```diff
@@ -33,6 +33,7 @@ This repository contains the source code for the Aspire Community Toolkit, a col
...
+| - **Learn More**: [`Hosting.Floci`](src/CommunityToolkit.Aspire.Hosting.Floci/README.md) <br /> - Stable 📦: [![CommunityToolkit.Aspire.Hosting.Floci](https://img.shields.io/nuget/v/CommunityToolkit.Aspire.Hosting.Floci)](https://nuget.org/packages/CommunityToolkit.Aspire.Hosting.Floci/) <br /> - Preview 📦: [![CommunityToolkit.Aspire.Hosting.Floci](https://img.shields.io/nuget/vpre/CommunityToolkit.Aspire.Hosting.Floci?label=nuget%20(preview))](https://nuget.org/packages/CommunityToolkit.Aspire.Hosting.Floci/absoluteLatest) | An Aspire component leveraging the Floci AWS emulator container. Includes a WithFlociUI() extension for running the Floci UI web console alongside the emulator. |
```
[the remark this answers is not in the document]
edmondshtogu: Updated the package description to properly populate readme.

tests/CommunityToolkit.Aspire.Hosting.Floci.Tests/AwsContainerResourceCreationTests.cs
```diff
@@ -0,0 +1,245 @@
+using Aspire.Hosting;
+
+namespace CommunityToolkit.Aspire.Hosting.Floci.Tests;
+
+public class ContainerResourceCreationTests
```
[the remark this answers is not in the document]
edmondshtogu: Renamed to `AwsContainerResourceCreationTests`, alongside the Azure/GCP ones.

tests/CommunityToolkit.Aspire.Hosting.Floci.Tests/TypeScriptAppHostTests.cs
```diff
@@ -0,0 +1,19 @@
...
+public class TypeScriptAppHostTests
+{
+    [Fact]
+    public async Task TypeScriptAppHostCompilesAndStarts()
+    {
+        await TypeScriptAppHostTest.Run(
+            appHostProject: "CommunityToolkit.Aspire.Hosting.Floci.AppHost.TypeScript",
+            packageName: "CommunityToolkit.Aspire.Hosting.Floci",
+            exampleName: "floci",
+            waitForResources: ["floci-aws", "floci-az", "floci-gcp", "floci-custom", "floci-gcp-custom", "floci-persistent", "floci-mount"],
```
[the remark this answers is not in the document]
edmondshtogu: Added `floci-aws-ui` to `waitForResources`.

edmondshtogu:
@aaronpowell can you take a look again please?

aaronpowell (approved):

PR #1481, opened 2026-07-25
feat(Kind): add AddManifest resource for kubectl apply after cluster ready

tamirdresher:
## Motivation

For dev-loop scenarios that run against a Kind cluster during F5 (operators, controllers, admission webhooks, or in-source contributions to an OSS Kubernetes project) the AppHost often needs to apply a set of `CustomResourceDefinition`, `Namespace`, `ServiceAccount`, `ClusterRole`, `ClusterRoleBinding`, or `ConfigMap` objects to the cluster **after the cluster reaches Running** but **before dependent resources start**.

Today `CommunityToolkit.Aspire.Hosting.Kind` handles the Helm side of this via `cluster.AddHelmChart(...)`, but if you just have static YAML you have to hand-roll a lifecycle hook in the AppHost. This PR adds the natural equivalent of `AddHelmChart` for the `kubectl apply` case.

This adds **runtime** manifest apply (at F5 / `aspire start`). It is distinct from `Aspire.Hosting.Kubernetes`'s `PublishAsKubernetesService(r => r.AdditionalResources.Add(...))`, which is **build-time** (baked into a generated Helm chart at `aspire publish`). Both have legitimate use cases; this PR fills the runtime gap for local Kind clusters.

## What this adds

A new resource + extension API surface, all public APIs exported for Aspire tooling:

```csharp
var cluster = builder.AddKindCluster("mycluster");

// Single file
cluster.AddManifest("crds", "./manifests/crds.yaml");

// Directory, recursive, into a namespace
cluster.AddManifest("platform", "./manifests/platform")
    .WithRecursive()
    .WithNamespace("platform");

// Kustomize overlay directory: detected at apply time and applied with kubectl apply -k
cluster.AddManifest("overlay", "./manifests/overlays/dev");

// Server-side apply with a stable field manager
cluster.AddManifest("operator", "./manifests/argocd/install.yaml")
    .WithServerSideApply(forceConflicts: true)
    .WithFieldManager("aspire-apphost");

// Inline content, applied through stdin without temp files
cluster.AddManifestFromContent("demo-ns", """
    apiVersion: v1
    kind: Namespace
    metadata:
      name: aspire-demo
    """);

// Downstream resources can wait for the apply to succeed
var crds = cluster.AddManifest("crds", "./manifests/crds.yaml");

builder.AddContainer("my-operator", "my-org/operator")
    .WithReference(cluster)
    .WaitFor(crds);
```

## Design

Mirrors the existing Kind deployed-resource pattern and K3s manifest semantics while keeping Kind's host-side kubectl execution model:

| Layer | Helm equivalent | New file |
|---|---|---|
| Resource | `KindHelmChartResource` | `K8sManifestResource : KindDeployedResource` |
| CLI orchestration | `HelmManager` | `KubectlManager` (wraps `IProcessRunner`, internal static argument builders for testability) |
| Extensions | `KindHelmChartResourceBuilderExtensions` | `KindManifestResourceBuilderExtensions` |

Lifecycle callback matches `AddHelmChart` step-for-step:

1. `WaitForResourceAsync(parent.Name, Running)`
2. `PublishAsync(BeforeResourceStartedEvent)`
3. `PublishUpdateAsync(Starting)`
4. try { `KubectlManager.ApplyAsync` -> `PublishUpdateAsync(Running)` }
5. catch { `PublishUpdateAsync(FailedToStart)` -> rethrow }

`KubectlManager.ApplyAsync` now probes `kubectl cluster-info` before apply so the Kind resource state and Kubernetes API reachability are both satisfied before manifests are applied.

## New public API

| Method | Maps to |
|---|---|
| `AddManifest(name, manifestPath)` | `kubectl apply -f <path> --kubeconfig=<cluster kubeconfig>` or `kubectl apply -k <path>` for Kustomize directories |
| `AddManifestFromContent(name, content)` | `kubectl apply -f - --kubeconfig=<cluster kubeconfig>` with content on stdin |
| `WithNamespace(ns)` (inherited from `KindDeployedResource`) | `--namespace <ns>` (does not create the namespace for manifests) |
| `WithRecursive()` | `--recursive` for normal file/directory applies; ignored with a warning for Kustomize/stdin |
| `WithServerSideApply(forceConflicts: false)` | `--server-side` (+ `--force-conflicts`) |
| `WithFieldManager(name)` | `--field-manager <name>` |
| `WithApplyTimeout(timeout)` | Bounded `kubectl apply` execution time (default 5 minutes) |
| `WithCrdWaitTimeout(timeout)` | CRD `Established` wait timeout (default 5 minutes) |
| `WithCrdWaitBehavior(behavior)` | `Fail` by default; `BestEffort` logs a warning and continues |

## Tests

xUnit v3 coverage in `KindManifestTests.cs` / `AddKindClusterTests.cs` now includes:

- Resource creation, parent wiring, name/path behavior, and inline content registration
- Null-guard tests for builder, name, manifestPath/content, and fieldManager
- AppHost-relative path resolution and absolute-path preservation
- Kustomize detection for lowercase and capitalized marker variants, including apply-time detection
- Property-set tests for `WithRecursive`, `WithServerSideApply(forceConflicts)`, `WithFieldManager`, `WithApplyTimeout`, `WithCrdWaitTimeout`, and `WithCrdWaitBehavior`
- Default-value tests for recursive/server-side/force-conflicts, field manager, namespace, apply timeout, and CRD wait behavior
- `KubectlManager.CreateApplyArguments`, `CreateWaitArguments`, and cluster-info retry behavior
- Apply timeout cancellation
- Default fail-fast and opt-in best-effort CRD wait behavior
- kubeconfig path redaction in debug logs

Full `CommunityToolkit.Aspire.Hosting.Kind.Tests` project passes locally: **206 / 206**.

## Real-world usage

Upstreamed from https://github.com/tamirdresher/aspire-argocd-dev-loop, which spins up an argo-cd F5 inner loop against a Kind cluster and uses `AddManifest` to install the ArgoCD manifests before the `argocd-server` and `repo-server` projects start. Prior to this API the sample carried a hand-rolled `OnBeforeStart` hook with its own `Process.Start("kubectl", ...)`.

## Open questions for reviewers

**Deferred:** naming consistency between this integration's `AddManifest` and `CommunityToolkit.Aspire.Hosting.K3s`'s `AddK8sManifest`. Both are `[AspireExport]`-ed public API in separate packages; unifying would be a breaking change for K3s consumers. Happy to align on either name if maintainers have a preference.

## Follow-up

I have a separate small PR queued: `WithPortMapping(hostPort, containerPort)` as a fluent shorthand over the existing `WithKindConfig(Action<KindConfigModel>)` for exposing NodePort services on `127.0.0.1`. Shipping it separately to keep this PR focused on the new resource type.

## Docs

README section updated in `src/CommunityToolkit.Aspire.Hosting.Kind/README.md`: usage examples for single file, recursive directory, Kustomize auto-detect, inline content, server-side apply, CRD wait behavior, namespace behavior, security notes, and the `API reference` table entries.

## Checklist

- [x] Design mirrors existing resource patterns (`KindHelmChartResource` and K3s `K8sManifestResource` semantics)
- [x] Public API export surface is covered where Aspire supports `[AspireExport]`
- [x] Tests pass locally (206 / 206)
- [x] README updated

## Follow-on docs

`learn.microsoft.com/dotnet/aspire/community-toolkit/hosting-kind` lives in `MicrosoftDocs/aspire-docs`, outside this fork's tree, and will need a matching update for `AddManifest`. That can be handled as a separate PR after this merges.


src/CommunityToolkit.Aspire.Hosting.Kind/CrdWaitBehavior.cs
```diff
@@ -0,0 +1,26 @@
...
+
+[assembly: Aspire.Hosting.AspireExport(typeof(Aspire.Hosting.CrdWaitBehavior))]
+
+namespace Aspire.Hosting;
+
+/// <summary>
+/// Specifies how Kind manifest resources handle CRD Established-condition wait failures.
+/// </summary>
+public enum CrdWaitBehavior
+{
```
MattKotsenas, line 14: It feels odd to introduce a new, parallel wait behavior?

src/CommunityToolkit.Aspire.Hosting.Kind/HelmManager.cs
```diff
@@ -78,4 +138,48 @@ internal static IReadOnlyList<string> CreateInstallArguments(KindHelmChartResour
 
         return arguments;
     }
+
+    internal static TimeSpan ComputeRetryBackoff(TimeSpan initialBackoff, int failureCount)
```
MattKotsenas, line 195: Many (most?) places just take a dependency on M.E.Resilence. Might be worth considering? More important if you want it to be user configurable, which I would push for eventually, because waiting 5 minutes for a timeout to occur can be painful in inner loop scenarios.

src/CommunityToolkit.Aspire.Hosting.Kind/DefaultProcessRunner.cs
```diff
@@ -20,15 +20,17 @@ public async Task<ProcessResult> RunAsync(
         IReadOnlyList<string> arguments,
         string? workingDirectory = null,
         IReadOnlyDictionary<string, string>? environmentVariables = null,
+        string? standardInput = null,
         CancellationToken cancellationToken = default)
     {
-        logger.LogDebug("Executing: {FileName} {Arguments}", fileName, string.Join(' ', arguments));
+        logger.LogDebug("Executing: {FileName} {Arguments}", fileName, RedactSensitiveArgs(arguments));
```
MattKotsenas: I'm not sure I'm in favor of redaction at the debug level?

src/CommunityToolkit.Aspire.Hosting.Kind/HelmManager.cs
```diff
@@ -78,4 +138,48 @@ internal static IReadOnlyList<string> CreateInstallArguments(KindHelmChartResour
...
+            return TimeSpan.MaxValue;
+        }
+
+        var multiplier = 1L << (failureCount - 1);
+        return initialBackoff.Ticks > TimeSpan.MaxValue.Ticks / multiplier
+            ? TimeSpan.MaxValue
+            : TimeSpan.FromTicks(initialBackoff.Ticks * multiplier);
+    }
+
+    private static bool ShouldRetryForCrdRace(ProcessResult result)
```
MattKotsenas: Is this a scenario we would even want to support? It feels like if we're going to have dependencies between, either we group them into a chart (and make that easier if needed) or we use `.WaitFor()` to do strict ordering.

Otherwise, I think we're letting the user's sloppiness leak in a way that might bite them later.

src/CommunityToolkit.Aspire.Hosting.Kind/KindClusterLifecycleHook.cs
```diff
@@ -5,31 +5,133 @@
...
+
+    private void RegisterTerminationHandlers()
+    {
+        lock (_registrationLock)
+        {
+            if (_terminationHandlersRegistered)
+            {
+                return;
+            }
+
```
MattKotsenas: Is it correct to hook up _all_ these handlers? That feels like you're going beyond what the `IHostApplicationLifetime` interface calls for? Should'nt we just hook stopping or something like that?

src/CommunityToolkit.Aspire.Hosting.Kind/KindClusterResourceBuilderExtensions.cs
```diff
@@ -137,6 +137,76 @@ public static IResourceBuilder<T> WithKubernetesVersion<T>(
...
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentException.ThrowIfNullOrWhiteSpace(hostPath);
+        ArgumentException.ThrowIfNullOrWhiteSpace(containerPath);
+        if (!containerPath.StartsWith("/", StringComparison.Ordinal))
+        {
+            throw new ArgumentException("Container path must be an absolute Linux path (for example '/var/local/data').", nameof(containerPath));
+        }
+
+        var normalizedHostPath = Path.IsPathRooted(hostPath)
+            ? Path.GetFullPath(hostPath)
```
MattKotsenas, line 199: I think you can never _not_ be rooted, correct? You're checking that just above?

src/CommunityToolkit.Aspire.Hosting.Kind/KindConfigGenerator.cs
```diff
@@ -49,7 +49,15 @@ internal static async Task<string> GenerateConfigAsync(IKindResource resource, C
...
-            imageAnnotation.Version is not null)
+            imageAnnotation.Image is not null)
+        {
+            foreach (var node in config.Nodes)
+            {
+                node.Image ??= imageAnnotation.Image;
+            }
+        }
+        else if (resource.TryGetLastAnnotation<KindNodeImageAnnotation>(out imageAnnotation) &&
+                 imageAnnotation.Version is not null)
```
MattKotsenas, line 60: It looks like you're setting up a situation where either Image or Version can be set, but if both are set one is silently ignored. I think at a minimum we should fail loudly in that situation, or change the modelling so it's not constructable.

src/CommunityToolkit.Aspire.Hosting.Kind/KindHelmChartResourceBuilderExtensions.cs
```diff
@@ -149,10 +149,69 @@ public static IResourceBuilder<KindHelmChartResource> WithHelmValue(
...
+    public static IResourceBuilder<KindHelmChartResource> WithCrdWaitRetry(
+        this IResourceBuilder<KindHelmChartResource> builder,
+        int maxAttempts = 3,
+        TimeSpan? backoff = null)
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentOutOfRangeException.ThrowIfLessThan(maxAttempts, 2);
+
+        builder.Resource.CrdWaitRetryMaxAttempts = maxAttempts;
+        builder.Resource.CrdWaitRetryBackoff = KubectlTimeouts.Normalize(backoff ?? TimeSpan.FromSeconds(5), nameof(backoff));
```
MattKotsenas: This 5 second default is specified in a few different places. Can we centralize to prevent potential drift?

src/CommunityToolkit.Aspire.Hosting.Kind/KindManifestResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,289 @@
...
+        [ResourceName] string name,
+        string manifestPath)
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentNullException.ThrowIfNull(name);
+        ArgumentException.ThrowIfNullOrWhiteSpace(manifestPath);
+
+        var absolutePath = System.IO.Path.IsPathRooted(manifestPath)
+            ? manifestPath
+            : System.IO.Path.GetFullPath(
```
MattKotsenas: Just commenting here, but applies more broadly:

Allowing for non-rooted paths and guessing at what the base path is sets users up for nasty surprises in the future. Relative to the AppHost would break in any situation where the app is built and then copied to a test machine (currently difficult for other reasons, but we shouldn't make it worse).

I think we could _either_ make it relative to the actual working directory, because that's under user control, or disallow relative paths.

src/CommunityToolkit.Aspire.Hosting.Kind/KindManifestResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,289 @@
...
+            clusterInfoMaxWait: resource.ClusterReadyTimeout);
+    }
+
+    /// <summary>
+    /// Recursively applies manifests from subdirectories (maps to <c>kubectl apply --recursive</c>).
+    /// Only meaningful when the manifest path is a directory.
+    /// </summary>
+    /// <param name="builder">The manifest resource builder.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{K8sManifestResource}"/>.</returns>
+    [AspireExport]
```
MattKotsenas, line 159: It feels kinda odd to model this on the resource itself, rather than via an annotation? It makes sense in that it'd be an annotation for what is currently a single type, but I assume there's even more params we aren't currently modelling, and thus the resource is likely to take breaking changes frequently as folks add an ever-growing list of properties?

MattKotsenas: I think doing it that way would also let you centralize the handling for validating that the path is a directory?

src/CommunityToolkit.Aspire.Hosting.Kind/KubectlManager.cs
```diff
@@ -0,0 +1,516 @@
...
+    private readonly Func<TimeSpan, CancellationToken, Task> _delayAsync = delayAsync ?? Task.Delay;
+    private readonly TimeSpan _clusterInfoMaxWait = clusterInfoMaxWait ?? ClusterInfoMaxWait;
+    private readonly TimeSpan _clusterInfoProbeTimeout = clusterInfoProbeTimeout ?? ClusterInfoProbeTimeout;
+
+    internal TimeSpan ClusterInfoMaxWaitForTesting => _clusterInfoMaxWait;
+
+    /// <summary>
+    /// Waits for the cluster API to answer, then applies the manifest via <c>kubectl apply</c>.
+    /// </summary>
+    public async Task ApplyAsync(K8sManifestResource resource, ILogger logger, CancellationToken cancellationToken)
```
MattKotsenas: It's not "wrong", but it feels a little odd to me that we're doing validation here around the resource's flags, and also that there's nothing ensuring (or even suggesting) that the `ILogger` is the resource logger for the resource, which I think you'd want in this case?

src/CommunityToolkit.Aspire.Hosting.Kind/KubectlManager.cs
```diff
@@ -0,0 +1,516 @@
...
+    {
+        ArgumentNullException.ThrowIfNull(resource);
+
+        var args = CreateClusterInfoArguments(resource.Parent.KubeconfigPath);
+        var nextDelay = ClusterInfoInitialDelay;
+        var stopwatch = Stopwatch.StartNew();
+
+        while (true)
+        {
+            if (stopwatch.Elapsed >= _clusterInfoMaxWait)
```
MattKotsenas: Another place where M.E.Reslience would probably be handy

tests/CommunityToolkit.Aspire.Hosting.Kind.Tests/AddKindClusterTests.cs
```diff
@@ -50,6 +53,85 @@ public async Task WithKubernetesVersionSetsVersion()
...
+            .WithNodeImage("myacr.azurecr.io/kindest/node:v1.32.2");
+
+        using var app = builder.Build();
+        var appModel = app.Services.GetRequiredService<DistributedApplicationModel>();
+
+        var resource = Assert.Single(appModel.Resources.OfType<KindClusterResource>());
+        var configPath = await KindConfigGenerator.GenerateConfigAsync(resource, CancellationToken.None);
+        try
+        {
+            var yaml = await File.ReadAllTextAsync(configPath);
```
MattKotsenas, line 92: This only asserts you have 3 parts, not that they're the value you expect?

tests/CommunityToolkit.Aspire.Hosting.Kind.Tests/AddKindClusterTests.cs
```diff
@@ -50,6 +53,85 @@ public async Task WithKubernetesVersionSetsVersion()
         }
     }
 
+    [Fact]
+    public async Task WithNodeImageSetsImageOnAllNodes()
+    {
+        var builder = DistributedApplication.CreateBuilder();
```
MattKotsenas: Only left once, but applies to all:

I think you want `DistributedApplicationTestingBuilder` here, correct?

tests/CommunityToolkit.Aspire.Hosting.Kind.Tests/AddKindClusterTests.cs
```diff
@@ -570,4 +817,85 @@ public void KindClusterResource_Constructor_SetsPathsCorrectly()
...
+            }
+
+            await Task.Delay(20);
+        }
+
+        throw new Xunit.Sdk.XunitException("Timed out waiting for asynchronous condition.");
+    }
+
+    private sealed class CapturingLogger : ILogger
+    {
```
MattKotsenas, line 929: There's an M.E.Logging.Testing for this use case.

tests/CommunityToolkit.Aspire.Hosting.Kind.Tests/FakeProcessRunner.cs
```diff
@@ -19,31 +19,44 @@ internal sealed class FakeProcessRunner : IProcessRunner
...
+            result = ResultsByFileName.TryGetValue(fileName, out var fileResult)
+                    ? fileResult
                     : Results.Count > 0
                         ? Results.Dequeue()
-                        : NextResult);
+                        : NextResult;
+            delay = Delays.Count > 0 ? Delays.Dequeue() : Delay;
         }
+
+        if (delay > TimeSpan.Zero)
```
MattKotsenas, line 49: Introducing real delays into the tests seems bad; can we not use `TimeProvider` and fake them instead?

MattKotsenas (changes requested):
Left comments even when I wasn't quite sure, so feel free to resolve threads that are incorrect. If you have any questions, please let me know. Thanks!

src/CommunityToolkit.Aspire.Hosting.Kind/KindClusterResourceBuilderExtensions.cs
tamirdresher: Good catch to call out the path handling here. The rooted-path check in this block is for `containerPath`; `hostPath` is still intentionally allowed to be relative and is resolved against `AppHostDirectory` today.

Separately, as part of the review fixes in this PR, `AddManifest` now rejects relative `manifestPath` values and requires an absolute path, because resolving them from build-time `AppHostDirectory` metadata was not portable once the AppHost is copied or published elsewhere.

That does leave an inconsistency between `WithNodeMount` host paths (currently permissive) and `AddManifest` manifest paths (now strict). I did not change `WithNodeMount` semantics in this PR, but I think it is worth a follow-up discussion whether host paths should also require absolute paths for consistency.

tests/CommunityToolkit.Aspire.Hosting.Kind.Tests/AddKindClusterTests.cs
tamirdresher: This assertion is counting occurrences of the exact expected YAML fragment (`image: myacr.azurecr.io/kindest/node:v1.32.2`), not just checking that the split produced three arbitrary parts.

That said, I am happy to strengthen it further if you would prefer a more structural assertion, for example by parsing the generated YAML and asserting the node image values directly.

src/CommunityToolkit.Aspire.Hosting.Kind/api/CommunityToolkit.Aspire.Hosting.Kind.cs
aaronpowell, line 1: This will get generated when the PR is merged so it can be deleted

tamirdresher: Good catch — I verified the checked-in GenAPI file matched the current public surface, but it doesn't need to be in this PR. I removed src/CommunityToolkit.Aspire.Hosting.Kind/api/CommunityToolkit.Aspire.Hosting.Kind.cs and pushed the update.

tests/CommunityToolkit.Aspire.Hosting.Kind.Tests/AddKindClusterTests.cs
tamirdresher: Yeah, I know about Microsoft.Extensions.Logging.Testing, figured the custom capturing logger here is simple enough (it's like 10 lines) that pulling in another test dependency wasn't worth it. Happy to switch if you feel strongly about standardizing on it though.

aaronpowell (approved):

PR #1546, opened 2026-08-26
feat(floci): model Cosmos API as a child resource

### LINKED ISSUE #1545: [Feature]: CommunityToolkit.Aspire.Hosting.Floci — Cosmos DB connection string helper for AddFlociAzure
### Package

CommunityToolkit.Aspire.Hosting.Floci

### Is your feature request related to a problem? Please describe.

`AddFlociAzure(...)` today models a single HTTP endpoint (`:4577`) and `WithReference(azure)` injects `ConnectionStrings__{name}` (the base URL) plus a ready-to-use `AZURE_STORAGE_CONNECTION_STRING` for the Blob/Queue/Table SDKs.

There's no equivalent for **Cosmos DB**, even though floci-az emulates the Cosmos SQL/NoSQL API on the same `:4577` endpoint. A .NET app that wants to talk to floci's Cosmos has to hand-build the connection string itself — combining the emulator endpoint (which is dynamic under `--isolated`/random ports), the `-cosmos` account path suffix, and the well-known Cosmos emulator key — instead of just calling `AddAzureCosmosClient("cosmos")`.

This is the one missing piece for a project to consume floci-az Cosmos through the standard Aspire connection-string flow.

### Describe the solution you'd like

A `WithCosmosReference` helper on the Floci Azure resource that injects a Cosmos connection string under a caller-chosen connection name, mirroring how the storage connection string is injected today:

```csharp
var azure = builder.AddFlociAzure("floci-az");

    .WithCosmosReference(azure)          // connectionName defaults to "cosmos"
    .WaitFor(azure);
```

App side is then the standard Aspire flow:

```csharp
builder.AddAzureCosmosClient("cosmos");
```

Injected env var (endpoint resolved by Aspire, so it tracks random/isolated ports):

```
ConnectionStrings__cosmos = AccountEndpoint={scheme}://{host}:{port}/devstoreaccount1-cosmos/;AccountKey=C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==;
```

Proposed signature (parallels the existing `withFlociAzureReference`, with its own TS binding since the generated bindings have no overload resolution):

```csharp
[AspireExport("withFlociAzureCosmosReference")]
public static IResourceBuilder<TDestination> WithCosmosReference<TDestination>(
    this IResourceBuilder<TDestination> builder,
    IResourceBuilder<FlociAzureContainerResource> floci,
    string connectionName = "cosmos",
    string? accountName = null)   // defaults to devstoreaccount1
    where TDestination : IResourceWithEnvironment;
```

Uses the well-known Cosmos emulator key (`C2y6…`), distinct from the Azurite storage key already used for `AZURE_STORAGE_CONNECTION_STRING`. The endpoint is built from the resource's existing `Scheme`/`Host`/`Port` reference expressions, so it resolves correctly for both a project (`localhost:{hostPort}`) and a sibling container.

### Scope / non-goals

- Just the connection-string plumbing. Client-side quirks of talking to the floci Cosmos emulator over HTTP from the .NET SDK (Gateway mode, HTTP/1.1 pinning, dev-cert bypass) are the app's concern, as with the existing storage helper.
- **Service Bus is intentionally out of scope here** — its AMQP data plane runs on a floci-launched Artemis *sidecar* whose port isn't modelled as an Aspire endpoint (only the `:4577` management plane is), so a SB connection-string helper needs an endpoint-modelling decision on the floci-az side first. Cosmos has no such issue: it's served on the endpoint already exposed.

### Additional context

I'm a floci-az contributor and have a working implementation ready; happy to open the PR against this issue.

### Are you willing to contribute a PR?

- [x] Yes

thomhurst:
Closes #1545

## Overview

Models the Cosmos DB API exposed by the Floci Azure emulator as a dedicated Aspire child resource. Dependents now use Aspire's standard `WithReference` flow, so Cosmos appears in the resource graph and can be swapped with another `IResourceWithConnectionString` for deployment.

## Usage

```csharp
var azure = builder.AddFlociAzure("floci-az");
var cosmos = azure.WithCosmos();

    .WithReference(azure)    // storage variables (optional)
    .WithReference(cosmos)   // ConnectionStrings__cosmos
    .WaitFor(azure);
```

```typescript
const azure = await builder.addFlociAzure('floci-az');
const cosmos = await azure.withCosmos();

await builder.addProject('api', '../MyApi/MyApi.csproj')
    .withFlociAzureReference(azure)
    .withReference(cosmos)
    .waitFor(azure);
```

App side remains the normal flow:

```csharp
builder.AddAzureCosmosClient("cosmos");
```

## What's included

- `WithCosmos(name = "cosmos", accountName = null)` returning `IResourceBuilder<FlociAzureCosmosResource>`.
- `FlociAzureCosmosResource` implements `IResourceWithConnectionString` and `IResourceWithParent<FlociAzureContainerResource>`.
- Standard `WithReference(cosmos)` injects `ConnectionStrings__{resourceName}`.
- Cosmos endpoint follows Aspire endpoint resolution and TLS scheme changes.
- Custom resource and account names remain supported.
- README and unit tests updated.
- Generated public API file left for CI regeneration.

## Testing

- `WithReferenceTests`: 11/11 passed.
- Floci project build: `net8.0`, `net9.0`, and `net10.0`; 0 warnings, 0 errors.

### DIFF
--- a/src/CommunityToolkit.Aspire.Hosting.Floci/api/CommunityToolkit.Aspire.Hosting.Floci.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.Floci/api/CommunityToolkit.Aspire.Hosting.Floci.cs
@@ -22,10 +22,6 @@ namespace Aspire.Hosting
         [AspireExport]
         public static ApplicationModel.IResourceBuilder<ApplicationModel.FlociAwsContainerResource> WithConfigFile(this ApplicationModel.IResourceBuilder<ApplicationModel.FlociAwsContainerResource> builder, string hostPath) { throw null; }
 
-        [AspireExport("withFlociAzureCosmosReference")]
-        public static ApplicationModel.IResourceBuilder<TDestination> WithCosmosReference<TDestination>(this ApplicationModel.IResourceBuilder<TDestination> builder, ApplicationModel.IResourceBuilder<ApplicationModel.FlociAzureContainerResource> floci, string connectionName = "cosmos", string? accountName = null)
-            where TDestination : ApplicationModel.IResourceWithEnvironment { throw null; }
-
         [AspireExport]
         public static ApplicationModel.IResourceBuilder<ApplicationModel.FlociAwsContainerResource> WithDataBindMount(this ApplicationModel.IResourceBuilder<ApplicationModel.FlociAwsContainerResource> builder, string source, bool isReadOnly = false) { throw null; }
 
--- a/src/CommunityToolkit.Aspire.Hosting.Floci/api/CommunityToolkit.Aspire.Hosting.Floci.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.Floci/api/CommunityToolkit.Aspire.Hosting.Floci.cs
@@ -22,6 +22,10 @@ namespace Aspire.Hosting
         [AspireExport]
         public static ApplicationModel.IResourceBuilder<ApplicationModel.FlociAwsContainerResource> WithConfigFile(this ApplicationModel.IResourceBuilder<ApplicationModel.FlociAwsContainerResource> builder, string hostPath) { throw null; }
 
+        [AspireExport("withFlociAzureCosmosReference")]
+        public static ApplicationModel.IResourceBuilder<TDestination> WithCosmosReference<TDestination>(this ApplicationModel.IResourceBuilder<TDestination> builder, ApplicationModel.IResourceBuilder<ApplicationModel.FlociAzureContainerResource> floci, string connectionName = "cosmos", string? accountName = null)
+            where TDestination : ApplicationModel.IResourceWithEnvironment { throw null; }
+
         [AspireExport]
         public static ApplicationModel.IResourceBuilder<ApplicationModel.FlociAwsContainerResource> WithDataBindMount(this ApplicationModel.IResourceBuilder<ApplicationModel.FlociAwsContainerResource> builder, string source, bool isReadOnly = false) { throw null; }
 
--- a/src/CommunityToolkit.Aspire.Hosting.Floci/FlociAzureContainerResource.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.Floci/FlociAzureContainerResource.cs
@@ -14,6 +14,10 @@ public class FlociAzureContainerResource(string name) : FlociContainerResource(n
     internal const string DefaultAccountName = "devstoreaccount1";
     internal const string DefaultAccountKey = "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMh0==";
 
+    // Well-known Cosmos DB emulator account key that floci-az accepts by default (no auth enforced).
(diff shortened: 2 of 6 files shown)


src/CommunityToolkit.Aspire.Hosting.Floci/api/CommunityToolkit.Aspire.Hosting.Floci.cs
```diff
@@ -22,6 +22,10 @@ public static partial class FlociHostingExtension
         [AspireExport]
         public static ApplicationModel.IResourceBuilder<ApplicationModel.FlociAwsContainerResource> WithConfigFile(this ApplicationModel.IResourceBuilder<ApplicationModel.FlociAwsContainerResource> builder, string hostPath) { throw null; }
 
+        [AspireExport("withFlociAzureCosmosReference")]
+        public static ApplicationModel.IResourceBuilder<TDestination> WithCosmosReference<TDestination>(this ApplicationModel.IResourceBuilder<TDestination> builder, ApplicationModel.IResourceBuilder<ApplicationModel.FlociAzureContainerResource> floci, string connectionName = "cosmos", string? accountName = null)
+            where TDestination : ApplicationModel.IResourceWithEnvironment { throw null; }
+
```
aaronpowell: This will be auto-generated by a CI job, so please revert that

aaronpowell (changes requested):
Doing the review and I feel like this kind of has to work around some of the design of how Aspire works since we're not really treating the floci resource as something that can provide a Cosmos-formatted connection string, and bypassing the `WithReference` design for that.

I wonder if this as an API wouldn't be better:

```csharp
var azure = builder.AddFlociAzure("floci-az");

var cosmos = azure.WithCosmos();

  .WithReference(azure)
  .WithReference(cosmos)
  .WaitFor(azure);
```

In this case the `WithCosmos()` method would return a `IResourceBuilder<FlociAzureCosmosResource>`, and then we treat it as a dedicated resource on its own which overrides the connection string creation to return a Cosmos formatted one.

This then models the CosmosDB part of Floci as a resource in its own right, making it visualise what the resource graph looks like, unlike your current proposal which would show the resource graph as not containing a CosmosDB resource since the resource _doesn't exist_ it's just something teased out of another resource. Then, since it's modelled as a resource, you can handle deployments easier by doing something like:

```csharp
IResourceBuilder<IResourceWithConnectionString> cosmos = builder.IsDeployment ?
  builder.AddAzureCosmos(...) :
  flociAzure.WithCosmos();
```

(yes, that code isn't 100% right, I'm doing it from memory 😅).

Does that make sense?

thomhurst:
Addressed review feedback in 088933e:

- Replaced `WithCosmosReference` with `azure.WithCosmos()`.
- Added `FlociAzureCosmosResource` implementing `IResourceWithConnectionString` and `IResourceWithParent<FlociAzureContainerResource>`.
- Dependents now use standard `.WithReference(cosmos)`.
- Added parent relationship so Cosmos appears in resource graph.
- Reverted generated API surface file.
- Updated C#/TypeScript docs and tests.

Validation: `WithReferenceTests` 11/11 passed; Floci project builds cleanly on net8.0/net9.0/net10.0.

aaronpowell (approved):

aaronpowell:
Thanks @thomhurst - looks right to go.

PR #1551, opened 2026-08-31
[K3s] Emit LF line endings in generated helm/kubectl scripts (fixes Windows-only busybox syntax error)

### LINKED ISSUE #1550: [K3s] AddHelmRelease and AddK8sManifest always fail on Windows — generated shell scripts have CRLF line endings, busybox ash aborts with `unexpected end of file (expecting "then")`
### Describe the bug

On Windows, `CommunityToolkit.Aspire.Hosting.K3s` `AddHelmRelease(...)` always fails. The generated
`/helm-install.sh` is injected into the `alpine/helm` container and executed by its `/bin/sh`
(busybox ash), which aborts immediately:

```
/helm-install.sh: line 15: syntax error: unexpected end of file (expecting "then")
```

The helm container exits with code 2 in under a second and the release is never installed — the
target namespace is never created. The k3s cluster itself comes up healthy, which makes the failure
look like a cluster problem rather than a script problem.

The cause is line endings. `BuildHelmScript` composes the script with `StringBuilder.AppendLine`,
which appends `Environment.NewLine` — `\r\n` on Windows, `\n` on Linux. busybox ash does not strip
the trailing CR, so `if [ "$_k3s_wait" -ge 600 ]; then` is tokenized as `then\r`, which is not the
`then` keyword. The `if` is therefore never closed and the parser reports the error at EOF.

**`AddK8sManifest` has the identical defect.** `BuildManifestScript` is built the same way (22
`AppendLine` calls, two `if … then` blocks) and runs in `alpine/kubectl`, which is also busybox. It
only escapes notice because `AddHelmRelease` fails first in the sample AppHost.

This is Windows-only. On Linux/macOS `Environment.NewLine` is already `\n`, which is why CI is green.

### Regression

**Not a regression — it has never worked on Windows.**  `BuildHelmScript` contained all 12 `AppendLine` calls in the commit that introduced the integration (`3a908021`, PR #1322, 2026-06-29). Every published version of `CommunityToolkit.Aspire.Hosting.K3s` is affected on Windows hosts.

### Steps to reproduce

```text
1. Environment: Windows 11 (reproduced on ARM64), .NET 10, Docker Desktop with Linux containers
   (Docker 29.7.2). Aspire CLI/DCP 13.5.2, AppHost SDK 13.5.3.
2. Create an AppHost, add `CommunityToolkit.Aspire.Hosting.K3s` 13.5.0, and paste:


var cluster = builder.AddK3sCluster("k8s").WithDataVolume();

cluster.AddHelmRelease(
    name: "podinfo",
    chart: "podinfo",
    repo: "https://stefanprodan.github.io/podinfo",
    version: "6.7.1",
    @namespace: "podinfo");

builder.Build().Run();


3. Run `aspire start`.
4. The `k8s` cluster reaches a healthy state (nodes Ready, API server up, coredns and flannel
   running).
5. The `podinfo-*` container (`alpine/helm:3.18.0`) exits within one second with exit code 2. Its
   log contains only:

   
   /helm-install.sh: line 15: syntax error: unexpected end of file (expecting "then")
   

6. `kubectl get ns` shows the `podinfo` namespace was never created.

Deterministic — reproduces on every clean start (verified 3×, including after wiping `.k3s`, all DCP
temp directories, and orphaned `dcp.exe` processes).

To confirm the cause directly, without an AppHost:


docker cp podinfo-<id>:/helm-install.sh .
file helm-install.sh      # reports "with CRLF line terminators"
docker run --rm -i --entrypoint /bin/sh alpine/helm:3.18.0 -n < helm-install.sh
# /bin/sh: syntax error: unexpected end of file (expecting "then")


The same check against `alpine/kubectl:1.36.0` with the `AddK8sManifest` script produces the same
error, demonstrating the second affected code path.
```

### Expected behavior

`AddHelmRelease` should install the chart regardless of the host OS: the helm container runs
`/helm-install.sh` to completion, exits 0, and the release is deployed into the requested namespace.
`AddK8sManifest` should likewise apply its manifests successfully from a Windows host.

The generated scripts are POSIX shell executed inside Linux containers, so they must always be
emitted with LF line endings — the host's `Environment.NewLine` should have no bearing on their
content.

### Screenshots

_No response_

### IDE and version

VS Code

### IDE version

_No response_

### Nuget packages

```text
- `CommunityToolkit.Aspire.Hosting.K3s` (13.5.0)
- `Aspire.Hosting.AppHost` (13.5.3)
- Aspire CLI / DCP 13.5.2, .NET 10

Container images involved:

- `alpine/helm:3.18.0` (its `/bin/sh` is `/bin/busybox`)
- `alpine/kubectl:1.36.0` (also busybox)
- `rancher/k3s:v1.36.0-k3s1`
```

### Root cause, byte-for-byte

`BuildHelmScript` in `src/CommunityToolkit.Aspire.Hosting.K3s/K3sBuilderExtensions.Helm.cs`:

```csharp
internal static string BuildHelmScript(HelmReleaseResource release)
{
    var sb = new StringBuilder("#!/bin/sh\nset -e\n");   // hardcoded LF
    ...
    sb.AppendLine("_k3s_wait=0");                        // Environment.NewLine
    sb.AppendLine($"until [ -f ... ] && helm list ...; do");
    sb.AppendLine("  if [ \"$_k3s_wait\" -ge 600 ]; then");
    ...
}
```

The script is injected via `WithContainerFiles` and copied into the container verbatim. Extracted
from the actual failing container:

```
#!/bin/sh\nset -e\n            <- LF  (from the seed string)
_k3s_wait=0\r\n                <- CRLF (AppendLine on Windows)
until [ -f /tmp/k3s-kubeconfig.yaml ] && ...; do\r\n
  _k3s_wait=$((_k3s_wait + 5))\r\n
  if [ "$_k3s_wait" -ge 600 ]; then\r\n    <- token is "then\r"
  ...
  sleep 5\r\n
done\r\n
helm repo add ... \r\n
helm repo update ... \r\n
helm upgrade --install ... --version '6.7.1'
```

The mixed endings are why the reported line number is misleading: the seed string is hardcoded LF,
everything after it is CRLF, and the error surfaces at EOF rather than at the offending `then\r`.

### Why the k3s init entrypoint is *not* affected

`K3sInitEntrypointScript`, injected into the k3s server and agent containers, is a C# **raw string
literal** (`"""…"""`). Roslyn normalizes line endings inside raw string literals to `\n` regardless
of the source file's own endings, so it is emitted as LF even on Windows. That asymmetry is exactly
why the cluster comes up healthy while the helm container dies — worth knowing when triaging, since
it steers you away from the script as the suspect.

The only other `AppendLine` in `src/` is `NgrokExtensions.cs`, which builds a YAML config. YAML
parsers accept CRLF, so that one is harmless.

### Ruled out

- k3s cluster is healthy: nodes Ready, API server up, coredns + flannel running.
- The script content is complete and correct, not truncated — all lines present.
- busybox `/bin/sh` in `alpine/helm:3.18.0` parses the same `if`/`until` constructs correctly when
  given LF endings.
- Not stale DCP state: reproduces after a full clean teardown.

### Suggested fix

Emit LF explicitly instead of `Environment.NewLine` in both `BuildHelmScript` and
`BuildManifestScript` — e.g. a small internal helper used in place of `AppendLine`:

```csharp
internal static StringBuilder AppendShellLine(this StringBuilder builder, string line) =>
    builder.Append(line).Append('\n');
```

Prefer this over normalizing the finished string with `sb.ToString().Replace("\r\n", "\n")`: the
replace would also rewrite CRLF appearing inside caller-supplied data — a `WithHelmValue` value, a
repo URL, a release name — on its way into `helm --set`. A helper only controls the terminators the
generator itself emits and leaves caller data byte-exact. (Caller values are already `ShellEscape`d
into POSIX single quotes, so a literal CR inside one is data, not a parse hazard.)

Worth adding a regression test that asserts the generated scripts contain no `\r`. Such a test fails
on Windows without the fix and is trivially green on Linux, which is the correct behaviour — the
invariant is identical on every host.

### Workaround until fixed

Run the AppHost on a Linux host or under WSL2, where `Environment.NewLine` is `\n`. Note that
switching from `AddHelmRelease` to `AddK8sManifest` is **not** a valid workaround on Windows — the
manifest path carries the same defect.

### Help us help you

henrikroschmann:
**Closes #1550**

`BuildHelmScript` and `BuildManifestScript` compose their scripts with
`StringBuilder.AppendLine`, which appends `Environment.NewLine` — `\r\n` on Windows. The
scripts are injected into Linux containers and executed by busybox `ash`, which does not
strip the trailing CR, so `if [ ... ]; then` is tokenized as `then\r`. The `if` is never
closed and the script dies at EOF with
`syntax error: unexpected end of file (expecting "then")`.

This makes `AddHelmRelease` fail 100% of the time on Windows hosts, and `AddK8sManifest`
carries the identical defect — it just escapes notice because the helm release fails first
in the sample AppHost. Both are fixed here.

### Changes

- **New** `K3sShellScript.AppendShellLine` — appends a hard `\n`, with XML docs explaining
  the busybox failure it exists to prevent.
- `K3sBuilderExtensions.Helm.cs` — 12 call sites switched off `AppendLine`.
- `K3sBuilderExtensions.Manifest.cs` — 22 call sites switched off `AppendLine`.
- Regression tests asserting the generated scripts contain no `\r`:
  `BuildHelmScriptUsesLfLineEndingsOnly`, `BuildManifestScriptUsesLfLineEndingsOnly`.

I used a helper rather than normalizing the finished string with
`sb.ToString().Replace("\r\n", "\n")`, because the replace would also rewrite CRLF inside
caller-supplied data — a `WithHelmValue` value, a repo URL, a release name — on its way
into `helm --set`. The helper only controls terminators the generator itself emits and
leaves caller data byte-exact.

Not affected, for reviewers wondering why the cluster comes up healthy while the helm
container dies: `K3sInitEntrypointScript` is a C# raw string literal, and Roslyn
normalizes line endings inside those to `\n` regardless of the source file's endings.

## PR Checklist

## Other information

The new regression tests fail on Windows without the fix and are trivially green on Linux —
which is correct, the invariant is the same on every host. Since CI runs on Linux, they
guard the intent rather than reproduce the failure there.

Verified on Windows 11 ARM64 / Docker 29.7.2 by running the **actual generated scripts** in
the real images against a live `rancher/k3s:v1.36.0-k3s1` cluster:

| Script | Image | Result |
| --- | --- | --- |
| pre-fix bytes, reproduced exactly | `alpine/helm:3.18.0` | `syntax error: unexpected end of file (expecting "then")`, exit 2 |
| fixed helm script | `alpine/helm:3.18.0` | `STATUS: deployed`, exit 0 — `podinfo` pod Running |
| fixed manifest script, plain YAML | `alpine/kubectl:1.36.0` | `configmap/app-config serverside-applied`, exit 0 |
| fixed manifest script, kustomize | `alpine/kubectl:1.36.0` | `Detected kustomization`, `configmap/monitoring-config serverside-applied`, exit 0 |

### DIFF
--- a/src/CommunityToolkit.Aspire.Hosting.K3s/K3sBuilderExtensions.Helm.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.K3s/K3sBuilderExtensions.Helm.cs
@@ -224,22 +224,22 @@ public static class K3sHelmBuilderExtensions
         // can appear in the bind-mount before the k8s hostname resolves in the helm
         // container. Using `helm list` (which calls the k8s API) verifies both the
         // file and the network path before proceeding.
-        sb.AppendLine("_k3s_wait=0");
-        sb.AppendLine($"until [ -f {K3sFileHelpers.ContainerKubeconfigPath} ] && helm list --kubeconfig {K3sFileHelpers.ContainerKubeconfigPath} > /dev/null 2>&1; do");
-        sb.AppendLine("  _k3s_wait=$((_k3s_wait + 5))");
-        sb.AppendLine("  if [ \"$_k3s_wait\" -ge 600 ]; then");
-        sb.AppendLine("    echo 'Timed out waiting for k3s cluster to be ready after 10 minutes' >&2");
-        sb.AppendLine("    exit 1");
-        sb.AppendLine("  fi");
-        sb.AppendLine("  echo 'Waiting for k3s cluster to be ready and reachable...'");
-        sb.AppendLine("  sleep 5");
-        sb.AppendLine("done");
+        sb.AppendShellLine("_k3s_wait=0");
+        sb.AppendShellLine($"until [ -f {K3sFileHelpers.ContainerKubeconfigPath} ] && helm list --kubeconfig {K3sFileHelpers.ContainerKubeconfigPath} > /dev/null 2>&1; do");
+        sb.AppendShellLine("  _k3s_wait=$((_k3s_wait + 5))");
+        sb.AppendShellLine("  if [ \"$_k3s_wait\" -ge 600 ]; then");
+        sb.AppendShellLine("    echo 'Timed out waiting for k3s cluster to be ready after 10 minutes' >&2");
+        sb.AppendShellLine("    exit 1");
+        sb.AppendShellLine("  fi");
+        sb.AppendShellLine("  echo 'Waiting for k3s cluster to be ready and reachable...'");
+        sb.AppendShellLine("  sleep 5");
+        sb.AppendShellLine("done");
 
         if (release.RepoUrl is not null)
         {
             var alias = $"aspire-k3s-{release.ReleaseName}";
-            sb.AppendLine($"helm repo add --force-update {ShellEscape(alias)} {ShellEscape(release.RepoUrl)}");
-            sb.AppendLine($"helm repo update {ShellEscape(alias)}");
+            sb.AppendShellLine($"helm repo add --force-update {ShellEscape(alias)} {ShellEscape(release.RepoUrl)}");
+            sb.AppendShellLine($"helm repo update {ShellEscape(alias)}");
         }
 
         var chartRef = release.RepoUrl is not null
--- a/src/CommunityToolkit.Aspire.Hosting.K3s/K3sBuilderExtensions.Manifest.cs
+++ b/src/CommunityToolkit.Aspire.Hosting.K3s/K3sBuilderExtensions.Manifest.cs
@@ -166,27 +166,27 @@ public static class K3sManifestBuilderExtensions
         // DCP sets up container network aliases asynchronously, so the kubeconfig file
         // can appear in the bind-mount before the k8s hostname resolves in the kubectl
         // container. Using `kubectl cluster-info` verifies both the file and the network.
-        sb.AppendLine("_k3s_wait=0");
-        sb.AppendLine($"until [ -f {K3sFileHelpers.ContainerKubeconfigPath} ] && kubectl cluster-info --kubeconfig {K3sFileHelpers.ContainerKubeconfigPath} > /dev/null 2>&1; do");
-        sb.AppendLine("  _k3s_wait=$((_k3s_wait + 5))");
-        sb.AppendLine("  if [ \"$_k3s_wait\" -ge 600 ]; then");
-        sb.AppendLine("    echo 'Timed out waiting for k3s cluster to be ready after 10 minutes' >&2");
-        sb.AppendLine("    exit 1");
-        sb.AppendLine("  fi");
-        sb.AppendLine("  echo 'Waiting for k3s cluster to be ready and reachable...'");
-        sb.AppendLine("  sleep 5");
-        sb.AppendLine("done");
+        sb.AppendShellLine("_k3s_wait=0");
+        sb.AppendShellLine($"until [ -f {K3sFileHelpers.ContainerKubeconfigPath} ] && kubectl cluster-info --kubeconfig {K3sFileHelpers.ContainerKubeconfigPath} > /dev/null 2>&1; do");
+        sb.AppendShellLine("  _k3s_wait=$((_k3s_wait + 5))");
+        sb.AppendShellLine("  if [ \"$_k3s_wait\" -ge 600 ]; then");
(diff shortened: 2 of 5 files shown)


henrikroschmann:
@dotnet-policy-service agree

aaronpowell (approved):

PR #1555, opened 2026-08-31
feat(floci): model Service Bus as a child resource of AddFlociAzure

### LINKED ISSUE #1552: [Feature]: CommunityToolkit.Aspire.Hosting.Floci — Service Bus connection string helper for AddFlociAzure
### Package

CommunityToolkit.Aspire.Hosting.Floci

### Is your feature request related to a problem? Please describe.

`AddFlociAzure(...)` injects a ready-to-use `AZURE_STORAGE_CONNECTION_STRING` for the storage SDKs, and a Cosmos connection-string helper is proposed in #1545 (PR #1546). There is no equivalent for **Service Bus**, so a .NET app cannot reach floci-az Service Bus through the standard Aspire connection-string flow (`builder.AddAzureServiceBusClient("servicebus")`) at all.

Unlike Blob/Cosmos, this is not just a missing helper: floci-az's Service Bus AMQP data plane terminates on an **Artemis sidecar container that floci-az spawns itself at runtime** — the Aspire-modeled resource is only the main container, so there is no `EndpointReference` to build a connection string from. Filed upstream as floci-io/floci-az#249; this issue tracks the toolkit half.

### Describe the solution you'd like

Once floci-az exposes a modelable AMQP endpoint (floci-io/floci-az#249), add a `WithServiceBusReference` helper mirroring the storage/Cosmos pattern:

```csharp
var azure = builder.AddFlociAzure("floci-az");

    .WithServiceBusReference(azure)      // connectionName defaults to "servicebus"
    .WaitFor(azure);
```

Injecting (endpoint resolved by Aspire so it tracks random/isolated ports):

```
ConnectionStrings__servicebus = Endpoint=sb://{host}:{amqpPort};SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=SAS_KEY_VALUE;UseDevelopmentEmulator=true;
```

With its own `[AspireExport]` TS binding, following the shape of `withFlociAzureCosmosReference` in #1546.

**Blocked by floci-io/floci-az#249** — until the AMQP port is declarable on the modeled container, the helper has nothing to reference.

### Additional context

I'm happy to contribute the PR once the upstream endpoint exists (as with #1546).

thomhurst:
Fixes #1552.

Adds `WithServiceBus()` to model the Floci Azure Service Bus AMQP data plane as a child resource. The emulator starts its Artemis sidecar at boot, and Aspire allocates proxyless AMQP/AMQPS endpoints. Referencing the child injects the connection string used by `AddAzureServiceBusClient("servicebus")`.

```csharp
var azure = builder.AddFlociAzure("floci-az").WithDockerSocket();
var serviceBus = azure.WithServiceBus();

builder.AddProject<MyApi>("api")
    .WithReference(serviceBus)
    .WaitFor(azure);
```

Requires floci-az 0.12.0, already pinned on this branch. That release includes namespace startup, batched sends, and .NET administration-client routing. The management plane remains on the parent emulator endpoint. Multiple emulator instances on one Docker host need distinct `FLOCI_AZ_DOCKER_RESOURCE_NAMESPACE` values; the README documents this requirement.

Validation includes eight unit tests and two Docker integration cases. The integration cases start the real emulator, verify both listeners before queue creation, resolve the consumer's `WithReference` environment, create a queue with `ServiceBusAdministrationClient`, and send, receive, and complete single messages and a three-message batch with `ServiceBusClient`. Each case uses an isolated Docker namespace and a bounded timeout.

Also fixes the TypeScript sample's bind mount to create a portable local directory instead of assuming `/tmp/floci-data` exists. The branch includes current `main`, which removes the obsolete Source Link dependency that blocked restore.

Local validation: `dotnet test --project tests/CommunityToolkit.Aspire.Hosting.Floci.Tests -c Release --no-progress --output Normal` passed all 75 tests, with no skips, including TypeScript AppHost compilation/startup and the Docker integration cases. NuGet audit remained enabled.


src/CommunityToolkit.Aspire.Hosting.Floci/api/CommunityToolkit.Aspire.Hosting.Floci.cs
aaronpowell, line 1: Drop these changes as we automate their addition

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Azure.cs
```diff
@@ -73,6 +76,89 @@ public static IResourceBuilder<TDestination> WithReference<TDestination>(
...
+        FlociAzureServiceBusResource serviceBus = new(
+            name, amqpPort ?? fallbackAmqpPort, amqpTlsPort ?? fallbackAmqpTlsPort, builder.Resource);
+
+        builder
+            .WithEnvironment("FLOCI_AZ_SERVICES_SERVICE_BUS_MOCKED", "false")
+            .WithEnvironment("FLOCI_AZ_SERVICES_SERVICE_BUS_START_ON_BOOT", "true")
+            .WithEnvironment("FLOCI_AZ_SERVICES_SERVICE_BUS_AMQP_PORT",
+                serviceBus.AmqpPort.ToString(CultureInfo.InvariantCulture))
+            .WithEnvironment("FLOCI_AZ_SERVICES_SERVICE_BUS_AMQP_TLS_PORT",
+                serviceBus.AmqpTlsPort.ToString(CultureInfo.InvariantCulture));
```
aaronpowell: I'm not sure this will result in the container binding those ports, but also, I have a bit of a concern on how it'll interact with DCP managing the endpoints.

Is there a reason we couldn't use `WithEndpoint` and let the endpoint allocation handle the port allocation, and then use the `WithEnvironment` method that takes a callback to set the environment variables based off the port allocated from the endpoint.

aaronpowell (changes requested):

aaronpowell:
@thomhurst - are you interested in being added as a contributor and being given the CODEOWNER for the floci integration?

src/CommunityToolkit.Aspire.Hosting.Floci/api/CommunityToolkit.Aspire.Hosting.Floci.cs
thomhurst: Dropped the hand-edited API baseline changes; the file now matches upstream `main` exactly.

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Azure.cs
thomhurst: Updated this to use Aspire endpoint allocation. The Service Bus child now exposes proxyless `amqp` and `amqps` endpoints via `WithEndpoint`, and the parent `WithEnvironment` callback reads their DCP-allocated ports. I kept the endpoints on the child because the Artemis sibling—not the floci-az container—binds those host ports; proxyless endpoints avoid forwarding traffic to the wrong container. Also bumped floci-az to 0.12.0, which contains start-on-boot support.

thomhurst:
Happy to be added as a contributor but not sure I'll have time to be a code owner and review all the changes if it gets busy

aaronpowell:
Well, the floci changes seem to mostly come from you 😅

aaronpowell (approved):
Looks good to merge when the external blocker is resolved.

afscrome:
Looks like this got released in https://github.com/floci-io/floci-az/releases/tag/0.12.0 .  Is this Pr ready for review?

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Azure.cs
```diff
@@ -73,6 +74,85 @@ public static IResourceBuilder<TDestination> WithReference<TDestination>(
...
+    /// <returns>A reference to the <see cref="IResourceBuilder{FlociAzureServiceBusResource}"/> for further configuration.</returns>
+    [AspireExport]
+    public static IResourceBuilder<FlociAzureServiceBusResource> WithServiceBus(
+        this IResourceBuilder<FlociAzureContainerResource> builder,
+        [ResourceName] string name = FlociAzureServiceBusResource.DefaultName,
+        int? amqpPort = null,
+        int? amqpTlsPort = null)
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentException.ThrowIfNullOrWhiteSpace(name);
```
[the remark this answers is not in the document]
aaronpowell: This check is a bit further down.d

src/CommunityToolkit.Aspire.Hosting.Floci/FlociHostingExtension.Azure.cs
```diff
@@ -73,6 +74,85 @@ public static IResourceBuilder<TDestination> WithReference<TDestination>(
...
+                throw new InvalidOperationException(
+                    $"Service Bus is already configured on '{builder.Resource.Name}' and cannot be reconfigured with different ports.");
+            }
+
+            return builder.ApplicationBuilder.CreateResourceBuilder(existing);
+        }
+
+        var serviceBus = new FlociAzureServiceBusResource(name, builder.Resource);
+        var serviceBusBuilder = builder.ApplicationBuilder
+            .AddResource(serviceBus)
```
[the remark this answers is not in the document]
aaronpowell: Copilot is technically correct here, if the `name` parameter isn't provided then yes, there will be a conflict, but it is a settable parameter so it can be resolved by the user with minimal effort.

aaronpowell (approved):

